PR #29, opened 2018-05-11
[Enhancement] Enable ordinal aggregation in aggregation layer (hex, grid, cluster)

heshan0131:
- Enable ordinal aggregation type `mode` and `count unique` in aggregation layer
- Added additional validation in visual channel updater to validate aggregation type and scale based on selected field
- Added more condition clause to layer visConfigSettings to render configuration selectively based on current layer config
- Specify aggregation type in layer legend and map popover 

**Hexbin aggregation by event type**
screen shot 2018-05-13 at 1 47 30 pm

**Legend**
screen shot 2018-05-13 at 1 47 40 pm

### DIFF
--- a/examples/demo-app/src/app.js
+++ b/examples/demo-app/src/app.js
@@ -50,7 +50,7 @@ class App extends Component {
 
   componentDidMount() {
     // load sample data
-    // this._loadSampleData();
+    this._loadSampleData();
   }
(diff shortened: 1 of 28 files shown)


heshan0131:
https://github.com/uber/kepler.gl/issues/98

PR #100, opened 2018-06-04
bugfix: TimeRangeSlider should not cache props.onChange

### LINKED ISSUE #99: props is cached in TimeRangeSlider.constructor
https://github.com/uber/kepler.gl/blob/master/src/components/common/time-range-slider.js#L68

onChange is cached in constructor and nothing handle the change of the onChange handler
this make this component not reusable and might cause issue since react reuse component whenever possible

that lines in constructor() should be moved to render()

rinick:
onChange should be updated in render() function to make TimeRangeSlider reusable
fix #99

### DIFF
--- a/src/components/common/time-range-slider.js
+++ b/src/components/common/time-range-slider.js
@@ -65,7 +65,6 @@ export default class TimeRangeSlider extends Component {
       width: 288
     };
     this._animation = null;
-    this._sliderThrottle = throttle(this.props.onChange, 20);
   }
 
   componentDidUpdate() {
@@ -119,9 +118,9 @@ export default class TimeRangeSlider extends Component {
   };
 
   render() {
-    const {domain, value, isEnlarged} = this.props;
+    const {domain, value, isEnlarged, onChange} = this.props;
     const {isAnimating} = this.state;
-
+    this._sliderThrottle = throttle(onChange, 20);
     return (
       <div className="time-range-slider">
         <TimeTitle
--- a/src/components/common/time-range-slider.js
+++ b/src/components/common/time-range-slider.js
@@ -65,6 +65,7 @@ export default class TimeRangeSlider extends Component {
       width: 288
     };
     this._animation = null;
+    this._sliderThrottle = throttle((value) => this.props.onChange(value), 20);
   }
 
   componentDidUpdate() {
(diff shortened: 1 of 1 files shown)


src/components/common/time-range-slider.js
```diff
@@ -119,9 +118,9 @@ export default class TimeRangeSlider extends Component {
   };
 
   render() {
-    const {domain, value, isEnlarged} = this.props;
+    const {domain, value, isEnlarged, onChange} = this.props;
     const {isAnimating} = this.state;
-
+    this._sliderThrottle = throttle(onChange, 20);
```
chrisirhc: This creates a new `_sliderThrottle` even if `onChange` is the same. It may cause many unnecessary re-renders.
How about saving the `onChange` in the instance and creating a new `_sliderThrottle` only as needed (memoization)?

chrisirhc:
This is awesome, thanks for finding this. However, this PR may introduce performance regressions. I've included a comment.

Perhaps the right place to do this type of logic in React 16, is [`getDerivedStateFromProps()`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops).

src/components/common/time-range-slider.js
rinick: since onChange is passed in as a lambda. so it's always a different lambda when parent widget renders. Is it really necessary or even possible to add that optimization?

chrisirhc: That's an excellent point. In that case, how about using a lambda for calling `this.props.onChange` and leaving this constructed once in the constructor?

src/components/common/time-range-slider.js
```diff
@@ -65,7 +65,7 @@ export default class TimeRangeSlider extends Component {
       width: 288
     };
     this._animation = null;
-    this._sliderThrottle = throttle(this.props.onChange, 20);
+    this._sliderThrottle = throttle((value) => this.props.onChange(value), 20);
```
chrisirhc: Minor nitpick, might want to use the spread operator to apply all arguments to ensure there's no issues in the future when more arguments are used in the `onChange` prop method.

rinick: done

chrisirhc:
Looks good to me 👍. Thanks!
In the future, we may want to introduce tests for such changes, but I'd say for fixing this now, it's good as-is, and also the functional change is small.

Issue #84, opened 2018-06-01
P1: Can't remove a dataset if the name is too long


yiyaya66:
![image pasted at 2018-5-11 15-22](https://user-images.githubusercontent.com/28815578/40854183-72aca1ce-6585-11e8-81d3-1cda6a2f03e6.png)

PR #109, opened 2018-06-07
Bugfix: Create ellipsis when dataset name is a long name

ctrimm:
The problem is documented in #84. Essentially, if you have a dataset without a space in the name, it will stretch the `.dataset-name` container really wide - making it impossible to see the delete or view the dataset you uploaded.

The solution I took was to add a `width`, `overflow`, and `text-overflow` to the `.dataset-name` class. This will also keep all dataset names on 1 line. When data sets had spaces and were very long, it would put the dataset name on multiple lines. I feel that this is a consistent approach to solving this issue.

Photo of proposed solution - 
screen shot 2018-06-06 at 6 34 05 pm

Looking forward to utilizing and contributing to kepler with a bigger PR in the future. Let me know if there are any issues with this proposed solution.

### DIFF
--- a/src/components/side-panel/source-data-catalog.js
+++ b/src/components/side-panel/source-data-catalog.js
@@ -36,7 +36,7 @@ const DatasetTitle = styled.div`
   color: ${props => props.theme.textColor};
   display: flex;
   align-items: flex-start;
-  
+
   .source-data-arrow {
     height: 16px;
   }
@@ -60,9 +60,16 @@ const DatasetTagWrapper = styled.div`
   color: ${props => props.theme.textColor};
   font-size: 11px;
   letter-spacing: 0.2px;
-  
+
   .dataset-color {
     margin-top: 5px;
+  }
+
+  .dataset-name {
+    width: 11rem;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
 `;
 


chrisirhc:
Nice clean fix. I think we can also consider adding a hover interaction similar to what's on http://yurigor.com/show-hidden-part-of-truncated-text-on-hover-touch/ . However, that can be a subsequent PR as well.

src/components/side-panel/source-data-catalog.js
```diff
@@ -60,10 +60,17 @@ const DatasetTagWrapper = styled.div`
...
   font-size: 11px;
   letter-spacing: 0.2px;
-  
+
   .dataset-color {
     margin-top: 5px;
   }
+
+  .dataset-name {
+    width: 11rem;
```
chrisirhc: Any way to use a constant or some variable so that this adapts to changes on the width of the side bar (if it gets updated)?

chrisirhc: Also, should this be `max-width` to retain the current behavior of showing the arrow at the end when the string is shorter than the width?

heshan0131:
@ctrimm are you able to update this PR based on the comment? If not, I can also take over from here with your permission

ctrimm:
@heshan0131 - Hey, sorry for the large delay on this - got caught up elsewhere. I believe that I have this fixed, I just wanted to manually test it and then push it up. It'll be up either later today or early tomorrow morning.

ctrimm:
@heshan0131 - Here is now a temporary solution to the long dataset name issue. It doesn't have a dynamic width per @chrisirhc's suggestion. But, it works for a simple stop-gap solution for the time being. Below is a screenshot - 

screen shot 2018-07-01 at 6 14 32 pm

src/components/side-panel/source-data-catalog.js
```diff
@@ -55,15 +55,23 @@ const DatasetTitle = styled.div`
...
 const DatasetTagWrapper = styled.div`
   display: flex;
   color: ${props => props.theme.textColor};
   font-size: 11px;
   letter-spacing: 0.2px;
-  
+  overflow: auto;
+
   .dataset-color {
     margin-top: 5px;
```
heshan0131: can you add `flex-shrink: 0;` so the square doesn't get squeezed?

ctrimm: Certainly. I'll add that property once I'm home from work. Thanks for the response @heshan0131.

ctrimm: Hey @heshan0131 where exactly do you want me to place this? The comment is on lien 65, but I didn't add any of that code.

heshan0131: To ‘.dataset-color’

ctrimm: Gotcha, thanks. Just updated that & pushed.

PR #119, opened 2018-06-15
WIP Add Tracking

### LINKED ISSUE #116: Add analytics tracking for usage kepler.gl
Consider adding a middleware for the redux store and push events from actions. 

Events we're interested are tracked in [this spreadsheet](https://docs.google.com/spreadsheets/d/1-uxOTnEgV-6PASYzmULhq-YHUZum906_t2eoDZ2vmHk/edit#gid=0).

benshope:
This adds some Google Analytics tracking to the demo site.

Resolves #116


examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,53 @@
+import { ActionTypes } from "kepler.gl/actions";
```
heshan0131: Please confit your editor settings to get rid of extra space in the bracket

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,53 @@
...
+
+import { LOCATION_CHANGE } from "react-router-redux";
+
+const {
+  ADD_DATA,
```
heshan0131: Maybe easier to not de-construct here and just use `ActionTypes. ADD_DATA` in the following function. So that people don't have to edit in 2 places when adding new action to be tracked

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,53 @@
...
+  UPDATE_MAP,
+  TOGGLE_MODAL,
+  TOGGLE_SIDE_PANEL
+} = ActionTypes;
+
+const getPayload = ({ payload }) => payload;
+
+const trackingInformation = {
+  [ADD_DATA]: ({ payload: { files } }) =>
+    files.map(({ name, size, type }) => ({ name, size, type })),
```
heshan0131: let's not track the name, only `type` and `size` would be sufficient

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,53 @@
...
+  MAP_STYLE_CHANGE,
+  UPDATE_MAP,
+  TOGGLE_MODAL,
+  TOGGLE_SIDE_PANEL
+} = ActionTypes;
+
+const getPayload = ({ payload }) => payload;
+
+const trackingInformation = {
+  [ADD_DATA]: ({ payload: { files } }) =>
```
heshan0131: `ADD_DATA` is not used, `LOAD_FILES` is the one you want to track

examples/demo-app/src/analytics.js
benshope: Moving the .prettierrc to the root directory fixes this - if it's okay to do that

examples/demo-app/src/analytics.js
benshope: Good call, yep!

examples/demo-app/src/analytics.js
benshope: Yep!

examples/demo-app/src/analytics.js
benshope: Right, better not to upload anything about the users' data that they might not want

benshope:
@heshan0131 ready for review again

chrisirhc:
One thing that I forgot to mention in #116, can we send an event to track when a user clicks on the Save/Export button? (See screenshot below)

image

benshope:
@chrisirhc got it, so even though it isn't firing an action we'd like to see when the user opens that pop

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,78 @@
+import {ActionTypes} from 'kepler.gl/actions';
```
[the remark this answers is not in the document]
benshope: Added!

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+
+const trackingInformation = {
+  [ActionTypes.LOAD_FILES]: ({files}) =>
+    files.map(({size, type}) => ({size, type})),
+  [ActionTypes.LAYER_HOVER]: ({
+    payload: {
+      info: {lngLat}
+    }
+  }) => lngLat,
+  [ActionTypes.LAYER_CONFIG_CHANGE]: ({
```
heshan0131: LAYER_TYPE_CHANGE

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+      newConfig
+    }
+  }) => ({
+    type,
+    newConfig
+  }),
+  [ActionTypes.MAP_STYLE_CHANGE]: getPayload,
+  [ActionTypes.TOGGLE_MODAL]: getPayload,
+  [ActionTypes.TOGGLE_SIDE_PANEL]: getPayload,
+  [ActionTypes.UPDATE_MAP]: withStoreInformation(getPayload),
```
heshan0131: this is map viewport change, don't track

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+      oldLayer: {type},
+      newConfig
+    }
+  }) => ({
+    type,
+    newConfig
+  }),
+  [ActionTypes.MAP_STYLE_CHANGE]: getPayload,
+  [ActionTypes.TOGGLE_MODAL]: getPayload,
+  [ActionTypes.TOGGLE_SIDE_PANEL]: getPayload,
```
heshan0131: We don't need to track this

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+  [ActionTypes.UPDATE_MAP]: withStoreInformation(getPayload),
+  [ActionTypes.SET_FILTER]: withStoreInformation(getPayload),
+  [ActionTypes.INTERACTION_CONFIG_CHANGE]: ({config}) => config,
+  [LOCATION_CHANGE]: x => x
+};
+
+const analyticsMiddleware = store => next => action => {
```
heshan0131: Can we pass the after state of the store to the middleware?

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }) => ({
+    type,
+    newConfig
+  }),
+  [ActionTypes.MAP_STYLE_CHANGE]: getPayload,
+  [ActionTypes.TOGGLE_MODAL]: getPayload,
+  [ActionTypes.TOGGLE_SIDE_PANEL]: getPayload,
+  [ActionTypes.UPDATE_MAP]: withStoreInformation(getPayload),
+  [ActionTypes.SET_FILTER]: withStoreInformation(getPayload),
```
heshan0131: assume this is where we track the type of filter being used. payload would be not wrapped. and it would sed {idx, prop, value}
```
const {prop} = payload;
if (prop === 'name') {
// get the type of the filter from the next store
const visState = store.demo.keplerGl.map.visState;
const filter  = visState.filters[idx]
const newType = filter.type;

const trackingPayload = {type: newType}
}

```

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,93 @@
...
+    console.warn(err);
+    /* eslint-enable */
+    return payload;
+  }
+};
+
+const trackingInformation = {
+  [ActionTypes.LOAD_FILES]: ({files}) =>
+    files.map(({size, type}) => ({size, type})),
+  [ActionTypes.LAYER_HOVER]: ({
```
heshan0131: We don't want to track LAYER_HOVER, it gets called all the time

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,93 @@
...
+      info: {lngLat}
+    }
+  }) => lngLat,
+  [ActionTypes.LAYER_TYPE_CHANGE]: ({payload: {newType}}) => ({
+    newType
+  }),
+  [ActionTypes.MAP_STYLE_CHANGE]: getPayload,
+  [ActionTypes.TOGGLE_MODAL]: getPayload,
+  [ActionTypes.SET_FILTER]: getFilterFromStore,
+  [ActionTypes.INTERACTION_CONFIG_CHANGE]: ({config}) => config,
```
heshan0131: if we want to track whether tooltip or brushing is enabled, we should just track that
`({config}) => ({[config.id]:{config.enabled]})`

examples/demo-app/src/analytics.js
```diff
@@ -0,0 +1,93 @@
...
+  [LOCATION_CHANGE]: x => x
```
heshan0131: to track how many layers and filters we can do
```
[ActionTypes.ADD_LAYER]: (payload, store) => ({
  total: store.demo.keplerGl.map.visState.layers.length
}),
[ActionTypes.ADD_FILTER]: (payload, store) => ({
  total: store.demo.keplerGl.map.visState.filters.length
}),
```

src/components/side-panel/panel-header.js
```diff
@@ -273,6 +277,14 @@ function PanelHeaderFactory() {
       );
     }
   }
+  return connect(undefined, {
```
heshan0131: you are adding tracking to `kepler.gl/src` directly here. We don't want to do that. Also, we should now randomly call connect in just this component. connect should only called once in the top level component. If you need, we can add a new `ui-state` action `TOGGLE_SIDE_PANEL_HEADER` and `showHeaderDropdown` prop in `ui-state` to trigger this behavior

examples/demo-app/src/analytics.js
benshope: All actions are currently tracked - should I specifically exclude this one?

examples/demo-app/src/analytics.js
benshope: Changed

benshope:
@heshan0131 this should be good to deploy - I had to fix a few very strange test failures so give it a thorough last check (particularly those parts)

src/components/side-panel/panel-header.js
```diff
@@ -204,26 +204,43 @@ const defaultActionItems = [
...
         showExportDropdown,
         hideExportDropdown
-      }) => (
+      } = this.props;
+
+      return (
         <StyledPanelHeader className="side-panel__panel-header">
           <StyledPanelHeaderTop className="side-panel__panel-header__top">
-            <logoComponent version={version}/>
+            <this.props.logoComponent appName={appName} version={version}/>
```
heshan0131: PanelHeader can't be rendered as a pure functional component, because we can only use `this.props.logoComponent` syntax to render the local component, otherwise, react will interpret `logoComponent` as a tag

PR #131, opened 2018-06-21
[Bug] Fix icon layer instructions

heshan0131:
- Move layer info modal to layer class
- Add a button to layer panel to open layer info modal if defined

icon-layer-how-to


src/components/side-panel/layer-panel/layer-config-group.js
```diff
@@ -38,6 +38,7 @@ export const StyledLayerConfigGroupLabel = styled.div`
 export const StyledLayerConfigGroup = styled.div`
   padding-left: 18px;
   margin-bottom: 18px;
+
```
javidhsueh: Extra line

PR #136, opened 2018-06-22
Fix package.json dependencies

chrisirhc:
Was trying to install this package on glitch.com and ran into some issues due to the way certain dependencies are specified.


package.json
```diff
@@ -90,17 +90,17 @@
     "react-file-drop": "^0.1.8",
     "react-map-gl": "^3.2.1",
     "react-modal": "^3.1.10",
-    "react-onclickoutside": "https://registry.npmjs.org/react-onclickoutside/-/react-onclickoutside-6.7.1.tgz",
```
heshan0131: the url is specified because if using unpm registry, this will try to get @uber/react-onclickoutside, although I haven't check, is that still the case?

chrisirhc: It's no longer the case. We also ensure that it's not the case through the yarn.lock .

Issue #150, opened 2018-07-02
download as an image feature is not working


axs165032:
I tried downloading the visualization, I was working on and the export as image feature did not work

heshan0131:
Fix is coming

heshan0131:
This should be resolved. Closing it for now

chintanp:
The export to image feature is still not working for me. The animation circle keeps on rotating, and nothing happens further.

heshan0131:
Hi @clintanp, which browser are you using? Can you open your browser console while you export and take a screenshot of the errors if there is any?

chintanp:
Please see attached. I am using Firefox 61.0 (64-bit) on Windows 10. 

![kepler gl_error_export](https://user-images.githubusercontent.com/752869/42484312-2ce7644a-83a7-11e8-986a-bcadb8acd87a.PNG)

heshan0131:
@chintanp 
Are you using the kepler.gl/#/demo app or running kepler.gl in your local environment
Seems like firefox throws an error when requesting CORS stylesheets.

I can't seem to repoduce this error in my firefox at the moment. (I m on mac). 
You can try to use chrome on windows as a workaround

chintanp:
I am using the http://kepler.gl/#/demo app and it doesn't work in Chrome (67.0) on my machine too.  Error in chrome as below: 

![kepler gl_error_export_chrome](https://user-images.githubusercontent.com/752869/42541385-f91d1a96-8456-11e8-8a69-6a5c55c1311e.PNG)

chrisirhc:
@chintanp are you still seeing this issue? If so, could you clear your cache and try again?
I'm unable to reproduce it and I do see the required `Access-Control-Allow-Origin` header sent by the server:
![image](https://user-images.githubusercontent.com/132584/42545294-e5dd94a0-846b-11e8-89ca-2833605aa21e.png)

chintanp:
I tried again after clearing cache etc. but I am still getting the same error and it doesn't work for me. If it works for everyone else, then maybe something wrong with my setup.

heshan0131:
@chintanp if you can open your network tab, find the refresh.css request, and send me a screenshot of the Header, that would be really helpful. I just helped someone debug this issue using chrome. As soon as I check the `disable cache` box, the image showed up.

[...]

PR #187, opened 2018-08-02
Fix unable to fetch external stylesheets when taking the screenshot

javidhsueh:
Fix issue #150


heshan0131:
LGTM. Thanks for investigating this.

PR #194, opened 2018-08-13
[Chore] Add install-and-start script

ericsoco:
I noticed that `path-exists` doesn't exist / is not an executable, and so the npm scripts (like `start`) that do `path-exists node_modules || npm i` were always running `npm i`.

This PR adds a simple node script that does the same thing, only correctly.

Also, this adds an `.nvmrc` that matches `package.json::engines` and updates `README`.

The `panel-header.js` newline is probably a prettier fix, it happened on build after initial install.


ericsoco:
@heshan0131 @macrigiuseppe landable?

scripts/install-and-start.js
```diff
@@ -0,0 +1,36 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+const {existsSync} = require('fs');
```
heshan0131: wonder if having ES6 and running it with `node` will error on older version of node. But since we require 9 and above, it should be fine

ericsoco: Shouldn't be a problem, [`const`](https://node.green/#ES2015-bindings-const) and [object destructuring](https://node.green/#ES2015-syntax-destructuring--declarations-with-objects) have been supported since `node@6`.

PR #206, opened 2018-08-22
[Bug] Fix server render error, remove react-ace

heshan0131:
- react-ace is causing server-side rendering error: window is not defined
- Remove `react-ace` reduces 1mb in bundle size


src/reducers/vis-state-updaters.js
```diff
@@ -56,10 +56,16 @@ import {
   mergeLayerBlending
 } from './vis-state-merger';
 
-import {LayerClasses, Layer} from 'layers';
+// LayerClasses contain ES6 Class, do not instatiate in iso rendering
```
benshope: necessary to keep this comment?

PR #234, opened 2018-09-13
Fix z-fighting issue between text label and scatter plot

javidhsueh:
- Enable fp64
 - Turn off depth test for text layer to avoid z-fighting
 - Turn off depth test for scatterplot layer when altitude column doesn't exist
 - Fix lint warning in map-legend.js (extra line)

### DIFF
--- a/src/layers/point-layer/point-layer.js
+++ b/src/layers/point-layer/point-layer.js
@@ -278,6 +278,10 @@ export default class PointLayer extends Layer {
               getTextAnchor: this.config.textLabel.anchor,
               getText: d => String(d.data[this.config.textLabel.field.tableFieldIndex - 1]),
               getColor: d => this.config.textLabel.color,
+              fp64: this.config.visConfig['hi-precision'],
+              parameters: {
+                depthTest: false
+              },
               updateTriggers: {
                 getPosition: data.getPosition,
                 getPixelOffset: this.config.textLabel.offset,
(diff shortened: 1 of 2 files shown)


javidhsueh:
fix-text-z

src/layers/point-layer/point-layer.js
```diff
@@ -248,7 +248,7 @@ export default class PointLayer extends Layer {
         pickable: true,
         // parameters
         parameters: {
-          depthTest: mapState.dragRotate
+          depthTest: this.config.columns.altitude.fieldIdx > -1
```
heshan0131: this should be, 
`mapState.dragRotate || this.config.columns.altitude.fieldIdx > -1`
We want to depth test in 3d mode so that the layer order is correctly shown

src/layers/point-layer/point-layer.js
```diff
@@ -278,6 +278,10 @@ export default class PointLayer extends Layer {
               getTextAnchor: this.config.textLabel.anchor,
               getText: d => String(d.data[this.config.textLabel.field.tableFieldIndex - 1]),
               getColor: d => this.config.textLabel.color,
+              fp64: this.config.visConfig['hi-precision'],
+              parameters: {
+                depthTest: false
```
heshan0131: Please add comment here:
// text will always show on top of all layers

src/layers/point-layer/point-layer.js
javidhsueh: I choose to go with ```this.config.columns.altitude.fieldIdx > -1``` only because this layer should be flat on the map when the altitude column is not used. It shouldn't matter whether in 3d mode or not.

PR #246, opened 2018-10-08
Add/Remove Layer Data

rameshbabuch:
Hi
In this PR I have added two methods to add/remove data from layers which is usefull to add/remove dynamic data. The two methods added are 
- addLayerData
- removeLayerData 

##### addLayerData
This method is similar to addDataToMap and also takes dataset key column to update the layer data.
This action takes an object as input with the following properties:
```
js
{
    datasets | object: same as UpdateVisData
    config | object: this object will contain the full kepler.gl instance configuration {mapState, mapStyle, visState}.
    key | string: dataset column 
    
}
```
Example
```js
this.props.dispatch(
                wrapTo('map1', addLayerData(
                    {
                        datasets: datasets,
                        config: config,
                        key: 'stations'
                    })
                )
            );
```

##### removeLayerData
This method is similar to removeLayerData and also takes dataset key column to update the layer data.
This action takes an object as input with the following properties:
```
js
{
    datasets | object: same as UpdateVisData
    key | string: dataset column 
    
}
```
 Example
```js
this.props.dispatch(
                wrapTo('map1', removeLayerData(
                    {
                        datasets: datasets,
                        key: 'stations'
                    })
                )
            );
```

- Feel free to add corretions/suggestions to this PR.


src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+ * @param action {datasets, options, config}
+ * @returns state
+ */
+export const removeLayerDataComposed = (state, action) => {
+  // extract mapData
+  const {payload: mapData} = action;
+  const {key: datasetKey} = mapData;
+  const {datasets, layers} = state.visState;
+  // check datasetKey matches with current state 
+  if (!datasets[datasetKey]) {
```
kpvg: can datasets be null?, do we need to verify if it is null or not

test/node/reducers/composer-state-test.js
```diff
@@ -86,8 +86,8 @@ test('#composerStateReducer - addDataToMapComposed: mapState should not be cente
   // init kepler.gl root and instance
   const state = keplerGlReducer({}, registerEntry({id: 'test'})).test;
   const mapStateProperties = {
-    latitude: 33.88608913680742,
-    longitude: -84.43459130456425
+    latitude: 29.23, //33.88608913680742,
```
kpvg: is this a test case? if not put these constant values inside a configuration file

src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+    if (!array2 || !array1)
+      return false;
+    // compare lengths - can save a lot of time 
+    if (array1.length !== array2.length)
+      return false;
+    for (var i = 0, l = array1.length; i < l; i++) {
+      // Check if we have nested arrays
+      if (array1[i] instanceof Array && array2[i] instanceof Array) {
+        // recurse into the nested arrays
+        if (!array1[i].equals(array2[i]))
```
kpvg: I think we need to call the same function (recursive) as below
if(! equals(array1[i], array2[i])) {

instead of array1[i].equals(array2[i])

src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+          return false;
+      } else if (array1[i] !== array2[i]) {
+        // Warning - two different object instances will never be equal: {x:20} != {x:20}
```
kpvg: ok, but don't we need to compare the values of the objects instead of the instances/references?

src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+  }
+
+  /* eslint-disable no-unused-vars */
+  const {
+    datasets: {[datasetKey]: dataset, ...newDatasets}
+  } = state.visState;
+  /* eslint-enable no-unused-vars */
+
+  // get layer index 
+  const indexes = layers.reduce((listOfIndexes, layer, index) => {
```
kpvg: If the requirement is only to find the indexes, can you verify the performance using a for loop against reduce, and use the better one
https://codeburst.io/reduce-vs-for-loop-3c1a84e63872

src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+      listOfIndexes.push(index);
+    }
+    return listOfIndexes;
+  }, []);
+  
+  function equals(array1, array2) {
+    // if the other array is a falsy value, return
+    if (!array2 || !array1)
+      return false;
+    // compare lengths - can save a lot of time
```
kpvg: preferably also add
  if(! Array.isArray(arr1)) {
     return false;
  }
  if(! Array.isArray(arr2)) {
     return false;
  }

src/reducers/composers.js
```diff
@@ -85,33 +75,147 @@ export const updateVisDataComposed = (state, action) => {
...
+
+  // get layer index 
+  const indexes = layers.reduce((listOfIndexes, layer, index) => {
+    if (layer.config.dataId === datasetKey) {
+      listOfIndexes.push(index);
+    }
+    return listOfIndexes;
+  }, []);
+  
+  function equals(array1, array2) {
```
kpvg: nit: if this method is only to compare two arrays, change the name to arrayEquals()

src/reducers/composers.js
RameshPaul: Yes, sometimes datasets may not contain datasetKey then we don't need to proceed further.

test/node/reducers/composer-state-test.js
RameshPaul: yes, It is a test case.

src/reducers/composers.js
RameshPaul: Since I am comparing only plain arrays, reference/instance comparison is enough.

src/reducers/composers.js
RameshPaul: This array comparison condition added.

heshan0131 (changes requested):
Please don't include yarn.lock changes in this diff

heshan0131:
Is the intention of `addLayerData` to append more rows to an existing dataset? If so, naming it `appendRowsToDataset` would be more descriptive.

src/reducers/composers.js
```diff
@@ -86,32 +77,134 @@ export const addDataToMapComposed = (state, action) => {
...
+  return mergedState;
+};
+/**
+ * Remove layer data from layer data
+ * @param state
+ * @param action {datasets, options, config}
+ * @returns state
+ */
+export const removeLayerDataComposed = (state, action) => {
+  // extract mapData
```
heshan0131: I don't think this is a reasonable way to remove rows from a dataset and update related layers. For one thing, you only removed rows from `layerData` which should not be directly edited, layerData should only be updated by `visState` reducer based on current filtered dataset and layer config. If you didn't remove the actual dataset rows that the layer is associated with. upon any filter or layer config changes, the layer will display the original dataset.

A reasonable way to do this, similar to `addRowsToDataset` is to edit the existing dataset, and call `addDataToMapComposed` to let the visState reducer handles the rest. 

You should also rename this function to `removeRowsFromDataset`. In terms of how to remove date, ideally not by direct comparison and `arrayEquals`, VERY SLOW. 

The argument can be 1. an array of `index` to filter `data.rows`, or 2. a `fieldName` and `uniqueValues` to filter the field value based on.  Option 1 requires the user to keep track of the data added to kepler.gl. Option 2 requires the existing dataset contains a field that has unique values. e.g. `id`, `name`, `index` etc. Either way is fine, rather based on your use cases

RameshPaul: Thanks @heshan0131 I totally missed filters case!, Initially I tried removing from `datasets` since I could not find a way to rerender the graph ended up in removing from `layerData`. 
Thanks for your detailed review now I am thinking to remove rows based on lat longs/geojson instead of comparing entire array since either of the ones(latlong/geojson) will be present for every Kepler graph, do you see any pitfalls with this approach?

README.md
```diff
@@ -26,7 +26,7 @@ For best results, use [nvm](https://github.com/creationix/nvm) `nvm install`.
 Install node (`> 6`), yarn, and project dependencies
 
 ```
-npm install --save kepler.gl
+npm install --save rameshpaul-kepler.gl
```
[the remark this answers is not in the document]
RameshPaul: Yeah, I messed up with the local version, I'll remove this.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "kepler.gl",
   "author": "Shan He <shan@uber.com>",
   "version": "0.1.6",
-  "description": "kepler.gl is a webgl based application to visualize large scale location data in the browser",
```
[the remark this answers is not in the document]
RameshPaul: Yeah, sure.

src/reducers/composers.js
ShirtlessKirk: `reduce` has been optimised in JS engines to be faster than a `for` loop for arrays of length ~= 15 or more. That being said, what is happening here is strictly a `filter`, plus the performance benefit from using the "better" implementation is negligible and is a micro-optimisation.

igorDykhta:
Closing. This is a stale PR with many conflicts in the current codebase.

PR #251, opened 2018-10-08
[Enhancement] click kepler.gl logo should go to kepler.gl website

heshan0131:
https://github.com/uber/kepler.gl/issues/231

### DIFF
--- a/src/constants/default-settings.js
+++ b/src/constants/default-settings.js
@@ -44,6 +44,7 @@ import {
 
 export const KEPLER_GL_NAME = 'kepler.gl';
 export const KEPLER_GL_VERSION = 'v1.0';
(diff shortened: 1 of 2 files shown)


src/constants/default-settings.js
```diff
@@ -44,6 +44,7 @@ import {
 
 export const KEPLER_GL_NAME = 'kepler.gl';
 export const KEPLER_GL_VERSION = 'v1.0';
+export const KEPLER_GL_WEBSITE = 'http://kepler.gl/#/';
```
btford: Does this need the `/#/` ?

Just curious.

macrigiuseppe: I think this is actually needed otherwise it will not kick in react router! Because of github pages

PR #260, opened 2018-10-22
Load data and kepler.gl file using URLs

### LINKED ISSUE #256: Dynamically loading kepler json from the web
I'm wondering if it's possible to develop a method for importing a Kepler json file that's hosted externally via a URL.

For context: I'm teaching students to use this in a class, and we don't have the ability to spin up our own Kepler instance. I'd like the students to be able to provide access to all of the interactive features of the maps. I know they can make the exported json file available to download and allow others to import it themselves, but I think it would be awesome if they could post the json file somewhere (perhaps as a Github gist?) and then provide readers a URL like http://kepler.gl/?data=gist.github.com/etc.

I think this would be a great way for people to make their maps shareable without access to a server without requiring you to store anything.

macrigiuseppe:
Content:
1. Added ability in demo website to load map/data using URLs
2. Refactored the way we load samples using promises

1. This diff will add the ability to load data and kepelr.gl files using urls. It is currently being implemented only in the demo app. once we agree upon UI and implementation we can move forward and embed the code into Kepler.gl core framework. 
The implementation takes advantage of the existing load file action defined in kepler.gl.

2. I refactored the way we load config and data for samples using promises.

New modal dialog for load from URLS

screen shot 2018-10-21 at 6 57 41 pm

kepler-data-url

kepler-keplerglfile-url

kepler-map-url

This work resolves #256


examples/demo-app/src/actions.js
```diff
@@ -62,45 +62,157 @@ export function setLoadingMapStatus(isMapLoading) {
...
+    request(url, (error, result) => {
+        error ? reject(error) : resolve(result.response)
+    })
+  });
+}
+
+// The following methods are only used to load SAMPLES
+/**
+ *
+ * @param options {
```
heshan0131: not valid JSDoc

package.json
```diff
@@ -28,13 +28,13 @@
     "fast-test": "NODE_ENV=test babel-tape-runner ./test/node.js && NODE_ENV=test babel-tape-runner -r ./test/setup-browser-env.js ./test/browser.js",
     "test-browser": "NODE_ENV=test babel-tape-runner -r ./test/setup-browser-env.js ./test/browser.js",
     "test-node": "NODE_ENV=test babel-tape-runner ./test/node.js | tap-spec",
-    "test": "npm run test-node && npm run test-browser",
+    "test": "npm run lint && npm run test-node && npm run test-browser",
     "cover": "nyc --reporter html --reporter cobertura --reporter json-summary npm test && nyc report",
     "start": "npm run install-and-start -- examples/demo-app start-local",
+    "start:": "npm run install-and-start -- examples/open-modal start-local",
```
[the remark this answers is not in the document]
macrigiuseppe: hmm, both scripts were already present in package.json. It was added in this commit: https://github.com/uber/kepler.gl/commit/25e18b054c931f49fede0312021d65b80b232179

I placed them close to each other to be more consistent.

yarn.lock
```diff
@@ -5,6 +5,7 @@
 "@deck.gl/core@^5.3.3":
   version "5.3.3"
   resolved "https://registry.yarnpkg.com/@deck.gl/core/-/core-5.3.3.tgz#a13c07e5fa3e22297fd450d6da8ab9aac334b1f0"
+  integrity sha512-1a4me1QRM3/wzEVzZRAflX/Uh67gE9/3fWjl35iVY7lvNrCP0wl2PUfaQkmeThTa4Sp/Qq24ZTpHx1QvkQYH8g==
```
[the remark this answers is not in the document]
macrigiuseppe: these entries are generated by the new yarn

examples/demo-app/src/actions.js
```diff
@@ -111,21 +242,19 @@ function loadRemoteMap(sample) {
  * @returns {function(*)}
  */
 export function loadSampleConfigurations(sampleMapId = null) {
-  return (dispatch) => {
+  return dispatch => {
     requestJson(MAP_CONFIG_URL, (error, samples) => {
       if (error) {
         Console.warn(`Error loading sample configuration file ${MAP_CONFIG_URL}`);
```
heshan0131: shouldn't this error also be caught and displayed in the UI?

examples/demo-app/src/actions.js
```diff
@@ -62,45 +71,167 @@ export function setLoadingMapStatus(isMapLoading) {
...
 
-      // Load data
-      requestMethod(sample.dataUrl, (dataError, result) => {
-        loadMapCallback(dispatch, dataError, result, sample, config);
-      });
+  // Load data
+  return new Promise((resolve, reject) => {
+    requestMethod(url, (error, result) => {
+      // TODO: we need to let users know about CORS issues
+      error ? reject(error) : resolve(result)
```
heshan0131: Error will not be presented in all failed request, statusCode  304 for example will not send an error back.

An example of over engineered (but catch all) error handling can be found here:

https://code.uberinternal.com/diffusion/VIVOY/browse/master/src/shared/api/request-utils.js$30-37

examples/demo-app/src/components/load-data-modal/load-remote-map.js
```diff
@@ -0,0 +1,155 @@
...
+    this.props.onLoadRemoteMap({dataUrl});
+  };
+
+  render() {
+    return (
+      <div>
+        <InputForm>
+          <StyledDescription>Load your map using your custom URL</StyledDescription>
+          <StyledInputLabel>
+            You can use the following formats: CSV | JSON | KEPLERGL.JSON. Make sure the url contains the file extension.
```
heshan0131: is KEPLERGL.JSON a required file extension to load the config json? if not, just mention, the json format can also be a kepler.gl config json. Otherwise it's a little confusing

examples/demo-app/src/actions.js
```diff
@@ -62,45 +71,167 @@ export function setLoadingMapStatus(isMapLoading) {
...
+/**
+ *
+ * @param dispatch
+ * @param error
+ * @param result
+ * @param options
+ * @param config
+ */
+function loadMapCallback(dispatch, error, result, options, config) {
+  if (error) {
```
heshan0131: you can return a `dispatch => {}` function, instead of passing the diaptch around
```
return dispatch => {
    if (error) {...}
    else {
       dispatch(...)
    }
}
```

examples/demo-app/src/actions.js
```diff
@@ -62,45 +71,167 @@ export function setLoadingMapStatus(isMapLoading) {
...
+    Console.warn(`Error loading datafile ${options.dataUrl}`);
```
heshan0131: if result in error, we only going to warn them? instead of displaying it in the UI?

macrigiuseppe: leftover from previous implementation that was shuffled around. I will updated it

examples/demo-app/src/actions.js
macrigiuseppe: I will review this one. I think before this diff we were using loadMapCallback in multiple locations but now we may not need to encapsulate it anymore

MosheMalin:
Hi,
This is great feature.
Can it be integrated as is into Kepler itself? (Personally I think the UI is just fine :-) )
Thanks.

Issue #264, opened 2018-10-22
styled-components being explicitly required throws multiple version issues


CharlieMartell:
**Describe the bug**
styled-components being explicitly required throws multiple version issues

**To Reproduce**
1. Make app with styled-components that has a different version
2. install kepler.gl

**Additional context**
See styled-components docs on best practices https://www.styled-components.com/docs/faqs#marking-styledcomponents-as-external-in-your-package-dependencies

PR #265, opened 2018-10-22
fix issue #264, move styled-components to peer-dep

CharlieMartell:
Move styled-components to devdep from dep & adds peer dep on it, per https://www.styled-components.com/docs/faqs#marking-styledcomponents-as-external-in-your-package-dependencies

Also upgrades styled-comp, no breaking changes indicated in their change log.


CharlieMartell:
closes https://github.com/uber/kepler.gl/issues/264

PR #266, opened 2018-10-23
Revert "fix issue #264, move styled-components to peer-dep"

macrigiuseppe:
Reverts uber/kepler.gl#265.
We need to revert this change for now since we discovered more changes are required!


macrigiuseppe:
Reverting latest change

PR #271, opened 2018-11-06
Added a SharedStreets Tiled Layer

starry97:
## Done
- Added a SharedStreets layer in Kepler.gl to visualize pick up / drop off data, which supports encodings changes.

## TODO
- [ ] Add a UI that allows users to select different SharedStreets data sources
- [ ] Calculate the value of visual encodings using the correct domain (not just using sample data domain)
- [ ] Make sure filtering works correctly (may need to send domain to redux)
- [ ] Make sure interactions work correctly
- [ ] Add tests for reducers

image


src/components/map-container.js
```diff
@@ -337,11 +339,22 @@ export default function MapContainerFactory(MapPopover, MapControl) {
           .reduce(this._renderLayer, []);
       }
 
+      const tiledSampleSSDataLayer = deckGlLayers.filter(deckGlLayer => {
```
Pessimistress: There is a major piece missing here - eventually you want to sort and render layers by data source, not `isTiled`. Instead of making a throw-away implementation here, I encourage you to think about what is needed as a base Layer class API to support the generic use case.

Add a TODO comment that explains any assumption/missing parts here.

src/components/map-container.js
```diff
@@ -337,11 +340,46 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
+      tiledDatasets.push('sharedstreets');
+
+      tiledDatasets.forEach(tiledDataset => tiledDataToLayers.set(tiledDataset, []));
+
+      const layersWithTiledData = new Set();
+
+      Object.keys(datasets).forEach(dataId => {
+        const isTiledData = datasets[dataId].isTiled;
+        if (isTiledData) {
+          layers.forEach(layer => {
```
Pessimistress: You can do this by traversing the layers array once.

src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
```diff
@@ -0,0 +1,138 @@
...
+        maxZoom: 12,
+        minZoom: 12,
+        renderSubLayers: this.renderSubLayers.bind(this)
+      })
+    ];
+  }
+
+  recomputeLayerData(keplerLayer, tile) {
+    if (tile.isLoaded) {
+      return keplerLayer.formatLayerData([], tile._data.allData, Array.from(tile._data.allData.keys()));
```
Pessimistress: pass `oldLayerData` here to avoid re-generating data if possible

Issue #130, opened 2018-06-21
heatmap loaded from config makes globe blue in local instance


magamonsta:
So I have an instance of kepler.gl running locally.
I made a heatmap of event data in yall's demo of kepler.gl, then exported the config and data as one, and when i loaded it into my local instance, the data and config loaded but everything was made blue (blue being the lowest color on the color spectrum for the heatmap)....the blue didn't happen when i reloaded the same data file into the demo of kepler.gl yall have running 

I am running an instance out of the kepler.gl/examples/demo-app 

(the data i'm using is scraped from wikipedia terrorism events across the world from 2016-current, it has lat, long, count and is over time so i've found it to be a good dataset for learning how kepler works)

![screenshot 17](https://user-images.githubusercontent.com/38330505/41729542-778d5ec8-7547-11e8-9f8c-936472be50cf.png)

magamonsta:
i've also checked with other data sources and it still happens, i loaded in your earthquake example data, changed it to a heatmap, exported the data and config and loaded into my local instance and everything turned light purple (the lowest color in the color spectrum)

i have also tried exporting the data and config separately and loading them into my local instance and it still happens

magamonsta:
i have also checked to see if it matters how the data is being loaded into my local instance kepler- loaded via a sample.json file like yall have for your demos or loaded via dragging and dropping data in - and it does NOT matter, the problem still happens

alfaifiisa:
Facing the same issue here with the heat map,
It takes the lowest color and makes it global for the whole map.

heshan0131:
closing it as this was fixed

PR #276, opened 2018-11-12
Fixing issue #130 for the heat map

alfaifiisa:
For issue #130 👍 
The main reason is that map box gl need the first color to have opacity of 0.

1. I added a color in the beginning of the colors array
2. replaced that color with rgba(0, 0, 0, 0)


alfaifiisa:
there was an issue with email used for these commit.

PR #277, opened 2018-11-12
Fixing global color issue #130 for the heat map

alfaifiisa:
For issue #130 👍
The main reason is that map box gl need the first color to have opacity of 0.

I added a color in the beginning of the colors array
replaced that color with rgba(0, 0, 0, 0)

### DIFF
--- a/src/layers/heatmap-layer/heatmap-layer.js
+++ b/src/layers/heatmap-layer/heatmap-layer.js
@@ -48,7 +48,7 @@ export const heatmapVisConfigs = {
 const heatmapDensity = (colorRange) => {
   const scaleFunction = SCALE_FUNC.quantize;
 
-  const colors = ['#000000', ...colorRange.colors]
+  const colors = ['#000000', ...colorRange.colors];
 
   const scale = scaleFunction()
     .domain([0, 1])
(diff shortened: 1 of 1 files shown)


src/layers/heatmap-layer/heatmap-layer.js
```diff
@@ -46,21 +46,24 @@ export const heatmapVisConfigs = {
...
-  return scale.range().reduce((bands, level) => {
+  const colorDensity = scale.range().reduce((bands, level) => {
     const invert = scale.invertExtent(level);
     return [
       ...bands,
       invert[0], // first value in the range
       `rgb(${hexToRgb(level).join(',')})` // color
     ]
   }, []);
+  colorDensity[1] = "rgba(0,0,0,0)"
```
heshan0131: comma

src/layers/heatmap-layer/heatmap-layer.js
```diff
@@ -46,21 +46,24 @@ export const heatmapVisConfigs = {
...
+  return colorDensity
```
heshan0131: comma

alfaifiisa: I have added the ;
and used a single quote

PR #279, opened 2018-11-14
[Docs] Add API Docs

heshan0131:


heshan0131:
https://github.com/uber/kepler.gl/issues/259

chrisdiana:
Would love to be able to contribute to this project some way. Looking forward to these docs getting merged in.

heshan0131:
https://github.com/uber/kepler.gl/issues/325

docs/api-reference/reducers/ui-state.md
```diff
@@ -0,0 +1,337 @@
...
+
+## uiStateUpdaters.toggleModalUpdater
+
+Show and hide modal dialog
+
+### Parameters
+
+-   `state` **[Object][41]** `uiState`
+-   `action` **[Object][41]** 
+    -   `action.payload` **([string][42] | null)** id of modal to be shown, null to hide modals. One of:-   [`DATA_TABLE_ID`][44]
```
macrigiuseppe: should [`DATA_TABLE_ID`][44]  be on a new line?

examples/demo-app/README.md
```diff
@@ -1,23 +1,29 @@
 # Demo App
 
-This is the src code of kepler.gl demo app.
+This is the src code of kepler.gl demo app. You can copy this folder out and run it locally.
```
macrigiuseppe: I don't think users can simply copy this directory and run the example outside of kepler.gl folder.

There are node dependencies fetched from root node_modules folder based on webpack project settings

heshan0131: we have 2 webpack file, the one included in the demo-app folder should be sufficient to run this. and if we are missing anything we should add it to its package.json

src/reducers/ui-state-updaters.js
```diff
@@ -18,43 +18,189 @@
...
   };
 };
 
+/**
+ * Show and hide modal dialog
+ * @mixin uiStateUpdaters.toggleModalUpdater
+ * @param {Object} state - `uiState`
+ * @param {Object} action
+ * @param {string|null} action.payload - id of modal to be shown, null to hide modals. One of:
+ *  - [`DATA_TABLE_ID`](../constants/default-settings.md#data_table_id)
```
macrigiuseppe: this is the related to my previous comment

README.md
```diff
@@ -94,23 +94,28 @@ const reducers = combineReducers({
...
+export default createStore(reducer, initialState, compose(...enhancers));
+
 ```
 
 If you mount kepler.gl reducer in another address instead of `keplerGl`, or the kepler.gl reducer is not
 mounted at root of your state, you will need to specify the path to it when you mount the component
 with the `getState` prop.
 
-#### 2. Mount kepler.gl Component
+### 2. Mount Component
```
macrigiuseppe: is it worth mentioning Kepler doesn't support server side rendering? People may run into some issues if they try to render kepler using SSR

README.md
```diff
@@ -351,39 +356,18 @@ const myCustomHeaderFactory = () => withState(
...
-    config | object: this object will contain the full kepler.gl instance configuration {mapState, mapStyle, visState}.
-}
-```
+Kepler.gl provides an easy API `KeplerGlSchema.getConfigToSave` to generate a json blob of the current kepler instance configuration.
 
-<p id="config-notice"> It is important to notice that the config object value will always have higher precedence than the options properties.
-For instance, if you provide <code>{centerMap: true}</code> as part of the options object and in your config object you pass
-the mapState configuration with latitude and longitude define as it follows:</p>
+The options properties will alway have higher precedence than config object value.
+For instance, if you provide `{centerMap: true}` as part of the options object and in your config object you pass
```
macrigiuseppe: Can you double check this? if i recall correctly this was the issue we were talking about were the config override centerMap: true

README.md
```diff
@@ -351,39 +356,18 @@ const myCustomHeaderFactory = () => withState(
...
-    options | object: same as UpdateVisData
-    config | object: this object will contain the full kepler.gl instance configuration {mapState, mapStyle, visState}.
-}
-```
+Kepler.gl provides an easy API `KeplerGlSchema.getConfigToSave` to generate a json blob of the current kepler instance configuration.
 
-<p id="config-notice"> It is important to notice that the config object value will always have higher precedence than the options properties.
-For instance, if you provide <code>{centerMap: true}</code> as part of the options object and in your config object you pass
-the mapState configuration with latitude and longitude define as it follows:</p>
+The options properties will alway have higher precedence than config object value.
```
macrigiuseppe: typo: always

README.md
heshan0131: I thought Charlie's diff didn't land

heshan0131: actually, you are right, the config still take precedence

PR #283, opened 2018-11-16
Render pickup/dropoff data with tooltip

starry97:
## Done
- Showing tooltips when hovering over tile layers

## TODO
- [ ] Brush should also work

kapture 2018-11-16 at 15 32 24

### DIFF
--- a/src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
+++ b/src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
@@ -21,7 +21,7 @@
 /* global fetch */
 
 import {CompositeLayer} from '@deck.gl/core';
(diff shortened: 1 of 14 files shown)


src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
```diff
@@ -99,12 +114,17 @@ export default class SharedstreetsLayer extends CompositeLayer {
...
-        id: `${subLayerProps.id}-${layer.id}`,
         data,
         idx,
         objectHovered,
         mapState,
         interactionConfig
+      }, {
+        sampleKeplerLayerId: this.props.id,
+        id: `${subLayerProps.id}-${layer.id}`,
+        tileX: tile.x,
```
Pessimistress: This is no longer needed?

PR #290, opened 2018-11-20
[Enhancement] Export identity actions individually and add JSDocs

heshan0131:


src/actions/actions.js
```diff
@@ -21,20 +21,127 @@
...
+   *   )
+   * );
+   */
+export const addDataToMap = createAction(
+  ActionTypes.ADD_DATA_TO_MAP,
+  (datasets, options, config) => ({datasets, options, config})
+);
+
+/**
+ * Reset all reducers to its initial state
```
macrigiuseppe: Maybe you can add more info on what this action does to the reducer

PR #298, opened 2018-11-27
[Refactor] Unnecessary ternary

norraell:
Replaced unnecessary ternary expression


heshan0131:
This change is unnecessary, `clicked` and `hovered` is intentionally set to `undefined`. The test is failing because of this change

norraell:
Thanks for the reply! Oh I see why, now that I think of it a second time. However, I could not find a test log, is there a way to show the log in the Travis CI Build?

heshan0131:
Yes, you can click on the job id https://travis-ci.com/uber/kepler.gl/jobs/161083310
Also you can `npm run test` locally

PR #302, opened 2018-12-03
Fix auto focus

CharlieMartell:
So there was an issue with keplergl where focus was set and stored in the config and then the data changed. We were still passing in the saved bounds but with new data causing it compute the new bounds based on the new data then overwrite itself with the parsedConfig bounds. 

Also... styleMapConfigUpdater was being called with the mapStyle but it doesn't do anything because it doesn't deconstruct mapStyle.

### DIFF
--- a/src/reducers/composers.js
+++ b/src/reducers/composers.js
@@ -19,12 +19,10 @@
 // THE SOFTWARE.
 
 import ActionTypes from 'constants/action-types';
-import {fitBoundsUpdater} from './map-state-updaters';
 import {toggleModalUpdater} from './ui-state-updaters';
 import {updateVisDataUpdater} from './vis-state-updaters';
 import {receiveMapConfigUpdater as stateMapConfigUpdater} from './map-state-updaters';
 import {receiveMapConfigUpdater as styleMapConfigUpdater} from './map-style-updaters';
-import {findMapBounds} from 'utils/data-utils';
 import KeplerGlSchema from 'schemas';
 
 // compose action to apply result multiple reducers, with the output of one
@@ -40,9 +38,6 @@ import KeplerGlSchema from 'schemas';
  * @returns state new reducer state
  */
 export const updateVisDataComposed = (state, action) => {
-  // keep a copy of oldLayers
-  const oldLayers = state.visState.layers;
-
   const visState = updateVisDataUpdater(state.visState, action);
 
   const defaultOptions = {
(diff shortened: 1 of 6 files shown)


src/reducers/composers.js
```diff
@@ -97,13 +97,13 @@ export const addDataToMapComposed = (state, action) => {
...
     ...mergedState,
-    mapState: stateMapConfigUpdater(mergedState.mapState, {payload: {mapState: parsedConfig && parsedConfig.mapState}})
+    mapState: stateMapConfigUpdater(parsedConfig.mapState, {payload: {mapState: mergedState.mapState}})
   };
 
   // Update mapStyle store
   mergedState = {
     ...mergedState,
-    mapStyle: styleMapConfigUpdater(mergedState.mapStyle, {payload: {mapStyle: parsedConfig && parsedConfig.mapStyle}})
+    mapStyle: {...parsedConfig.mapStyle,  ...mergedState.mapStyle}
```
CharlieMartell: So `styleMapConfigUpdater` doesn't need to be called here cause it is called above on mergedState and only modifies not in mapStyle, furthermore it didnt do anything before because it is not deconstructed in `receiveMapConfigUpdater`

src/reducers/composers.js
```diff
@@ -97,13 +97,13 @@ export const addDataToMapComposed = (state, action) => {
   // Update mapState store
   mergedState = {
     ...mergedState,
-    mapState: stateMapConfigUpdater(mergedState.mapState, {payload: {mapState: parsedConfig && parsedConfig.mapState}})
+    mapState: stateMapConfigUpdater(parsedConfig.mapState, {payload: {mapState: mergedState.mapState}})
```
CharlieMartell: these are going the wrong way before, we were overwriting what we just calculated with what is originally passed in, causing us to calculate bounds then overwrite them with the previous incorrect bounds.

heshan0131: But this is the intended behavior, if you pass mapState in the config, it is always going to override the calculated map bounds. 
https://github.com/uber/kepler.gl/blob/master/README.md#user-content-config-notice

If you want to enable auto center. add {options: {centerMap: true}} to `addDataToMap` and omit passing `mapState` inside `config`

src/reducers/composers.js
heshan0131: what do you mean by `cause it is called above on mergedState`? I don't see it being called anywhere else.

styleMapConfigUpdater is imported from `map-style-updaters.js`
https://github.com/uber/kepler.gl/blob/master/src/reducers/map-style-updaters.js#L177-L215

CharlieMartell: updated the PR :thumb:

src/reducers/composers.js
CharlieMartell: These weren't necesarily the wrong way its just we were calculating bounds in the visState updater even though it was in mapState which was then getting overridden when calling `mapStateUpdaters` - `receiveMapConfigUpdater`. I have adjusted the code now so that bounds are calculated when we update mapState instead. This requires passing the visState to the mapStateUpdater however.

PR #304, opened 2018-12-04
Fix auto focus / respect calculated bounds

CharlieMartell:
Revision to previous diff, 

Previous diff: 
* Moved the logic of calculating bounds into mapStateUpdaters but this doesn't allow focusing on just the newest data set (and we didn't want to break this)
* Instead focused on the center between all data sets, worse user experience

This diff: 
* Should maintain the same behavior in most of the cases, where this is different is: 
     There is a saved config with some lat/lng in mapState, and data/ config are added at once in one call to addDataToMap with option centerMap: true. What would happen is we would calculate the bounds in `updateVisDataUpdater` to the new layer data, but then overwrite the calculated bounds with those in `map-state-updated - receiveMapConfigUpdater` as it would do `{...mergedConfig, ...parsedConfig}`.

### DIFF
--- a/src/reducers/map-state-updaters.js
+++ b/src/reducers/map-state-updaters.js
@@ -61,7 +61,7 @@ export const receiveMapConfigUpdater = (state, action) => {
     ...state,
     ...(action.payload.mapState || {}),
     latitude: state.latitude ? state.latitude : latitude,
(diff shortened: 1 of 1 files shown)


CharlieMartell:
To revisit later

PR #305, opened 2018-12-07
Use the correct domain to calculate accessors in SharedstreetsLayer

starry97:
- use the correct domain to calculate accessors in SharedstreetsLayer -- we expand the domain when new data is loaded and call `updateLayerConfig`

### DIFF
--- a/src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
+++ b/src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
@@ -24,16 +24,32 @@ import {CompositeLayer} from '@deck.gl/core';
 import {TileLayer as DeckGLTileLayer} from '@deck.gl/experimental-layers';
 import dataProcessor from 'processors';
 import {memoize} from 'utils/utils';
+import { mergeDomain } from 'utils/domain-utils';
  
 export default class SharedstreetsLayer extends CompositeLayer {
   initializeState() {
     this.setState({
-      oldLayerDataMaps: new Map()
+      oldLayerDataMaps: new Map(),
+      // stores domains for each encoded field and for each z-level 
+      layerDomains: new Map(),
+      // stores domains for each encoded field and for each z-x-y tile
+      tileDomains: new Map()
     })
   }
   
   shouldUpdateState({oldProps, props, context, changeFlags}) {
     return changeFlags.propsOrDataChanged || changeFlags.stateChanged;
+  }
+
+  getPickingInfo({info, sourceLayer}) {
+    // info.sourceLayer is the deck gl layer that renders data.
+    const idx = info.sourceLayer.props.idx;
+    const tile = info.tile;
+    const {allData, fields} = tile._data;
+    info.allData = allData;
+    info.fields = fields;
+    info.idx = idx;
+    return info;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
```diff
@@ -24,18 +24,34 @@ import {CompositeLayer} from '@deck.gl/core';
 import {TileLayer as DeckGLTileLayer} from '@deck.gl/experimental-layers';
 import dataProcessor from 'processors';
 import {memoize} from 'utils/utils';
+import { mergeDomain } from 'utils/domain-utils';
```
heshan0131: In kepler.gl's .prettierrc brack spacing is set to false.

src/utils/domain-utils.js
```diff
@@ -0,0 +1,36 @@
+import { ALL_FIELD_TYPES } from "constants";
+import { unique, notNullorUndefined } from "utils/data-utils";
+
+export function mergeDomain(oldDomain, newDomain, fieldType) {
+  switch (fieldType) {
+    case ALL_FIELD_TYPES.real:
+    case ALL_FIELD_TYPES.integer:
+    case ALL_FIELD_TYPES.timestamp:
+      return mergeNumericFieldDomain(oldDomain, newDomain);
+    case ALL_FIELD_TYPES.boolean:
```
heshan0131: boolean field domain is always true or false, so you can just return the `oldDomain` to avoid triggering `updateTriggers`

src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
```diff
@@ -116,13 +156,69 @@ export default class SharedstreetsLayer extends CompositeLayer {
...
+   * @param {*} data data that's loaded in the tile
+   */
+  getTileDomain(layer, tileId, fieldName, channel, data) {
+    const {id: layerId} = layer;
+    const keyObj = {
+      layerId,
+      fieldName,
+      tileId
+    };
+    const key = JSON.stringify(keyObj, Object.keys(keyObj).sort());
```
Pessimistress: Make the filter array a constant

test/node/utils/domain-utils-test.js
```diff
@@ -0,0 +1,73 @@
...
+  mergeDomain
+} from 'utils/domain-utils';
+import {
+  ALL_FIELD_TYPES
+} from 'index';
+
+test('DomainUtils -> mergeDomain', t => {
+  const numericFieldTypes = [ALL_FIELD_TYPES.real, ALL_FIELD_TYPES.integer, ALL_FIELD_TYPES.timestamp];
+  for (let i = 0; i < numericFieldTypes.length; i++) {
+    t.deepEqual(
```
Pessimistress: Use t.is to check if it's indeed the same object

src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
```diff
@@ -57,22 +73,46 @@ export default class SharedstreetsLayer extends CompositeLayer {
...
+          fields: allDataFields,
+          filteredIndexForDomain: Object.keys(allData)
+        };  
+        const visualChannels = Object.values(layer.visualChannels);
+        for (let j = 0; j < visualChannels.length; j++) {
+          const channel = visualChannels[j];
+          const field = layer.config[channel.field];
+          if (field) {
+            const newDomain = this.getTileDomain(layer, tileId, field.name, channel, data);
+            this.updateLayerDomain(layer, z, field, channel, newDomain);
```
Pessimistress: Ideally we can avoid updating a layer domain if this tile has already been merged? This can get pretty expensive depending on the type of the field.

src/deckgl-layers/sharedstreets-layer/sharedstreets-layer.js
starry97: Filter array? I didn't use a filter array here. Can you say more about it?

Pessimistress: The second argument (`Object.keys(keyObj).sort()`) is a constant. Don't recalculate it every time.

mojodna:
How is data from the PBFs surfaced in the histogram widget? Is it additive according to data that's been loaded from tiles? Do data points get dropped when tiles are removed from cache?

I was looking into visualizing data loaded from MVTs last week and dropped it when it looked to me that Kepler _really_ wanted to deal with entire datasets at a time. In that case, I wanted the histogram widget to be reflective of what was visible in the viewport since it wasn't possible to load the whole dataset at once and partial data would be confusing.

Partial data being confusing is what I'm wondering about here. Maybe not for Shared Streets data, but for future data sources that follow this approach.

PR #317, opened 2019-01-03
[Enhancement] Upgrade type-analyzer to pass 0/1 as integer

heshan0131:


heshan0131:
https://github.com/uber/kepler.gl/issues/124

Issue #307, opened 2018-12-10
Notification UI/UX


macrigiuseppe:
**Is your feature request related to a problem? Please describe.**
The current state of Kepler doesn't provide a standard way of displaying errors and it's difficult for users to understand when something is not working the way it should be and for developers to extend UI in a cohesive and consistent way as part of the workflow.

**Describe the solution you'd like**
If we consider Kepler.gl workflow we have the main steps:
- Pick source and load data.
- Create your visualization.
- Since Kepler.gl can also be consumed as library, the notification feature should be flexible enough to accommodate extensions/plugins.

Given the above use cases, Kepler.gl should provide flexible enough set of APIs to support multiple areas/topics out of the box, e.g. file/url/sharing and in more details
-- Ability to register new topics (maybe this is not needed, we can automatically detect the topic field from the notification entry)
-- Ability to register new notifications
-- Ability to remove notifications

**Implementation**
This section will describe the implementation of the solution in details and will cover new APIs and their functionalities.

We can start form the the storage of the notifications
The new feature should provide the ability to:
- Register notifications through API by calling a method, e.g. __registerNotification__, and providing the notification type, e.g. error, and which topic of the app the message is targeting hence we need to develop a little bit more on the second parameter and understand how this will be used to address the UI/UX. The register notification api will return the id of the newly generated entry
- Remove notifications: the user can call at any time the removeNotification and pass the notification id to remove the notification from the screen.

***Storing solution***
We are going to take advantage of the uiState reducer and create a property object called __notifications__. Within the new notifications property we are going create an entry for each part of the application we want to target, e.g. file. The final result should look something like the following snippet:

```javascript
const reducer = {
      uiState: {
        notifications: {
          file: [
            {
              uuid: 123456,
              type: 'warning',
              message: 'Your file contains invalid rows',
              topic: 'file'
            }
          ],
          url: [
            {
              uuid: 678901,
              type: 'success',
              message: 'Your url is has been loaded correctly',
              topic: 'url'
            }
          ],
          // this is generated by a consumer app
          sharing: [
            {
              uuid: 567890,
              type: 'error',
              message: 'The application is not able to upload the current map onto your dropbox account'
              topic: 'sharing'
            }
          ]
        }
      }
    };
```
***Register a new notification***
Kepler.gl will provide a new API (action/helper) to register a new notification, the input of the new API will be an object representing the notification.  In order to make the creation of new notifications we are going to provide helper methods
The signature of said methods are the followings:

[...]

PR #333, opened 2019-01-16
[Feature] Notification systems with new UI panel and helpers to generate messages

macrigiuseppe:
This pull request is a follow up to RFC #307 .

New features:
- Ability to create notifications using helpers
  - __createNotification__
  - __successNotfication__
  - __errorNotfication__
- React components can now dispatch two new actions:
  - __addNotification({message: 'this is a message', type: 'error', topic: 'global, ...})__
  - __removeNotification(notificationId)__
- Two new React components
  - NotificationPanel: container and list of notifications. This component receives the uiState with all notifications and it will only display notifications with the topic set to 'global'
  - NotificationPanelItem: component to display notification message. The new item supports markdown messages.

Both components are injected using dependency injection hence they are also replaceable by using custom components

kepler-notification


src/components/notification-panel/notification-item.js
```diff
@@ -0,0 +1,123 @@
...
+    render() {
+      const {notification, removeNotification} = this.props;
+      return (
+        <NotificationItemContent
+          className={`notification-item`}
+          {...this.props}
+          onClick={() => this.setState({isExpanded: !this.state.isExpanded})}
+          isExpanded={this.state.isExpanded}>
+          <NotificationIcon
+            className={`notification-item--icon`}>
```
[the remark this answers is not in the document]
macrigiuseppe: yeah, i actually had a variable before in there. I will change it

src/components/notification-panel/notification-item.js
```diff
@@ -0,0 +1,123 @@
...
+      this.state = {
+        isExpanded: false
+      };
+    }
+
+    render() {
+      const {notification, removeNotification} = this.props;
+      return (
+        <NotificationItemContent
+          className={`notification-item`}
```
macrigiuseppe: use regular quotes

src/components/notification-panel/notification-item.js
```diff
@@ -0,0 +1,123 @@
...
+    props.theme.notificationPanelItemHeight * (1 + Number(props.isExpanded)) 
+  }px;
+  font-size: 10px;
+  margin-bottom: 1rem;
+  padding: 1em;
+  -webkit-border-radius: 4px;
+  -moz-border-radius: 4px;
+  border-radius: 4px;
+  -webkit-box-shadow: ${props => props.theme.boxShadow};
+  -moz-box-shadow: ${props => props.theme.boxShadow};
```
CharlieMartell: afaik the styled-components autoprefixer should handle adding these

macrigiuseppe: thank you

src/reducers/ui-state-updaters.js
```diff
@@ -164,3 +165,16 @@ export const setExportDataUpdater = (state, action) => ({
     data: !state.exportData.data
   }
 });
+
+export const addNotificationUpdater = (state, {payload}) => ({
+  ...state,
+  notifications: [
+    ...state.notifications,
```
heshan0131: `...(state.notifications || [])`

src/components/kepler-gl.js
```diff
@@ -279,6 +287,9 @@ function KeplerGlFactory(
               this.root = node;
             }}
           >
+            {uiState.notifications.length !== 0 ?
```
heshan0131: even if `uiState.notifications.length === 0` `NotificationPanel` should still work right? I would profer apply less logic here and just pass `uiState.notifications` to `< NotificationPanel>` to let it figure out what to do.

src/components/kepler-gl.js
```diff
@@ -204,6 +207,11 @@ function KeplerGlFactory(
         clicked
       } = visState;
 
+      const notificationPanelFields = {
+        uiStateActions,
+        uiState
```
heshan0131: If you are not using other properties in `uiState`, should just pass in `uiState.notifications` here, making the `notifications -> notificationPanel` logic simple and clear

src/reducers/ui-state-updaters.js
macrigiuseppe: notifications has a default value in the reducer. Do you still want to apply the OR statement?

src/components/kepler-gl.js
macrigiuseppe: I have been thinking about this, in fact the first version had only notifications. I thought that passing the full uiState would provide more information and context about the current state of the UI. Currently we only use notifications but maybe in the future we could take advantage of other properties to change the notification panel behavior

examples/demo-app/src/app.js
```diff
@@ -147,6 +150,26 @@ class App extends Component {
     window.localStorage.setItem(BannerKey, 'true');
   };
 
+  /* eslint-disable max-nested-callbacks */
+  _loadMockNotifications = () => {
+    setTimeout(() => {
```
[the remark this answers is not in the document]
macrigiuseppe: BTW this is just for showing an example of notifications. I will do it, i really like the approach

src/components/kepler-gl.js
macrigiuseppe: done

src/components/kepler-gl.js
macrigiuseppe: Let's only use notificaitons and removeNotification as action

PR #350, opened 2019-01-30
[Feature] Collapsible layer group

heshan0131:
- Add collapsible layer group
- Default state to be collapsed

collapsible-layer-group


heshan0131:
https://github.com/uber/kepler.gl/issues/340

macrigiuseppe:
It looks good. 
Only one comment, I am not sure the three point icon is the best choice. Usually that icon is not used for expanding/collapsing but more as action menu. I have seen the same icon used to describe a drag&drop action.

PR #361, opened 2019-02-15
[Bug] Fix missing 3d building layer in image export

heshan0131:
- pass mapStyle to plot container which containers `threeDBuildingColor`


heshan0131:
* pass mapStyle to plot container which containers `threeDBuildingColor`

PR #363, opened 2019-02-17
[WIP] Upgraded to react16/redux4 and react router 4 for example and websites

macrigiuseppe:
Upgrading to React16 and redux4


macrigiuseppe:
-

PR #366, opened 2019-02-21
[Feat] Single map page

macrigiuseppe:
This diff provides the feature to export current Kepler.gl Map onto a single html map with both data and configuration along with necessary javascript dependency.

single-page-map

This is one of the possible UX solution.

### Alternative
We could merge some of the export options into a single modal and use tabs or other UX elements to switch between different export capabilities.

### DIFF
--- a/src/components/modal-container.js
+++ b/src/components/modal-container.js
@@ -36,7 +36,10 @@ import ExportImageModalFactory from './modals/export-image-modal';
 import ExportDataModalFactory from './modals/export-data-modal';
 import ExportConfigModalFactory from './modals/export-config-modal';
 import AddMapStyleModalFactory from './modals/add-map-style-modal';
+import ExportMapModalFactory from './modals/export-map-modal';
 
+// Template
+import {exportMapToHTML} from 'templates/export-map';
 import {
   ADD_DATA_ID,
   DATA_TABLE_ID,
@@ -48,6 +51,7 @@ import {
   EXPORT_CONFIG_ID,
   ADD_MAP_STYLE_ID
 } from 'constants/default-settings';
+import {EXPORT_MAP_ID} from '../constants/default-settings';
 
 const DataTableModalStyle = css`
   height: 85%;
@@ -73,6 +77,7 @@ ModalContainerFactory.deps = [
   ExportImageModalFactory,
   ExportDataModalFactory,
   ExportConfigModalFactory,
+  ExportMapModalFactory,
   AddMapStyleModalFactory
 ];
 
@@ -83,6 +88,7 @@ export default function ModalContainerFactory(
   ExportImageModal,
   ExportDataModal,
   ExportConfigModal,
+  ExportMapModal,
   AddMapStyleModal
 ) {
   class ModalWrapper extends Component {
@@ -180,6 +186,19 @@ export default function ModalContainerFactory(
       );
 
       this._closeModal();
+    };
+
+    _onExportMap = () => {
+      // we are saving both data and config together
+      // TODO: storing a large amount of data in html could be a limitation
+      // but it will work for now as first version
+      const dump = KeplerGlSchema.save(this.props);
+
+      this._downloadFile(
+        exportMapToHTML(dump),
+        'text/html',
+        'kepler.gl.html'
+      );
     };
 
     /* eslint-disable complexity */
@@ -290,7 +309,6 @@ export default function ModalContainerFactory(
             break;
 
           case EXPORT_DATA_ID:
-
             template = (
               <ExportDataModal
                 {...uiState.exportData}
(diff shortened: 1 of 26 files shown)


macrigiuseppe:
The following screenshot will replace the current one in documentation

k-save-and-export-1

And the following one will be added as part of the new feature

screen shot 2019-02-21 at 4 04 19 pm

macrigiuseppe:
New version with input field to provide mapbox access token

single-page-map

src/components/modal-container.js
```diff
@@ -48,6 +51,7 @@ import {
   EXPORT_CONFIG_ID,
   ADD_MAP_STYLE_ID
```
heshan0131: you can add `EXPORT_MAP_ID` in the import above

src/components/modals/export-map-modal.js
```diff
@@ -0,0 +1,96 @@
...
+  exportHtml,
+  onExportMapboxAccessToken
+}) => (
+  <StyledExportMapModal>
+    <div className="export-map-modal">
+      <div className="title">
+        Export your map into a single html file
+      </div>
+      <div className="subtitle">
+        You will be able to save your configuration and data into a single file and download it on your device.
```
heshan0131: You will be able to save you __map__ into an interactive html file.

src/components/modals/export-map-modal.js
```diff
@@ -0,0 +1,96 @@
...
+      </div>
+
+      <StyledSection>
+        <div>
+          Please provide your Mapbox access token
```
heshan0131: Maybe add a link to https://docs.mapbox.com/help/how-mapbox-works/access-tokens/

src/components/side-panel.js
```diff
@@ -39,6 +39,7 @@ import {
   EXPORT_CONFIG_ID,
   PANELS
```
heshan0131: add `EXPORT_MAP_ID` here

src/components/modals/export-map-modal.js
macrigiuseppe: screen shot 2019-02-22 at 2 46 55 pm

macrigiuseppe:
@heshan0131 updated with your feedback

src/templates/export-map.js
```diff
@@ -0,0 +1,142 @@
...
+      <!-- You can also specify a specific versions by doing the following -->
+      <!--<script src="https://unpkg.com/kepler.gl@0.2.2/umd/keplergl.min.js"></script>-->
+    
+      <!--MapBox token-->
+      <script>
+        /**
+         * Provide your MapBox Token
+         * This will be used in app.js (imported later)
+         * */
+        const MAPBOX_TOKEN = '${options.mapboxApiAccessToken || 'PROVIDE_MAPBOX_TOKEN'}';
```
heshan0131: Talked to Chris about this, maybe we should create a unique token for exported html maps, and rotate this token every month. This way we have a better way to tracking how many map views are from exported html maps.

chrisirhc: For tracking, you may also want to include a version number with the approach in #56 .

macrigiuseppe: > For tracking, you may also want to include a version number with the approach in #56 .
> 
> When we do rotate token, would that break their exported HTML? We may want to inform the users somehow that we've included a default token but expect them to change it if they need it to work as long as they need it to (via their own token). It would be great if users can swap it with their own respective token if they needed to.

@chrisirhc 
We currently allow users to provide their own Mapbox Token as they export the map. I can update the interaction and provide a more suitable message for this new use case.

When you mention "it would be great if users can swap it with their own....", are you referring at any given time after they download the html file? or just when they initially export their maps?

chrisirhc: > When you mention "it would be great if users can swap it with their own....", are you referring at any given time after they download the html file? or just when they initially export their maps?

Consider thinking from the user's point of view: Since the Mapbox token rotates, the user would need to take certain actions if the Mapbox token we provided expired, after the html is exported.
* Would it be obvious to them what action they need to take? (Does the user see an error? Does the error give them information on how to resolve the issue?)
* What steps are required by the user to resolve the issue?
* Are those steps difficult?

macrigiuseppe: Got it.

Currently in Kepler.gl we simply show a log error message: "Mapbox Token is required but nothing in the UI".

Before we implement checks and suggestions in this feature, we should focus on kepler.gl framework messaging by implementing the following:
- Show a notification error if Mapbox Token is missing
- As part of the message we can include a permalink to the user guide where we explain how to update the token when it is expired.

The above changes will give us more flexibility in providing steps and solutions it will also improve the general kepler.gl experience.

As part of the "Export map experience":
- We can always add a default mapbox token with the ability to override it when they export the map.
- When we rotate the token, the mapbox error notification will provide the link with instructions on how to update their token

chrisirhc: Sounds good to me. Thanks 👍

macrigiuseppe: I created a new issue to track the necessary work to add an error notification when mapbox token is missing: https://github.com/uber/kepler.gl/issues/426

macrigiuseppe:
Kepler.gl will provide a mapbox token* so users can export their maps.

* The token will be periodically rotated therefore users must either use their own or follow instructions to gather a newer token from Kepler.gl user guide

macrigiuseppe:
Added a link to help users update existing html file and replace tokens

Screen Shot 2019-03-15 at 12 18 20 PM

Issue #293, opened 2018-11-21
Vector tiles support


LHdT:
Hello,

I'm trying to use kepler as a viewer for my new application.
I've been scanning your redux app structure and I eventually realised there’s no straightforward way to create a custom reducer to load PBF vector tile layers.
I've also checked ongoing developments and found few commits working with pbf load on your git repositories as:
- 00905ccda44ddfb21f0016622a1bc0c146000183
- 50d53c9080aaf8169fa4818a6d7813aba8a2a8db
In order to know the actual status of an eventual pbf loader in kepler.gl, I'd kindly ask for the roadmap for this development and if there will be a future version supporting this functionality on kepler.
If this is the case, just know if there is any estimated release date for that.

Kind regards and thank you all.

heshan0131:
Does your vector tile come from a backend service? Or a static file that you manually upload

LHdT:
It comes from a backend service to be rendered on the map.

LHdT:
Hello Shan, We did on first approach by converting vectortiles to geoJson and then load it into map by this code

![imagen](https://user-images.githubusercontent.com/13876611/49357929-1e322280-f6d1-11e8-98de-722751c52282.png)
![imagen](https://user-images.githubusercontent.com/13876611/49357933-225e4000-f6d1-11e8-84c2-a903b1912ff8.png)
![imagen](https://user-images.githubusercontent.com/13876611/49357938-27bb8a80-f6d1-11e8-87de-336ea235e56b.png)
the main problem of this approach is performance, even though I tried to mitigate it through memorization
![imagen](https://user-images.githubusercontent.com/13876611/49358198-1c1c9380-f6d2-11e8-92ad-d68010424e80.png)
On the other hand, in order to run this vector tiles inside your project we modified it a little
https://github.com/LHdT/kepler-vt.git
as you can see using vector tiles this way means a lack of performance ,as any time I change perspective it starts downloading and converting data.
¿Did you now any other Path we'd follow in order to increase app performance?

Kind Regards and thank you very much.

heshan0131:
@starry97 is working on loading vectile data and rendering it in kepler.gl.
You can follow her PR and branch
https://github.com/uber/kepler.gl/tree/tile-layer
https://github.com/uber/kepler.gl/pull/305

macrigiuseppe:
@LHdT check out [Unfolded studio](https://studio.unfolded.ai/) for vector tiles support and much more.

Issue #268, opened 2018-10-27
Access to Mapbox Object


treyhakanson:
I would like to have a reference to kepler's underlying Mapbox object so I can implement additional functionality. Is there currently a way to get access to the object via something similar to `react-map-gl`'s `getMap` method?

heshan0131:
close as #372 is merged into `^0.2.4`

PR #372, opened 2019-02-26
[Feat] Add getMapboxRef prop

robert-claypool:
Adds a `getMapboxRef` prop which you can set to your custom function. That function is passed down into KeplerGl's MapContainer and is used as a callback to expose the MapContainer's `react-map-gl` [InteractiveMap](https://uber.github.io/react-map-gl/#/Documentation/api-reference/interactive-map) instance. In other words, `getMapboxRef` gives KeplerGl's parent component a reference to the inner Mapbox map.

Is this a good pattern or an anti-pattern? See my update to `examples/demo-app/src/app.js` on how it would be used.

See #268. Exposing access to the full [Mapbox GL API](https://docs.mapbox.com/mapbox-gl-js/api/) should also open a solution for #293 and #118.

Kepler.gl has powerful client-side visualizations for GeoJSON. By having access to the Mapbox GL API, developers can easily add [raster](https://docs.mapbox.com/mapbox-gl-js/example/map-tiles/) and vector layers from a variety of sources, and further customize map interactions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,16 @@ version displayed in side panel header
 
 Action called when click Save Map Url in side panel header.
 
+##### `getMapboxRef(mapbox, index)` (Function, optional)
+
+- Default: `undefined`
+
+Function called when `KeplerGL` adds or removes a `MapContainer` component having an inner Mapbox map.
+
+The `mapbox` argument is an [`InteractiveMap`](https://uber.github.io/react-map-gl/#/Documentation/api-reference/interactive-map) when added or `null` when removed.
+
+The `index` argument is 0 for a single map or 1 for an additional map (since `KeplerGL` supports an optional split map view).
+
 ##### `actions` (Object, optional)
 
 - Default: `{}`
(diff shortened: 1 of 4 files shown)


robert-claypool:
I believe the build failure is unrelated to my diff.

robert-claypool:
@heshan0131, thank you for the review.

I believe 5975a95f6 addresses your concerns... from the extended commit message:
> This simplifies the handling of getMapboxRef so that KeplerGL forwards your callback function to MapContainer unchanged. It also adds support for multiple maps. Use 'index' (0 or 1) to identify which map triggered the callback. If 'mapbox' is null, then that map has been removed by toggleSplitMap.

robert-claypool:
README updated

heshan0131:
Thanks, this looks good to me. I am working with the FOSSA team to fix the build, but we can land this now if needed

robert-claypool:
No hurry, I'll use my fork until version v0.2.3 lands.

robert-claypool:
The force-push was to rebase `robert-claypool/expose-mapbox-map` against `kepler.gl/master`.

Issue #355, opened 2019-02-08
Kepler.gl won't run unless user exports mabox token in the environment


macrigiuseppe:
**Describe the bug**
When using Kepler.gl in an environment which doesn't export the mapbox token we get the following error

Can't read env of null. The problem is related to the following file and line:
3d-buildings-utils:27
```
const MAPBOX_TOKEN = process.env.MapboxAccessToken;
```
**To Reproduce**
In order to reproduce the issue
1. Change the demo app and pass the mapboxtoken hardcoded to kepler.gl. This will simulate a use case without the process object
2. Run the demo app
3. Open the console and you should see the error

**Expected behavior**
The internal part of the library should not rely on external variable, such as the process object, and instead the mapbox token should be propagated from the kepelr.gl component to the rest of the components.
This is a blocker for single page export, since there is no process object.
We need to define a single point where users provide their own mapbox token.

macrigiuseppe:
Possible alternatives:
- Pass mapboxAccessToken as a paramenter and bind getTileData with mapboxAccessToken when calling renderLayers()
- Create a mapbox client where ew can funnel all mapbox requests into and inject mapboxAccessToken

heshan0131:
https://github.com/uber/kepler.gl/pull/360

macrigiuseppe:
fixed

Issue #353, opened 2019-02-08
Export Image Broken


sTranaeus:
**Describe the bug**
Whenever I click Export Image, Kepler freezes and I have to close the window and reopen it. This happens every time I have tried to export an image for the past 1-2 weeks. When I inspect the webpage on Chrome, there is a re-occuring `bundle.js:79 Uncaught TypeError: Cannot read property '3d building' of undefined` error after clicking Export Image.

**To Reproduce**
1. Create anything on Kepler
2. Click Export Image
3. Kepler should freeze and become unresponsive.

**Expected behavior**
Expected the Export Image window to appear as it used to - which would allow me to configure some image settings and then export it.

**Desktop (please complete the following information):**
 - OS: OS X 10.14.2
 - Browser: Chrome
 - Version: 71.0.3578.98

wonderino:
I also experienced this situation after updating to v0.2.1

anapastore:
Similar issues with Safari and Firefox (Ubuntu and OS)

heshan0131:
https://github.com/uber/kepler.gl/pull/360

heshan0131:
https://github.com/uber/kepler.gl/pull/361

macrigiuseppe:
fixed

anapastore:
Still not properly working. It saves an empty figure.

macrigiuseppe:
> Still not properly working in Safari. It saves an empty figure.

@anapastore can you please open a new issue? thank you

Issue #326, opened 2019-01-13
Better image export


macrigiuseppe:
**Describe the bug**
The current export functionality requires to load css and it's trigger a CORS request and it fails.

**To Reproduce**
1. Create a map
2. Try to export the map

**Expected behavior**
Export image should work properly

Multiple tasks were filed against this issue:
#212 
#155 
#150 

**Possible Implementation**
- Use [html-screen-capture-js](https://www.npmjs.com/package/html-screen-capture-js) to generate image (rulesToAddToDocStyle for ratio)

heshan0131:
https://github.com/uber/kepler.gl/pull/360

Issue #212, opened 2018-08-23
Image export seems broken


schnerd:
**Describe the bug**
When using **Share** -> **Export Image**, the image never finishes loading. The following errors appear in the developer console:

Chrome
![image](https://user-images.githubusercontent.com/875591/44547652-3669cd00-a6f2-11e8-8a1b-c820bf251131.png)

Safari
![image](https://user-images.githubusercontent.com/875591/44547722-5ef1c700-a6f2-11e8-9074-89cfc4e2e621.png)

**To Reproduce**
1. Go to http://kepler.gl/#/demo
2. Click on 'Share'
3. Click on 'Export Image'

schnerd:
Hmmm actually I just found #150. Disabling cache in chrome fixed the issue.

Is there anyway we can force a cache invalidation for that resource in cloudfront?

heshan0131:
https://github.com/uber/kepler.gl/pull/360

PR #418, opened 2019-03-01
[Bug] Fix image export component failed to render

heshan0131:
- Load uber front css from kepler.gl s3 bucket
- Move Selection Button out from ExportImageModal, it's causing the entire component fail to render on prod due to minification


heshan0131:
#353 #150 #326 #212 #155 #355

Fixing it for good this time!

Issue #364, opened 2019-02-18
Couldn't use with React 16.4+ due to styled-components deps on componentWillMount


VigneshPeriasami:
**Bug Description**
`styled-components` ThemeProvider v2.2.4 relies on componentWillMount which on react 16.4+ results in broadcast undefined error

`styled-components` issue link related to lifecycle dependencies. https://github.com/styled-components/styled-components/issues/1575

**styled-components ThemeProvider LineOfError**
Using KeplerGL component in React 16.4+ breaks at ThemeProvider broadcast.subscribe
https://github.com/styled-components/styled-components/blob/v2.2.4/src/models/ThemeProvider.js#L66

**To Reproduce**
Use KeplerGL component in react v16.4+

**Expected behavior**
KeplerGL component to render in React v16.4+

VigneshPeriasami:
I can see the fix is already in progress, this will do https://github.com/uber/kepler.gl/pull/363

macrigiuseppe:
> I can see the fix is already in progress, this will do #363

This should be fixed soon

macrigiuseppe:
@VigneshPeriasami have a look at this pull request https://github.com/uber/kepler.gl/pull/423

VigneshPeriasami:
@macrigiuseppe looks good to me, there are few more components under `common` which depends on `innerRef` deprecated prop from `styled-components` while back I have created a forked copy to fix issues that I was running into.

Thanks, really appreciate the efforts.

**innerRef dependent components**
https://github.com/VigneshPeriasami/kepler.gl/commit/9de55a9a9342261a87b13fa28ee46a6af8d09cff

PR #423, opened 2019-03-04
[Chore] Upgraded libraries: react, styled-components

macrigiuseppe:
This pull request will upgrade the following libraries react, styled-components and other needed libraries to transition to react 16

Changes
- react: 16.8.4
- react-dom: 16.8.4
- Styled-components: 4.1.3
- Styled-components is now a peer dependencies.

** This pull request will bring a breaking changes:
- styled-components as peer dependency
- react 15 is no longer supported


src/components/common/range-brush.js
```diff
@@ -120,9 +120,12 @@ export default class RangeBrush extends Component {
...
 
   render() {
-    return <StyledG className="kg-range-slider__brush"
-                    innerRef={comp => {
-      this.rootContainer = comp;
-    }}/>;
+    return (
+      <StyledG
+        className="kg-range-slider__brush"
+        innerRef={comp => {
```
VigneshPeriasami: `innerRef` has been deprecated in `styled-components` v4 we could just use ref now.

https://www.styled-components.com/docs/api#deprecated-innerref-prop

src/components/side-panel/map-style-panel/map-style-selector.js
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
+import styled from "styled-components";import React from 'react';
```
heshan0131: new line

heshan0131:
This should trigger a minor version bump to => 0.3.0

package.json
```diff
@@ -102,17 +102,16 @@
...
     "react-redux": "^4.0.0",
-    "react-tooltip": "https://registry.npmjs.org/react-tooltip/-/react-tooltip-3.3.0.tgz",
+    "react-tooltip": "^3.9.2",
     "react-vis": "^1.8.0",
     "redux": "^3.0.4",
     "redux-actions": "^2.2.1",
     "reselect": "^3.0.1",
-    "styled-components": "^2.2.4",
     "supercluster": "^3.0.0",
-    "type-analyzer": "https://registry.npmjs.org/type-analyzer/-/type-analyzer-0.2.1.tgz",
```
heshan0131: the reason url is specified here is because there is also `unpm` version of these libraries. And unpm install will find those. Not sure if it is still the case. But I would preview leave the `url` here, so they don't break things

chrisirhc: I believe it's no longer the case. Just checked via doing `npm info type-analyzer`

package.json
```diff
@@ -190,8 +189,9 @@
     "webpack-stats-plugin": "^0.2.1"
   },
   "peerDependencies": {
-    "react": "0.14.x - 16.x",
-    "react-dom": "0.14.x - 16.x"
+    "react": "16.x",
+    "react-dom": "16.x",
```
chrisirhc: Have we considered the impact of dropping support for React 15?

macrigiuseppe: ** This is a breaking change.
We are no longer supporting react 15 consumer apps

chrisirhc:
I saw in https://github.com/uber/kepler.gl/issues/364#issuecomment-469166853 that innerRef is being deprecated?

macrigiuseppe:
> I saw in [#364 (comment)](https://github.com/uber/kepler.gl/issues/364#issuecomment-469166853) that innerRef is being deprecated?

Yeah I saw that.

Styled-components is deprecating innerRef. My change is adding one innerRef to a non styled-component element (no issue).

React 16 still supports innerRef but has also taken a completely different approach which requires the following changes: https://reactjs.org/docs/forwarding-refs.html

I am reserving to deal with Refs in a different pull request

chrisirhc:
> Styled-components is deprecating innerRef. My change is adding one innerRef to a non styled-component element (no issue).

Hm.. not sure I understand this statement. Does that mean it's is an innerRef for a component that we created? `innerRef` isn't a React thing from what I gathered, I thought it was just `ref`.

macrigiuseppe:
> > Styled-components is deprecating innerRef. My change is adding one innerRef to a non styled-component element (no issue).
> 
> Hm.. not sure I understand this statement. Does that mean it's is an `innerRef` for a component that we created? `innerRef` isn't a React thing from what I gathered, I thought it was just `ref`.

The innerRef was used by styled-compoenents. In order to remove any innerRef references I went ahead and replaced them with Refs to make sure we are aligned with stylled-components v4 guidelines.

PR #430, opened 2019-03-06
[Enhancement] Automatically loading custom dependencies when inject custom component factory

heshan0131:
- Automatically provide custom factory dependencies while injecting them
- Create an example demo using custom components

screen shot 2019-03-06 at 1 14 21 pm
[__add a custom sidebar that scales down the side panel, and move it to the edge of the screen__]


heshan0131:
https://github.com/uber/kepler.gl/issues/356

src/components/container.js
```diff
@@ -167,6 +167,13 @@ export function injectComponents(recipes) {
         Console.error(errorMsg.wrongPairType);
         return inj;
       }
+
+      // collect dependencies of custom factories, if there is any.
+      // Add them to the injector
+      const customDependencies = flattenDeps([], recipe[1]);
```
heshan0131: @btford

examples/replace-component/src/actions.js
```diff
@@ -0,0 +1,6 @@
+import {createAction} from 'redux-actions';
+
+export const setMapConfig = createAction(
+  'SET_MAP_CONFIG',
+  payload => payload
```
btford: Does `createAction` require a second arg if the payload formatter is the identity fn?

src/components/container.js
btford: 👍

examples/replace-component/src/actions.js
heshan0131: certain version of redux-actions will throw error if payload creator is undefined.

PR #431, opened 2019-03-07
[Enhancement] Slider: use clientX to calculate delta to support windows IE and Tableau kepler.gl

heshan0131:
- This is to address MouseEvent movementX property not supported in internet explorer
- A fix for kepler.gl tableau plugin

### DIFF
--- a/src/components/common/slider/slider-bar-handle.js
+++ b/src/components/common/slider/slider-bar-handle.js
@@ -23,6 +23,7 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import styled from 'styled-components';
+import {getMouseDeltaX} from './slider-handle';
 
 const StyledSlider = styled.div`
   position: relative;
@@ -77,13 +78,10 @@ export default class SliderHandle extends Component {
 
   mousemove = e => {
     e.preventDefault();
-    if (e.movementX === 0) {
-      const deltaX = e.clientX - this.prevX;
-      this.props.sliderBarListener(deltaX);
-    } else {
-      this.props.sliderBarListener(e.movementX);
-    }
-    this.prevX = e.clientX;
+
+    const {deltaX, prevX} = getMouseDeltaX(e, this.prevX);
+    this.props.valueListener(deltaX);
+    this.prevX = prevX;
   };
 
   handleTouchStart = e => {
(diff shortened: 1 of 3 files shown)


chrisirhc:
Since this is a fix for Tableau and IE, perhaps we should state that in the PR title?

src/components/common/slider/slider-bar-handle.js
```diff
@@ -76,7 +77,13 @@ export default class SliderHandle extends Component {
 
   mousemove = e => {
     e.preventDefault();
-    this.props.sliderBarListener(e.movementX);
+    if (e.movementX === 0) {
+      const deltaX = e.clientX - this.prevX;
```
chrisirhc: Any reason we're using clientX instead of screenX as specified in the spec: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX ?

chrisirhc: Perhaps consider de-duplicating these two implementations to increase code re-use.
This also could read a bit like a React hook, if you re-organize it.

src/components/common/slider/slider-bar-handle.js
```diff
@@ -62,9 +62,10 @@ export default class SliderHandle extends Component {
   state = {mouseOver: false};
   prevX = 0;
 
-  handleMouseDown = () => {
+  handleMouseDown = (e) => {
     document.addEventListener('mouseup', this.mouseup);
     document.addEventListener('mousemove', this.mousemove);
+    this.prevX = e.clientX;
```
akre54: Why store on `this` instead of putting in state?

src/components/common/slider/slider-bar-handle.js
heshan0131: Read about the difference. Looks like they are both fine in our use case. It's unlikely people change their screen resolution, or move the  viewport while dragging.

__screenX and screenY__
Relative to the top left of the physical screen/monitor, this reference point only moves if you increase or decrease the number of monitors or the monitor resolution.

__clientX and clientY__
Relative to the upper left edge of the content area (the viewport) of the browser window. This point does not move even if the user moves a scrollbar from within the browser.

heshan0131:
The build failure is related to fossa, doesn't seem to have anything to do with the code changes

src/components/common/slider/slider-handle.js
```diff
@@ -23,6 +23,7 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import styled from 'styled-components';
+import {getMouseDeltaX} from './slider-handle';
```
chrisirhc: I'm not sure I see this function defined anywhere in this file. Also, it seems to be importing from the same file.

heshan0131:
Asking fossa team on build failure. it is not a merge blocker

PR #432, opened 2019-03-07
[Bug] Range brush width change should not trigger onBrush callback

heshan0131:
Range brush is triggering `setFilter` whenever width changes. Slows down performance significantly during mounting, closing or opening side panel

### DIFF
--- a/src/components/common/range-brush.js
+++ b/src/components/common/range-brush.js
@@ -42,7 +42,6 @@ export default class RangeBrush extends Component {
   };
 
   componentDidMount() {
-    const {range: [min, max], value: [val0, val1]} = this.props;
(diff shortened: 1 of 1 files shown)


akre54:
I had a lot of trouble getting the callbacks to work right between react and d3, since d3 wants to act based on events and updates by firing its own internal events.

This fix was in place so you could actually set the brush back to the full width. Any thoughts on how we can support this otherwise?

heshan0131:
Can you use `_move` instead of `_reset`?

akre54:
I think it also had an infinite loop, but I don't remember.

Looks like d3-brush was refactored since I last used it, so maybe it might be fixed?

PR #433, opened 2019-03-08
[Bug] Range slider: correctly setting ranch brush selection when mount

heshan0131:
- Width change should not trigger this._brush
- Check whether value has changed before triggering props.onBrush
ezgif com-video-to-gif 1

### DIFF
--- a/src/components/common/range-brush.js
+++ b/src/components/common/range-brush.js
@@ -47,7 +47,8 @@ export default class RangeBrush extends Component {
     // and programmatic brushing (brush.move). We need these flags to
     // distinguish between the uses.
     //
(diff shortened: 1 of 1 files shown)


src/components/common/range-brush.js
```diff
@@ -47,7 +47,8 @@ export default class RangeBrush extends Component {
     // and programmatic brushing (brush.move). We need these flags to
     // distinguish between the uses.
     //
-    // We don't use state because that would trigger another `componentDidUpate`
+    // We don't use state because that would trigger another `componentDidUpdate`
```
akre54: 😅

PR #445, opened 2019-03-17
[Chore] Dropbox enhancements

macrigiuseppe:
Changes to Dropbox UX flow and UI


examples/demo-app/src/actions.js
```diff
@@ -377,7 +377,14 @@ export function exportFileToCloud(handlerName = 'dropbox') {
 }
 
 export function setCloudLoginSuccess() {
```
macrigiuseppe: after a successful login it will also start the uploading without user interaction

examples/demo-app/src/constants/default-settings.js
```diff
@@ -94,7 +94,8 @@ export const LOADING_SAMPLE_LIST_ERROR_MESSAGE = 'Not able to load sample galler
 export const LOADING_SAMPLE_ERROR_MESSAGE = 'Not able to load sample';
 export const LOADING_URL_MESSAGE = 'You can use the following formats: CSV | JSON | Kepler.gl config json. Make sure the url contains the file extension.';
 export const CORS_LINK = 'https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS';
-export const KEPLER_DISCLAIMER = '* Kepler.gl will save your map data to your personal cloud storage, only people with the URL can access your map and data.';
+export const KEPLER_DISCLAIMER = '* Kepler.gl will save your map data to your personal cloud storage, only people with the URL can access your map and data. ' +
+  'You can edit/delete the data file in your Dropbox account anytime. If you do not have a Dropbox account, please signup over here';
```
briantcl: > If you do not have a Dropbox account, please signup over here

I will suggest removing this part of the sentence since right now the maps can be uploaded directly after a user has signed-up. So that we can consolidate the user flow into one single button.

src/components/side-panel/panel-header.js
```diff
@@ -195,7 +196,7 @@ export const SaveExportDropdown = ({
 
         {onSaveMap ? (
           <PanelItem
```
heshan0131: You can make each of these `PanelItem` a factory, or the `SaveExportDropdown` can be a factory

examples/demo-app/src/components/sharing/cloud-storage.js
```diff
@@ -113,6 +113,8 @@ const ExportCloudModal = ({
   onCloudLoginSuccess
 }) => {
   const metaUrl = get(info, ['metadata', 'url']);
+  const error = get(info, ['error']);
```
heshan0131: `get` should be added to `demo-app`'s dependency

examples/demo-app/src/app.js
```diff
@@ -336,6 +336,7 @@ class App extends Component {
                   width={width}
                   height={height - (showBanner ? BannerHeight : 0)}
                   onSaveMap={this._isCloudStorageEnabled() && this._toggleCloudModal}
+                  onSaveMapLabel="Share Public URL"
```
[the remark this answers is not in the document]
macrigiuseppe: > the naming of this variable suggested that it is a function, but it's actually a string. should we just call it `SaveMapLabel`

of course

examples/demo-app/src/components/sharing/cloud-storage.js
macrigiuseppe: it's already in demo-app package.json: "lodash.get": "^4.4.2",

examples/demo-app/src/constants/default-settings.js
macrigiuseppe: Sounds good!!

src/components/side-panel/panel-header.js
macrigiuseppe: I will implement this one. 
I went down this path because i wanted to launch dropbox on time but I took a note here
In fact in my diff i put this comment: https://github.com/uber/kepler.gl/pull/445/files#diff-c77a89b67856c85be3f5916723ae00b5R179

macrigiuseppe: SaveExportDropdown is not part of the DI chain. In fact PanelHeader reads ationItems (default prop) which uses SaveExportDropdown as component to render the action menu.
I can see two options here:
1. Since we use actionsItems to include SaveExportDropdown.  ActionItems can be a Dependency injected into panel-header.
2. We use the same approach used in __replace-component__ example to avoid propagating a new prop 4 level deep into the tree and implement a custom PanelHeaderFactory like the following:
```
import React from 'react';
import {PanelHeaderFactory} from 'kepler.gl/components';

// Custom Panel Header renders default panel header, changing its default props
// to avoid rendering any action items on the top right
export function CustomPanelHeaderFactory() {
  const PanelHeader = PanelHeaderFactory();
  const currentActionItem = PanelHeader.defaultProps.actionItems[0];

  PanelHeader.defaultProps = {
    ...PanelHeader.defaultProps,
    actionItems: [
      {
        ...currentActionItem,
        dropdownComponent: (props) => (
          <currentActionItem.dropdownComponent
            {...props}
            saveMapLabel={'Share map URL'}
          />
        )
      }

    ]
  };
  return PanelHeader;
}

export default CustomPanelHeaderFactory;

```

heshan0131: Let's do a combination of option 1 & 2

Make `SaveExportDropdown` a factory, and a dep of PanelHeaderFactory. You can assign it to defaultProps inside the factory
```js
PanelHeaderFactory.deps = [SaveExportDropdownFactory];
function PanelHeaderFactory(SaveExportDropdown) {
  return class PanelHeader extends Component {
    static defaultProps = {
      logoComponent: KeplerGlLogo,
      actionItems: [{
        id: 'save',
        iconComponent: Save,
        onClick: () => {},
        label: 'Share',
        dropdownComponent: SaveExportDropdown
      }]
    };
}
```

Make each of the panel export item a factory, and dep of `SaveExportDropdown`
```js
export const SaveExportDropdownFactory = (ExportImage, ExportData, ExportConfig, ExportMap, SaveMap) => {
  const SaveExportDropdown = (props) => {
    return (
      <StyledPanelDropdown show={show} className="save-export-dropdown">
        <ClickOutsideCloseDropdown className="save-export-dropdown__inner"
          show={show}
          onClose={onClose}>
          <ExportImage
            onClickHandler={props.onExportImage}
            onClose={props.onClose}
          />
          {props.onSaveMap ? (
            <SaveMap
              onClickHandler={props.onSaveMap}
              onClose={props.onClose}
            />
          ) : null}
...
SaveExportDropdownFactory.deps = [
  ExportImageFactory,
  ExportDataFactory,
  ExportConfigFactory,
  ExportMapFactory,
  SaveMapFactory
]
```

Export of the export item factory can be:

```js
export const SaveMapFactory = () => {
  const SaveMap = (props) => (
    <PanelItem {...props}/>
  )
  SaveMap.defaultProps = {
    label: "Save Map",
    icon: Share
  }
  return SaveMap;
}
```

Then in `CustomPanelHeader`, we can create a `CustomSaveMapFactory`
```js
export const CustomSaveMapFactory = () => {
  const CustomSaveMap = SaveMapFactory();
  CustomSaveMap.defaultProps = {
    ...CustomSaveMap.defaultProps,
    label: "Save Dropbox Map"
  }

  return CustomSaveMap;
}
```


Don't forget to export all new factories in `component/index.js`
```js
export {default as PanelHeaderFactory,
  SaveExportDropdownFactory,
  ExportImageFactory,
  ExportDataFactory,
  ExportConfigFactory,
  ExportMapFactory,
  SaveMapFactory
} from './side-panel/panel-header'
```

heshan0131: Code example in this commit
https://github.com/uber/kepler.gl/commit/81ec2f93130d06a82a6e43096feaf5d88196eb35

macrigiuseppe: @heshan0131 Merged changes

PR #451, opened 2019-03-25
[Feat] Gpu data filtering - 1

heshan0131:
Step. 1 GPU Data filtering feature (https://github.com/uber/kepler.gl/issues/328)
- Add `gpu` property to filter. 
- Implement logic to manage which filter is gpu enabled
- Remove filtered data array from filter result (not used anywhere)
- Implement `set` function to immutably edit reducer state

### DIFF
--- a/examples/demo-app/src/app.js
+++ b/examples/demo-app/src/app.js
@@ -115,7 +115,7 @@ class App extends Component {
     // }
 
     // load sample data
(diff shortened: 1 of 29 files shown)


heshan0131:
renamed from `gpu-data-filter` to `gpu-data-filter-1`

Issue #177, opened 2018-07-25
geojson Points can't be visualized as Kepler point layer (only polygon)


stephanietuerk:
First: This is an amazing project! Thank you!

Issue: When I import a geojson file consisting of Point geometry types, Kepler defaults to visualizing this with a polygon layer. If I try to switch the layer to point (so that I can use visualization techniques like hexbin), I can't get it to read the lat/lon coordinates of the Point geometries/display the points. 
Selecting value _geojson in Lat and Lng boxes does not make points appear in map.

I can resolve this problem by transforming geojson back into regular json or csv by extracting lat/lon coords from geometry (using Python), but if geojsons are supported, would be nice for Point geometries to display properly.

MacBook Pro Retina 13-inch early 2015
OS: 10.13.5
Chrome Version 67.0.3396.99 (Official Build) (64-bit)

geojson snippet:
{
"type": "FeatureCollection",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:OGC:1.3:CRS84" } },
"features": [
{ "type": "Feature", "properties": { "geo_precision": 1, "count": 1 }, "geometry": { "type": "Point", "coordinates": [ 60.8944, 32.2632 ] } },
{ "type": "Feature", "properties": { "geo_precision": 1, "count": 1 }, "geometry": { "type": "Point", "coordinates": [ 61.0667, 34.6667 ] } },
{ "type": "Feature", "properties": { "geo_precision": 1, "count": 2 }, "geometry": { "type": "Point", "coordinates": [ 61.23, 31.8261 ] } },
....

sashuk:
First, thanks a lot for such a great tool! :fireworks: 

Concerning the issue: the line objects (GeoJSON layers) are not detected correctly as well, the example below is detected as the polygon layer.

```
{
	"type": "FeatureCollection",
	"features": [{
		"geometry": {
			"type": "LineString",
			"coordinates": [
				[39.248915, -6.961785],
				[39.250374, -6.952925]
			]
		},
		"type": "Feature",
		"properties": {
			"gid": 81,
			"id": null
		}
	}]
}
```

prune998:
same issue here. Any info on this ? Is it sa bug, a lack of feature or we're doing something wrong ?

stephanietuerk:
> Any info on this ? Is it sa bug, a lack of feature or we're doing something wrong ?

[...]

PR #456, opened 2019-03-27
Upgrade deck.gl to 6.4

Pessimistress:
# Change List

- Remove custom picking module, use transparent `highlightColor`
- Remove `hi-precision` from `layer.config.visConfig`
- Handle ScatterplotLayer API change (simulate old behavior until UI updates)
  - Add `filled` to `layer.config.visConfig`
  - Add `strokeColor` `strokeColorRange` to `layer.config.visConfig`
  - Add `strokeColor` to `visualChannels` allow stroke color by selected field
- Handle GeoJsonLayer API change (disable point feature outlines)
  - Add `strokeColor` to `visualChannels` allow stroke color by selected field
  - Allow `strokeColor` to be configured separately from `fillColor`
  - Allow all settings in point layer to be applied in geojson feature type point

Reference: http://deck.gl/#/documentation/overview/upgrade-guide

Screen Shot 2019-05-13 at 12 39 58 PM
**Point ountline by value**

Screen Shot 2019-05-13 at 12 41 16 PM
**Polygon ountline by value**


heshan0131:
#507

heshan0131:
#177

src/components/side-panel/layer-panel/layer-configurator.js
```diff
@@ -109,6 +112,36 @@ export default class LayerConfigurator extends Component {
           </ConfigGroupCollapsibleContent>
         </LayerConfigGroup>
 
+        {/* outline color */}
+          {layer.type === LAYER_TYPES.point ? (
```
macrigiuseppe: i think you have one extra white space here before __{__

src/components/side-panel/layer-panel/layer-configurator.js
```diff
@@ -529,8 +525,12 @@ export default class LayerConfigurator extends Component {
 
     return (
       <StyledLayerVisualConfigurator>
-        {/* Color By */}
-        <LayerConfigGroup label={'color'} collapsible>
+        {/* Fill Color */}
+        {featureTypes.polygon || featureTypes.point? (<LayerConfigGroup
```
macrigiuseppe: if you want to avoid the ternary operator __?__ and the __: null__, you can do simply
```
(featureTypes.polygon || featureTypes.point) && (...)
```

src/schemas/vis-state-schema.js
```diff
@@ -76,7 +77,8 @@ class DimensionFieldSchemaV0 extends Schema {
     };
   }
 
-  load(field, config, accumulated) {
+  load(field, parents, accumulated) {
+    const [config] = parents.slice(-1);
```
macrigiuseppe: could parents be empty?

If empty you may have issues when you access config later on line 83

src/schemas/vis-state-schema.js
```diff
@@ -96,7 +98,8 @@ class DimensionScaleSchemaV0 extends Schema {
   save(scale) {
     return {[this.key]: scale};
   }
-  load(scale, config, accumulated) {
+  load(scale, parents, accumulated) {
+    const [config] = parents.slice(-1);
```
macrigiuseppe: same as before

src/components/side-panel/layer-panel/layer-configurator.js
heshan0131: I think `? : null` has been a pattern in all our components, so I prefer to keep it this way. Also if it's in Component render function `(Boolean(a) && <div/>)` will result in `false` been rendered on the screen when Boolean(a) is falsy

src/schemas/vis-state-schema.js
heshan0131: This is been called recursively from `VisStateSchema -> LayerSchema -> LayerConfigSchema` `parents` contains all the parent node `[visState, layer, config]`, it should never be empty. If it is, it means something is wrong, and I would rather the error been thrown instead of a silent pass.

src/schemas/vis-state-schema.js
heshan0131: Same as explained above

davidwedekind:
The issue seems not to be resolved for the demo app. I tried to visualize a single point. Pulling in the following geojson would create a polygon but not a point:

{
	"type": "FeatureCollection",
	"features": [
		{
			"type": "Feature",
			"geometry": {
				"type": "Point",
				"coordinates": [
					13.404672,
					52.5246
				]
			},
			"properties": {
				"location": "WeWork Hackescher Markt"

			}
		}
	]
}

Am I doing something wrong or has the demo just not upgraded.

Thx,
David

mrcgates:
I am still having this issue, with Kepler.gl v. 1.1.5  and Deck.gl v. 7.1.11. I even added Lat/Lng to properties. Any ideas? Sample geoJson:
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          -114.34458,
          32.87754
        ]
      },
      "properties": {
        "id": "1234",
        "created": 1567002206379,
        "relativeVelocity": 1104,
        "radialVelocity": 24.69,
        "altitude": 98,
        "Lng": -114.34458,
        "Lat": 32.87754
      }
    }
  ]
}

heshan0131:
Geojson format will always result in a Polygon layer. But you have all the configuration options available for the points in a polygon layer as in a point layer

PR #463, opened 2019-04-02
[Enhancement] Improvement to single map export feature

macrigiuseppe:
This pull requests resolves item https://github.com/uber/kepler.gl/issues/446

### DIFF
--- a/examples/umd-client/index.html
+++ b/examples/umd-client/index.html
@@ -18,7 +18,7 @@
   <script src="https://unpkg.com/styled-components@4.1.3/dist/styled-components.min.js" crossorigin></script>
 
   <!-- Load Kepler.gl latest version-->
-  <script src="https://unpkg.com/kepler.gl/umd/keplergl.min.js" crossorigin></script>
+  <script src="https://unpkg.com/kepler.gl@0.2.4/umd/keplergl.min.js" crossorigin></script>
   <!-- You can also specify a specific versions by doing the following -->
   <!--<script src="https://unpkg.com/kepler.gl@0.2.2/umd/keplergl.min.js"></script>-->
 
--- a/examples/umd-client/index.html
+++ b/examples/umd-client/index.html
@@ -18,7 +18,7 @@
   <script src="https://unpkg.com/styled-components@4.1.3/dist/styled-components.min.js" crossorigin></script>
 
   <!-- Load Kepler.gl latest version-->
-  <script src="https://unpkg.com/kepler.gl@0.2.4/umd/keplergl.min.js" crossorigin></script>
+  <script src="https://unpkg.com/kepler.gl/umd/keplergl.min.js" crossorigin></script>
   <!-- You can also specify a specific versions by doing the following -->
   <!--<script src="https://unpkg.com/kepler.gl@0.2.2/umd/keplergl.min.js"></script>-->
 
--- a/src/components/modals/export-map-modal.js
+++ b/src/components/modals/export-map-modal.js
@@ -21,7 +21,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import {GITHUB_EXPORT_HTML_MAP, MAPBOX_ACCESS_TOKEN} from 'constants/user-guides';
+import {GITHUB_EXPORT_HTML_MAP} from 'constants/user-guides';
 
 const StyledExportMapModal = styled.div`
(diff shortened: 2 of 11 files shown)


website/webpack.config.js
```diff
@@ -190,12 +190,6 @@ module.exports = env => {
       logInstruction('You can get the token at https://www.dropbox.com/developers');
       throw new Error('Missing Export DropboxClientId Access token');
     }
-    if (!process.env.MapboxExportToken) {
```
heshan0131: why remove this? The process should exit if toke nare missing

macrigiuseppe: this is actually only optional now. We have decided to not provide a default token until we have more info from Mapbox.

macrigiuseppe: gonna undo based on this: https://github.com/uber/kepler.gl/issues/446

macrigiuseppe:
Updated export map modal dialog
k-save-and-export-5

src/components/modals/export-map-modal.js
```diff
@@ -58,14 +58,17 @@ const StyledInput = styled.input`
   }
 `;
 
+const DISCLAIMER = "* We will provide a temporary token. If you don't provide your own token, " +
+  "map will break. You can later follow this instruction to edit html and input token.";
```
chrisirhc: > If you don't provide your own token, map will break
If you don't provide your own token, the map may fail to display at any time as we rotate our token to avoid misuse of the token. 

If you don't provide your own token, the map may fail to display at any time as we rotate our token to avoid misuse of the token. 

> You can later follow this instruction 

You can change the Mapbox token at any time using the following instructions:

heshan0131: can just say `If you don't provide your own token, the map may fail to display at any time as we rotate ours to avoid misuse`.  shorter and less use of the word `token`

examples/umd-client/index.html
```diff
@@ -18,7 +18,7 @@
   <script src="https://unpkg.com/styled-components@4.1.3/dist/styled-components.min.js" crossorigin></script>
 
   <!-- Load Kepler.gl latest version-->
-  <script src="https://unpkg.com/kepler.gl/umd/keplergl.min.js" crossorigin></script>
+  <script src="https://unpkg.com/kepler.gl@0.2.4/umd/keplergl.min.js" crossorigin></script>
```
heshan0131: why add a specific version here?

src/templates/export-map.js
```diff
@@ -41,7 +41,7 @@ export const exportMapToHTML = options => {
       <script src="https://unpkg.com/react-virtualized@9.21.0/dist/umd/react-virtualized.js" crossorigin></script>
       
       <!-- Load Kepler.gl latest version-->
-      <script src="https://unpkg.com/kepler.gl/umd/keplergl.min.js" crossorigin></script>
+      <script src="https://unpkg.com/kepler.gl@0.2.4/umd/keplergl.min.js" crossorigin></script>
```
heshan0131: same here, why lock to a specific version? This is going to make it hard to manage versions in the future unless we have a deploy script that automatically edits them

examples/umd-client/index.html
macrigiuseppe: This is to make sure in the future if we make breaking changes to Kepler.gl (require newer version of a library) they will not break downloaded maps

src/templates/export-map.js
macrigiuseppe: same story here

macrigiuseppe:
updated

examples/umd-client/index.html
heshan0131: Yea, but this is an example right? For easy following example, we should just use latest version.

src/templates/export-map.js
heshan0131: The exported map should use whatever version the current demo app is on, which should be the latest. You can imagine people use the demo app to create a map and some feature is missing because the exported map is fixed to @0.2.4.

examples/umd-client/index.html
macrigiuseppe: Based on our conversation i will revert this one and look for a better solution as part of a different task

macrigiuseppe:
@chrisirhc @heshan0131 I applied the changes requested, please review!

macrigiuseppe:
The feature developed in this code will be moved into a new issue https://github.com/uber/kepler.gl/issues/482 and pull request later on

Issue #328, opened 2019-01-13
GPU data filtering integration


macrigiuseppe:
**Problem**
Data filtering is currently executed on the CPU in javascript. The logic generates a new array every time it runs, causing deck.gl to invalidate all buffers and recreate them. CPU filtering causes lagging on large datasets, especially during time playback.

**Solution**
We will move dynamic data filtering logic into the GPU side, using customized shader modules by deck.gl (rfc)[https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/v6.0/data-filter-rfc.md. This will not apply to all filters, only the filters that require constant updates: time filter and range slider. 

**Limitation**
1. Only one filter can be GPU accelerated, we need to add logic to determine which filter to be handled by the GPU during runtime. 
    - [Solution] Add GPU filter priority for each filter type. 0: time filter, 1: range filter, 3: multi selection & boolean.
    - [Alternative] Allow multiple filters to be added to GPU filter shader module. 

2. Aggregation layer might not be filtered in GPU. 

### Implementation Steps
1. Add `.gpu` property to filter. Implement logic to manage which filter is gpu enabled
https://github.com/uber/kepler.gl/pull/451 #464 
2. Implement data filtering logic and format layer data logic  #469 
3. Add gpu filter logic to deck.gl layer and shader with [Deck.gl layer extension](https://github.com/uber/deck.gl/pull/2684) #604 #858 
4. Add gpu filter logic to aggregation layers.#858 
5. Generalize base-layer class to incoporate gpu filtering methods 
6. Cover edge cases, address layers can only be cpu filtered (heatmap) #858 
7. Adding layer tests #618 
8. Implement CPU filtering to support export filtered data to csv #890 
9. Upgrade deck.gl to 7.3 Use deck.gl brushing module #901 
7. Benchmarking

#### Testcase 1
<img width="1025" alt="Screen Shot 2019-04-05 at 12 41 44 PM" src="https://user-images.githubusercontent.com/3605556/55652381-3362af00-57a0-11e9-9c3e-5b39e5fb3f48.png">

* Data Size: 680k rows, 
* Number of Layers: 1
* Filter: range filter

**Before**
<img width="633" alt="Screen Shot 2019-04-05 at 12 40 26 PM" src="https://user-images.githubusercontent.com/3605556/55652425-4b3a3300-57a0-11e9-87ad-5aff6b0968a0.png">

**After**
[...]

PR #464, opened 2019-04-02
[Feat] Gpu data filter 2

heshan0131:
Step. 2 GPU Data filtering feature (#328)

- refactor filterData to filterDataset
- filter dataset based on gpu or cpu mode
- set `gpuChannel` in filter and `gpuFilter` in dataset to be passed to layer


src/deckgl-layers/scatterplot-brushing-layer/scatterplot-brushing-layer.js
```diff
@@ -54,9 +53,9 @@ export default class ScatterplotBrushingLayer extends ScatterplotLayer {
   initializeState() {
     super.initializeState();
 
-    this.getAttributeManager().addInstanced({
-      instanceFilterValue: {size: MAX_GPU_FILTERS, accessor: 'getFilterValue'}
-    });
+    // this.getAttributeManager().addInstanced({
```
macrigiuseppe: do you still need this one? otherwise you can delete it

src/layers/layer-update.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable guard-for-in */
+
+/**
+ *
+ * @param {Object} updateTriggers {getPosition: {column}, getData: {filteredIndex}}
+ * @param {Object} oldUpdateTriggers
+ * @returns {}
```
macrigiuseppe: No return type!

src/layers/layer-update.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function diffUpdateTriggers(updateTriggers, oldUpdateTriggers = {}) {
+  const triggerChanged = {};
```
macrigiuseppe: do you need both variables since at the end you are assigning the value of triggerChanged to reason?
Can't you just return triggerChanges?

src/layers/layer-update.js
```diff
@@ -0,0 +1,46 @@
...
+  let reason = false;
```
macrigiuseppe: You shouldn't use false to set a variable which later one you will use to store an object
```
triggerChanged[triggerName] = true;
reason = triggerChanged;
```

I think you can also change the name to reasons since you may have multiple properties in the object

src/layers/layer-update.js
```diff
@@ -0,0 +1,46 @@
...
+
+  for (const triggerName in updateTriggers) {
```
macrigiuseppe: you can rewrite this loop using Object.entries(object) 
```
Object.entries(updateTriggers).reduce((acc, pair) => {
  const {[0]: triggerName, [1]: newTriggers = {}} = pair;
  const oldTriggers = oldUpdateTriggers[triggerName] || {};
  const diffReason = compareUpdateTrigger(newTriggers, oldTriggers, triggerName);
 .....
}, {triggerChanged});
```

src/layers/point-layer/point-layer.js
```diff
@@ -60,6 +61,8 @@ export default class PointLayer extends Layer {
     this.registerVisConfig(pointVisConfigs);
     this.getPosition = memoize(pointPosAccessor, pointPosResolver);
     this.getText = memoize(pointLabelAccessor, pointLabelResolver);
+    // this._oldLayerData = null;
```
macrigiuseppe: remove this comment if not necessary

src/layers/point-layer/point-layer.js
```diff
@@ -138,9 +140,34 @@ export default class PointLayer extends Layer {
...
+    const triggerChanged = diffUpdateTriggers(dataUpdateTriggers, this._oldDataUpdateTriggers);
+    this._oldDataUpdateTriggers = dataUpdateTriggers;
+
+    return triggerChanged;
+  }
+
+  calculateDataAttribute(allData, filteredIndex, getPosition) {
+    const data = [];
+    for (let i = 0; i < filteredIndex.length; i++) {
+      // data = filteredIndex.reduce((accu, index) => {
```
macrigiuseppe: if not necessary delete it

src/layers/point-layer/point-layer.js
```diff
@@ -138,9 +140,34 @@ export default class PointLayer extends Layer {
...
+  getChangedTriggers(dataUpdateTriggers) {
+    const triggerChanged = diffUpdateTriggers(dataUpdateTriggers, this._oldDataUpdateTriggers);
+    this._oldDataUpdateTriggers = dataUpdateTriggers;
+
+    return triggerChanged;
+  }
+
+  calculateDataAttribute(allData, filteredIndex, getPosition) {
+    const data = [];
+    for (let i = 0; i < filteredIndex.length; i++) {
```
macrigiuseppe: you can replace this loop with a reduce statement
```
const data = filteredIndex.reduce((acc, index) => {
...
}, []);
```

src/layers/point-layer/point-layer.js
```diff
@@ -228,6 +239,7 @@ export default class PointLayer extends Layer {
       getColor,
       getRadius,
       getText
+      // getFilterValue
```
macrigiuseppe: comment

src/utils/dataset-utils.js
```diff
@@ -92,16 +92,18 @@ export function createNewDataEntry({info = {}, data}, datasets = {}) {
...
 
+  const allIndexes = allData.map((_, i) => i);
   return {
     [dataId]: {
       ...datasetInfo,
       color: datasetInfo.color || getNewDatasetColor(datasets),
       id: dataId,
       allData,
+      allIndexes,
       // TODO: no need to make a copy anymore, only save fieldedIndex
```
macrigiuseppe: do you still this TODO?

src/utils/filter-utils.js
```diff
@@ -268,110 +272,181 @@ export function updateFilterField(filter, idx, dataset, filters) {
...
+ * Compare filter records to get what has changed
+ * @param {Object} filterRecord
+ * @param {Object} oldFilterRecord
+ * @returns {{dynamicDomain: Object, fixedDomain: Object, cpu: Object, gpu: Object}} changed filters based on type
+ */
+export function diffFilters(filterRecord, oldFilterRecord = {}) {
+  let filterChanged = {};
+
+  /* eslint-disable no-loop-func */
+  for (const record in filterRecord) {
```
macrigiuseppe: if you want you can combine Object.entries and reduce to retrieve keys and values of a given object

src/deckgl-layers/scatterplot-brushing-layer/scatterplot-brushing-layer.js
heshan0131: Need it later to implement deck.gl filter attributes

src/layers/layer-update.js
heshan0131: The return value should be `false` if nothing changed, or `triggerChanged` as an object

src/layers/point-layer/point-layer.js
heshan0131: This is a big loop that goes through the entire data array. `for loop` is more performative than `reduce`. I have decided to always use `for loop` when the data array needs to be iterated

src/layers/point-layer/point-layer.js
heshan0131: 👍

src/utils/dataset-utils.js
heshan0131: deleted

src/utils/filter-utils.js
heshan0131: updated

src/layers/layer-update.js
heshan0131: updated

src/layers/layer-update.js
heshan0131: The argument here is shorter code or code that's easier to understand. I feel like reduce is fancier, but in a function like this I feel like readability is more important

chrisirhc:
Curious, have we tried this out on mobile?

heshan0131:
> Curious, have we tried this out on mobile?
This should be no difference between gpu filtering and brushing. All we are doing is sending additional uniforms and attributes to shader. I tested the brushing example that I built for deck.gl example, works on mobile

Issue #339, opened 2019-01-22
[Feature] Kepler.gl plugin for Tableau


heshan0131:
**Feature**
Using kepler.gl inside tableau

**Solution**
Using Tableau's [Extensions API](https://www.tableau.com/developer/extensions) we can load kepler.gl 

**Issues**
- An efficient way to pass data from tableau to kepler.gl
- A more cohesive styling
  - Allow sidebar layer visual configurations to collapse: [[340]](https://github.com/uber/kepler.gl/issues/340)
  - Solve scrolling conflict
- Allow adding multiple dataset
- Allow passing a custom theme object to change the styling [[374]](https://github.com/uber/kepler.gl/issues/374)

briantcl:
currently blocked by the Tableau team 2019 version

heshan0131:
Waiting for UCF and Tableau to sign legal documents, target publish in 2 weeks

heshan0131:
We finally got all the legal documents for tableau extension! Will target to release it next week!

macrigiuseppe:
Support policy page: https://github.com/keplergl/kepler.gl/pull/724

TimJMartin:
@heshan0131 just wondering if there is any further update on when the plugin might be released. thanks

Hesham-91:
@heshan0131 is this happening anytime soon or does anyone know a way where I can integrate Kepler.gl in Tableau and make interactive with other components of the dashboard?

macrigiuseppe:
we have provided all the necessary documentation and code to tableau. They will review it and include kepler.gl in the extension gallery.

carstonhernke:
Update to those finding this thread: the kepler.gl Tableau plugin appears to now be available on the Tableau Extension Gallery. [https://extensiongallery.tableau.com/products/108](https://extensiongallery.tableau.com/products/108)

Nandkishorbbd:
Hi,we are facing the limit of 10000 rows in kepler.gl extension for tableau.. Any idea how does it works for rows more than 10k?

nikhildha1:
Hi, I don't know if this thread is still active but I'm also facing the same issue as @Nandkishorbbd. Is there a plan to release a new version of the extension with a greater row limit?

heshan0131:
Unfortunately, the 10000 rows limit is set by tableau, I would wait for Tableau to relax on this limit

PR #465, opened 2019-04-02
[Enhancement] expand bottom widget to full length if in read only mode

heshan0131:
#339


mendezke:
is it not available as Tableau Extension?

PR #472, opened 2019-04-16
[Enhancement] Cleanup load map style tasks

heshan0131:
- Add `requestMapStyles` action
- Consolidate style loading using `getLoadMapStyleTasks` called by `requestMapStyles` and `receiveMapConfg`
- Add tests

### DIFF
--- a/src/reducers/map-style-updaters.js
+++ b/src/reducers/map-style-updaters.js
@@ -30,12 +30,15 @@ import {
   editTopMapStyle,
   editBottomMapStyle
 } from 'utils/map-style-utils/mapbox-gl-style-editor';
-import {DEFAULT_MAP_STYLES, DEFAULT_LAYER_GROUPS} from 'constants/default-settings';
+import {
+  DEFAULT_MAP_STYLES,
+  DEFAULT_LAYER_GROUPS
+} from 'constants/default-settings';
 import {generateHashId} from 'utils/utils';
 import {LOAD_MAP_STYLE_TASK} from 'tasks/tasks';
 import {loadMapStyles, loadMapStyleErr} from 'actions/map-style-actions';
 import {rgb} from 'd3-color';
-import { hexToRgb } from 'utils/color-utils';
+import {hexToRgb} from 'utils/color-utils';
 
 const DEFAULT_BLDG_COLOR = '#D1CEC7';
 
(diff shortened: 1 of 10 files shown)


src/reducers/map-style-updaters.js
```diff
@@ -171,17 +177,25 @@ function getMapStyles({
...
+  const buildingColor =
+    backgroundLayer &&
+    backgroundLayer.paint &&
+    backgroundLayer.paint['background-color']
+      ? backgroundLayer.paint['background-color']
+      : DEFAULT_BLDG_COLOR;
   // brighten or darken building based on style
-  const operation = style.id.match(/(?=(dark|night))/) ? 'brighter':  'darker';
+  const operation = style.id.match(/(?=(dark|night))/) ? 'brighter' : 'darker';
   const rgbObj = rgb(buildingColor)[operation]([0.2]);
```
macrigiuseppe: maybe it's better to create a constant for the 0.2 value

src/reducers/map-style-updaters.js
```diff
@@ -289,6 +314,11 @@ export const loadMapStylesUpdater = (state, action) => {
 // do nothing for now, if didn't load, skip it
 export const loadMapStyleErrUpdater = (state, action) => state;
```
macrigiuseppe: delete __action__ if you are not using it

src/reducers/map-style-updaters.js
```diff
@@ -368,19 +402,24 @@ export const resetMapConfigMapStyleUpdater = (state) => {
...
-      style: JSON.parse(JSON.stringify(style)),
-      label: style.name,
-      // gathering layer group info from style json
-      layerGroups: getLayerGroupsFromStyle(style)
-    } : {}),
+    ...(style
+      ? {
+          id: style.id || generateHashId(),
+          // make a copy of the style object
+          style: JSON.parse(JSON.stringify(style)),
```
macrigiuseppe: you are trying to clone an object here. Cloning directly a js object is more performant that using JSON.
https://jsperf.com/deep-copy-vs-json-stringify-json-parse/5

Lodash has several methods to support clone and deep clone

heshan0131: done

PR #479, opened 2019-04-17
[Enhancement] load svg icons from aws, add bundle analyzer, reduce bundle size -1mb

heshan0131:
- add script to analyze bundle size
- load svg icons from inside icon layer


heshan0131:
- svg icon layer currently loads icon geometry from a json file. webpack will inline it twice, adds 1mb to bundle size

Bundle analyze before:
Screen Shot 2019-04-16 at 6 18 18 PM

Bundle size before: 4.53mb

Screen Shot 2019-04-16 at 6 18 04 PM

Bundle analyze after
Screen Shot 2019-04-16 at 5 00 02 PM


Bundle size after: 3.56mb
Screen Shot 2019-04-16 at 6 11 40 PM

heshan0131:
https://github.com/uber/kepler.gl/issues/157

heshan0131:
@chrisirhc can't find your issue about icon file too larger. Maybe you can add it here

src/layers/icon-layer/icon-info-modal.js
```diff
@@ -117,31 +116,34 @@ const IconTable = styled.div`
...
+  const IconInfoModal = () => (
+    <div className="icon-info-modal">
+      <div className="icon-info-modal__description">
+        <span>
+          In your csv, create a column, put the name of the icon you want to
+          draw in it. You can leave the cell empty if you do not want the icon
+          to show for some points. When the column is named{' '}
+        </span>
+        <code>icon</code>
+        <span> Voyager will automatically create a icon layer for you.</span>
```
macrigiuseppe: Replace Voyager with Kepler

src/layers/icon-layer/icon-layer.js
```diff
@@ -20,27 +20,15 @@
...
-    [curr.id]: curr.mesh.cells.reduce((prev, cell) => {
-      cell.forEach(p => {
-        Array.prototype.push.apply(prev, curr.mesh.positions[p]);
-      });
-      return prev;
-    }, [])
-  }),
-  {}
-);
+export const SVG_ICON_URL = `${CLOUDFRONT}/icons/svg-icons.json`;
```
macrigiuseppe: are we sure we want to create this connection between static assets and cloudfront bucket?
This will create a direct dependency with the specific cloudfront url and our assets, if someone removes the icon file from cloudfront it may break kepler.gl instances.

What about using kepelr.gl github repo as an alternative solution?

src/layers/icon-layer/icon-layer.js
```diff
@@ -104,13 +96,40 @@ export default class IconLayer extends Layer {
...
       id: 'iconInfo',
-      template: IconInfoModal,
+      template: this._layerInfoModal,
       modalProps: {
         title: 'How to draw icons'
       }
     };
   }
 
+  async getSvgIcons(callback) {
```
macrigiuseppe: remove __callback__ param because it is not used.

Do you need this method to be part of this class? I think it could be an external method to fetch all data and use the callback to set the necessary property by providing a callback or the consumer can use await and retrieve the result.

if you initialize more icon layers you may perform the same network calls multiple times. If you externalize the method you may also think about storing the data into a variable /cache to speed subsequent calls

src/layers/icon-layer/icon-layer.js
heshan0131: We are already loading several assets, e.g. layer icons, logos, map styles from the same cloudfront, this doesn't add a connection. Just following the old practice

src/layers/icon-layer/icon-info-modal.js
heshan0131: done

src/layers/icon-layer/icon-layer.js
heshan0131: I have thought about it before, and decided to go this route. The svg icons are only used in the icon layer. So the loading part should be handled in the layer life cycle. This will make it easier later when we modulize kepler.gl,  packaging the layers of kepler.gl into its own package nad and have users to selectively import layers they needed. 

I agree if multiple icon layer were created, this call will be called multiple times. But this is not always a bad thing. Icon layer should be able to accept icon props and draw difference selection of svg icons. So you can image we will add a icons prop to this layer, and different layers might how a different collection of icons

Issue #480, opened 2019-04-18
Cannot add data to map by "browse your files"


zihaogt:
I could not upload data using the "browse your files" function. Nothing was loaded after selecting an .csv or .geojson file. It used to work a few weeks earlier. The drag and drop function still works for the same file.

 - OS: Windows 10
 - Browser: Chrome
 - Version: 73.0.3683.103

heshan0131:
Are you using kepler.gl/#/demo website or running kepler.gl locally? Can you open your browser console and send us any error you can see?

zihaogt:
I am using kepler.gl/#/demo website. I only see two warnings in the Chrome browser console:

bundle.js:131 Global fetch is deprecated and will be unsupported in a future version. Please pass fetch function as option when instantiating dropbox instance: new Dropbox({fetch})

bundle.js:51 This page appears to be missing CSS declarations for Mapbox GL JS, which may cause the map to display incorrectly. Please ensure your page includes mapbox-gl.css, as described in https://www.mapbox.com/mapbox-gl-js/api/.

heshan0131:
Thank you. I was able to reproduce the bug. Fix will come out soon.

t496971418:
I have the same question.🧐

t496971418:
![WX20190425-122710](https://user-images.githubusercontent.com/25219343/56709801-cc5c6880-6755-11e9-8172-71b1a5ac91c6.png)

macrigiuseppe:
This was fixed and release in 1.0.0-1 version

PR #486, opened 2019-04-22
Removed conditions for file upload handler

macrigiuseppe:
Fixes issue #480


macrigiuseppe:
> Might consider change the name of this function to `handleFileInput` to signal that this is not only for file drop

Done

Issue #485, opened 2019-04-19
[Bug] Brush is brokem


heshan0131:
Brush is broken with deck.gl 6.4

Issue #484, opened 2019-04-19
[Bug] Highlight broken in point layer


heshan0131:
**Describe the bug**
When you click on a point, it should stay yellow

chrisirhc:
Was able to reproduce this today on kepler.gl. 
![highlight](https://user-images.githubusercontent.com/132584/58727517-72287600-8399-11e9-9e3d-b5ba8b45a674.gif)

Moving this back to "Bug".

PR #487, opened 2019-04-22
[Bug] Fix point layer brushing

heshan0131:
kepler.gl manually edits deck.gl default layer shaders to enable brushing. This is sketchy practice overall since whenever deck.gl layer shader changes, brushing will break. 

After fixing the shader edit util for `scatterplot-brushing-layer`. I added test util to test render deck.gl layer using `@deck.gl/test-utils`. This way at least we fail the test if manually editing deck.gl shader errors out and we will catch it in the test.


heshan0131:
#485

.travis.yml
```diff
@@ -1,10 +1,23 @@
 language: node_js
+addons:
```
macrigiuseppe: are these used to test webgl in a server env?

heshan0131: Yes, this is added because we have`gl` in devDependencies. And we run headless-gl test in travis

heshan0131:
#484

Issue #346, opened 2019-01-28
Kepler.gl should allow developers to pass a StyledComponents theme to override the original style


macrigiuseppe:
**Describe the bug**
Currently Kepler.gl implements styling using styled-components and inject its own theme. Unfortunately there is no way for the parent app to inject a different styled-components theme to override/extend the existing one.

IgorWilbert:
Hey @macrigiuseppe ! Are you still working on this? Thanks!

macrigiuseppe:
> Hey @macrigiuseppe ! Are you still working on this? Thanks!

I haven't started yet, but i will get on this soon.

macrigiuseppe:
My solution:
kepler.gl should access a new prop: them which will allow developers to pass a custom theme and in keplergl.js

we perform the following changes:
```javascript
import {theme as basicTheme} from 'styles/base';

render() {
  return (
    <ThemeProvider theme={{...this.props.them, ...theme}}> 
      ...
    </ThemeProvider>
  );
}
```

macrigiuseppe:
@IgorWilbert created a new pull request: https://github.com/uber/kepler.gl/pull/424

PR #285, opened 2018-11-17
[Feature] Jupyter notebook integration

heshan0131:
Add jupyter notebook app in examples/jupyter-notebook

PR #489, opened 2019-04-23
[Feat] Add a light theme to KeplerGl Prop

heshan0131:
- Allow passing `theme: 'light'` to `KeplerGl` component

Screen Shot 2019-04-22 at 6 21 52 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,11 +214,11 @@ By Parsing `mint: false` kepler.gl will keep the component state in the store ev
 
 Read more about [Components][components].
 
-##### `theme` (Object, optional)
+##### `theme` (Object | String, optional)
 
 - default: `null`
 
-Object used to customize Kepler.gl style. Kepler.gl will use the value passed as input to override values from [theme](https://github.com/uber/kepler.gl/blob/master/src/styles/base.js).
+You can pass theme name or object used to customize Kepler.gl style. Kepler.gl provide an `'light'` theme besides the default 'dark' theme. When pass in a theme object Kepler.gl will use the value passed as input to override values from [theme](https://github.com/uber/kepler.gl/blob/master/src/styles/base.js).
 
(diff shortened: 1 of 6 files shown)


heshan0131:
#339 #285 #346

README.md
```diff
@@ -214,11 +214,11 @@ By Parsing `mint: false` kepler.gl will keep the component state in the store ev
 
 Read more about [Components][components].
 
-##### `theme` (Object, optional)
+##### `theme` (Object | String, optional)
 
 - default: `null`
 
-Object used to customize Kepler.gl style. Kepler.gl will use the value passed as input to override values from [theme](https://github.com/uber/kepler.gl/blob/master/src/styles/base.js).
+You can pass theme name or object used to customize Kepler.gl style. Kepler.gl provide an `'light'` theme besides the default 'dark' theme. When pass in a theme object Kepler.gl will use the value passed as input to override values from [theme](https://github.com/uber/kepler.gl/blob/master/src/styles/base.js).
```
macrigiuseppe: do you think we should provide a list of possible values? or a location in the code/docs where they can find values

src/constants/default-settings.js
```diff
@@ -105,6 +105,11 @@ export const DIMENSIONS = {
   }
 };
 
+export const THEME = keyMirror({
```
macrigiuseppe: a think of description here will help people to understand they can use these values as kepler prop

macrigiuseppe:
Nice one:
Screen Shot 2019-04-23 at 10 49 48 AM

README.md
heshan0131: Hmm yea, we should definitely better document this. I am hoping the variable name itself should be self-explanatory.

src/constants/default-settings.js
heshan0131: done

PR #506, opened 2019-05-08
[Feat] add onViewStateChange callback to KeplerGl

heshan0131:
Matching deck.gl's [`onViewStateChange`](https://deck.gl/#/documentation/deckgl-api-reference/deck?section=onviewstatechange-function-) Function

Screen Shot 2019-05-08 at 2 46 51 PM

### DIFF
--- a/src/components/map-container.js
+++ b/src/components/map-container.js
@@ -389,7 +389,7 @@ export default function MapContainerFactory(MapPopover, MapControl) {
     }
 
     _onViewportChange = (viewState) => {
-      if (typeof this.props.onViewStateChange === 'function' && this.props.index === 0) {
+      if (typeof this.props.onViewStateChange === 'function') {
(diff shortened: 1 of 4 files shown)


src/components/map-container.js
```diff
@@ -388,12 +388,19 @@ export default function MapContainerFactory(MapPopover, MapControl) {
       }
     }
 
+    _onViewportChange = (viewState) => {
+      if (typeof this.props.onViewStateChange === 'function' && this.props.index === 0) {
```
macrigiuseppe: When in split mode:
Kepler performs the callback only when moving the left map but not the right one.
Should kepler send updates despite of which map the user is trying to move?

even in split mode this method is called only on the container users are trying to move so we shouldn't get any duplicated event

heshan0131: that's true, updated

PR #509, opened 2019-05-14
[Bug] Fixed layer list sorting dnd effect

macrigiuseppe:
This diff will fix the drag and drop issue when sorting current layers.


src/components/side-panel/layer-manager.js
```diff
@@ -205,7 +205,7 @@ function LayerManagerFactory(AddDataButton, LayerPanel, SourceDataCatalog) {
               lockAxis="y"
               useDragHandle={true}
             >
-              {layerOrder.map(idx => {
+              {layerOrder.map((idx, index) => {
```
heshan0131: probably name it to `layerIdx` for clarity

PR #513, opened 2019-05-16
[Enhancement] Detecting mapbox token validity

macrigiuseppe:
## New approach
Kepler.gl will perform validation and show a warning in the browser console if the mapbox token passed as react prop is not valid.

Screenshot
Screen Shot 2019-05-20 at 1 42 27 PM

The message will also contain a link where users can retrieve info to create a new mapbox token.

==========
### Old Approach
Kepler will perform a validity check on the mapbox token passed to the instance. If the token is not valid kepler will show a notification error.
Screen Shot 2019-05-16 at 2 42 35 PM

missing-token

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,6 +172,8 @@
     "gl": "^4.2.2",
     "jsdom": "^11.1.0",
     "json-loader": "^0.5.4",
+    "lodash.get": "^4.4.2",
+    "lodash.set": "^4.3.2",
     "nyc": "^11.4.1",
     "prettier": "^1.7.0",
     "progress-bar-webpack-plugin": "^1.9.3",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "docs": "babel-node ./scripts/documentation.js",
     "lint": "eslint src webpack/*.js examples/*.js examples/**/src examples/**/*.js website/*.js website/src --fix",
     "web": "(yarn && yarn install:web && yarn start:web)",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -174,6 +174,8 @@
     "immutable": "^4.0.0-rc.12",
     "jsdom": "^11.1.0",
     "json-loader": "^0.5.4",
+    "lodash.get": "^4.4.2",
```
macrigiuseppe: used for testing only

src/components/notification-panel/notification-item.js
```diff
@@ -103,7 +103,7 @@ export default function NotificationItemFactory()
             className="notification-item--message"
             expanded={this.state.isExpanded}
             theme={this.props.theme}>
-            <ReactMarkdown source={notification.message} />
+            <ReactMarkdown linkTarget="_blank" source={notification.message} />
```
macrigiuseppe: open a notification link in a new tab to avoid losing current work

src/reducers/ui-state-updaters.js
```diff
@@ -201,6 +206,20 @@ export const INITIAL_UI_STATE = {
 };
 
 /* Updaters */
+
+export const initUIStateUpdater = (state, action) => {
+  const isTokenValid = validateToken((action.payload || {}).mapboxApiAccessToken);
```
macrigiuseppe: do a sync validation of the token and update notification and modal.
If the token is not valid we also hide the data molda

Issue #498, opened 2019-05-03
Provide UI feedback for operations


macrigiuseppe:
**Is your feature request related to a problem? Please describe.**
Currently Kepler.gl doesn't provide a UI feedback when operations are successful or failed.

**Describe the solution you'd like**
When operations (list below) are performed, Kepler.gl should display a notification message providing information about the outcome of the operation along with details if necessary.

Operations:
- Upload file
- Upload url
- Share
- Export

Notifications should be provided through the notifications system we have in place.

PR #514, opened 2019-05-17
[Enhancement] Upload file feedback

macrigiuseppe:
## Features
Provide UI feedback when user uploads files:
- success: success notification with name of the file
- failure: show error description reason

Moved file related action from vis-state-updaters to combined reducers to handle multiple updates.
Each single upload action will now contain either data or error along with file info object.

#### Empty file:
empty-file

#### Valid file:
valid-file

I can see the option where we can provide more info with a successful notifications, e.g. # of rows.
Not showing the successful notification could also be another options.

### DIFF
--- a/examples/custom-reducer/package.json
+++ b/examples/custom-reducer/package.json
@@ -8,7 +8,6 @@
     "kepler.gl": "^1.0.0-2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-palm": "^3.1.2",
     "react-redux": "^5.1.1",
     "react-virtualized": "^9.21.0",
     "redux-actions": "^2.2.1",
--- a/examples/custom-reducer/package.json
+++ b/examples/custom-reducer/package.json
@@ -8,8 +8,8 @@
     "kepler.gl": "^1.0.0-2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-palm": "^1.1.2",
-    "react-redux": "^4.0.0",
+    "react-palm": "^3.1.2",
+    "react-redux": "^5.1.1",
     "react-virtualized": "^9.21.0",
     "redux-actions": "^2.2.1",
     "styled-components": "^4.1.3"
@@ -45,5 +45,9 @@
     "webpack-dev-server": "^3.1.14",
     "webpack-hot-middleware": "^2.24.3",
     "webpack-stats-plugin": "^0.2.1"
+  },
+  "_commnet": "this is temporary until we release the next version with react-palm-3.1.2",
+  "resolutions": {
+    "react-palm": "3.1.2"
   }
 }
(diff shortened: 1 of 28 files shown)


src/components/kepler-gl.js
```diff
@@ -87,7 +88,7 @@ KeplerGlFactory.deps = [
 function KeplerGlFactory(
   BottomWidget,
   MapContainer,
-  ModalWrapper,
+  ModalContainer,
```
macrigiuseppe: to be consistent with the other properties

examples/custom-theme/package.json
```diff
@@ -8,8 +8,7 @@
     "kepler.gl": "^1.0.0-2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-palm": "^1.1.2",
-    "react-redux": "^4.0.0",
```
macrigiuseppe: we don't need this dependency since we are using __enhanceReduxMiddleware__ which will inject react-palm into application consumer

examples/custom-reducer/package.json
```diff
@@ -8,8 +8,7 @@
     "kepler.gl": "^1.0.0-2",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
-    "react-palm": "^1.1.2",
```
macrigiuseppe: we don't need this dependency since we are using enhanceReduxMiddleware which will inject react-palm into application consumer

src/components/notification-panel/notification-item.js
```diff
@@ -54,6 +54,11 @@ const NotificationMessage = styled.div`
   p {
     margin-top: 0;
   }
+  a {
```
heshan0131: prefer not to use `!Important` it will be hard to override. you can add a class before a
```
styled.div`
  .notification-link {
      a {
          color: white;
      }
  }
`
```

src/reducers/combined-updaters.js
```diff
@@ -182,3 +189,132 @@ export const addDataToMapUpdater = (state, {payload}) => {
 };
 
 export const addDataToMapComposed = addDataToMapUpdater;
+
+/**
+ * Trigger file loading dispatch `addDataToMap` if succeed, or `loadFilesErr` if failed
```
heshan0131: Should mention the result here is `withTasks(nextState, loadFileTasks)`

src/utils/file-utils.js
```diff
@@ -32,3 +32,26 @@ export const processFileToLoad = (file) => {
...
   }
 };
+
+/**
+ * Print human readable file size
+ * @param bytes number of bytes
+ * @return {string} human readable version of file size
+ */
+export const sizeOf = bytes => {
+  if (bytes === 0) {
```
heshan0131: what if bytes is `undefined`?

heshan0131:
Please run `npm run docs` since some actions has been moved

heshan0131:
#498

heshan0131:
Note that this PR doesn't capture file processing error, invalid geojson or csv format etc. Which is usually the reason why the spinner keeps spinning. Future PR should address it

PR #540, opened 2019-06-03
Add ScenegraphLayer

georgios-uber:


src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+      size: {
+        ...super.visualChannels.size,
+        range: 'radiusRange',
+        property: 'radius',
+        channelScaleType: 'radius'
+      }
+    };
+  }
+
+  get layerInfoModal() {
```
heshan0131: If you don't plan to add any modal to explain how to draw a screen graph layer, you don't need this. But I think  we can create a simple modal that explains the supported modal format and put a screenshot there :)

This is what icon layer has:
Screen Shot 2019-06-03 at 1 54 19 PM

Screen Shot 2019-06-03 at 1 54 27 PM

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+    return {
+      id: 'iconInfo',
+      template: this._layerInfoModal,
+      modalProps: {
+        title: 'How to draw icons'
+      }
+    };
+  }
+
+  static findDefaultLayerProps({fieldPairs, fields}) {
```
heshan0131: This function will get called when dataset is added to kepler,  it automatically create layers based on certain column name conversion. If you don't plan to automatically create a 3d modal layer, you don't need this static method

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+        accu.push({
+          index,
+          data: allData[index]
+        });
+
+        return accu;
+      }, []);
+    }
+
+    const getRadius = rScale
```
heshan0131: `getRadius` is not a valid prop for `ScreenGraphLayer`, you don't need to calculate it here and pass to the render function

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+      ? d => this.getEncodedChannelValue(rScale, d.data, sizeField)
+      : 1;
+
+    const getColor = cScale
```
heshan0131: Looks like you are using a constant `[255, 255, 255, 255]` for `getColor` in `renderLayer` function below, if color is always a constant, you don't need this accessor either

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+      },
+      isVisible: true
+    }));
+
+    return props;
+  }
+
+  // TODO: fix complexity
+  /* eslint-disable complexity */
+  formatLayerData(_, allData, filteredIndex, oldLayerData, opt = {}) {
```
heshan0131: `formatLayerData` only got called when filters changed, or color, radius, size by value is changed. its output should be a new data array or attributed accessors. If `sizeScale `,`angleX`, `angleY`, `angleZ` is directly passed to Deckgl layer, you don't have to calculate them here. Just directly pass them in the `renderLayer`function below

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,312 @@
...
+    const getColor = cScale
+      ? d => this.getEncodedChannelValue(cScale, d.data, colorField)
+      : color;
+
+    return {
+      data,
+      getPosition,
+      getColor,
+      getRadius,
+      sizeScale: sizeScale3D,
```
heshan0131: basically, you only need to return `data` and `getPosition` here. `sizeScale` and `getOrientation` can be assigned in the `renderLayer` function below

src/layers/scenegraph-layer/scenegraph-layer.js
georgios-uber: Screen Shot 2019-06-07 at 3 47 11 PM

src/layers/scenegraph-layer/scenegraph-layer.js
georgios-uber: Removing for now

src/layers/scenegraph-layer/scenegraph-layer.js
georgios-uber: I want to have a color picker like the other layers.

georgios-uber:
Will rebase on master now...

src/layers/scenegraph-layer/scenegraph-layer.js
```diff
@@ -0,0 +1,239 @@
...
+
+        scenegraph:
+          this.config.visConfig.scenegraph ||
+          'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/Duck/glTF-Binary/Duck.glb',
+
+        sizeScale,
+        getTranslation: [0, 0, 0],
+        getScale: [0.1, 0.1, 0.1],
+        getOrientation: [angleX, angleY, angleZ],
+        getColor: [255, 255, 255, 255],
```
heshan0131: Make all the default constans
```js
DEFAULT_MODEL = 'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/Duck/glTF-Binary/Duck.glb';
DEFAULT_TRANSITION = [0, 0, 0]
DEFAULT_SCALE = [0.1, 0.1, 0.1],
DEFAULT_COLOR = [255, 255, 255, 255];
```

georgios-uber: done

deeayeen:
Adding gltf file returns an error on kepler.gl/demo:

"GET blob:https://kepler.gl/scene.bin net::ERR_FILE_NOT_FOUND"

Steps to reproduce: Add new layer, selected 3d, choose gltf file, upload, recieve error in console.

PR #544, opened 2019-06-07
Upgrade to  deck.gl 7 and luma.gl 7

1chandu:
- Upgrade API usage.
- Remove custom `h3-hexagon-cell` layer and use deck.gl `H3HexagonLayer`.
- Customize `ColumnLayer` to support per instance `coverage`.


examples/webpack.config.local.js
```diff
@@ -50,7 +55,34 @@ function makeLocalDevConfig(EXAMPLE_DIR = LIB_DIR) {
         // Imports kepler.gl library from the src directory in this repo
         'kepler.gl': SRC_DIR,
         react: resolve(EXAMPLE_DIR, './node_modules/react'),
-        'styled-components': resolve(EXAMPLE_DIR, './node_modules/styled-components')
+        'styled-components': resolve(
```
georgios-uber: @heshan0131 I think we should use luma/deck/loaders from a master checkout instead of using node modules.
This way we can develop in luma/deck/loaders and have live reload.

heshan0131: sure, that requires luma/deck/loaders be cloned to a local folder. You can change `NODE_MODULES_DIR` to a custom folder that contains all 3 modules manually. Most of the time we just want to make sure the version of luma/deck/loaders leasted in package.json should work properly

Issue #531, opened 2019-05-29
Layer Legend By Color does not show the different categories anymore


alex-arp:
**Describe the bug**
When coloring a geo-object according to a certain variable, kepler.gl used to show a legend of the color scheme corresponding to categories/buckets of values from the variable.

This is not the case anymore.

The issue arised when using .geojson data and in Google Chrome browser.

**To Reproduce**
1. Go to StrokeColor
2. Click on StrokColor Based On
3. Select variable in dataset
4. Click on legend button on top right corner of the map
5. See error

**Screenshots**

Dropping some data:
![Screenshot 2019-05-29 at 14 21 19](https://user-images.githubusercontent.com/43449134/58560982-184f7100-821e-11e9-920f-c59922bd33e0.png)

Colouring by a variable:
![Screenshot 2019-05-29 at 14 21 49](https://user-images.githubusercontent.com/43449134/58561018-2604f680-821e-11e9-92e7-282d3d8d58b4.png)

What user used to see in the legend:
![legend copy](https://user-images.githubusercontent.com/43449134/58561041-3026f500-821e-11e9-8244-7ac103e4be4a.png)

**Desktop**
 - OSX
 - Chrome browser
 - Version 74.0.3729.169 (Official Build) (64-bit)

Thanks in advance for your help!

heshan0131:
Verified bug. Fix will come soon

heshan0131:
This should be fixed, please verify and reopen if there is any thing else

<img width="808" alt="Screen Shot 2019-06-27 at 8 11 06 AM" src="https://user-images.githubusercontent.com/3605556/60277842-335fe000-98b3-11e9-890d-092b7cfa1c29.png">

PR #546, opened 2019-06-07
[Bug] Display color legend for stroke color scale

heshan0131:
- Handle cases when layer has more than 1 color channel
Screen Shot 2019-06-07 at 1 57 54 PM

#531

### DIFF
--- a/src/components/map/map-legend.js
+++ b/src/components/map/map-legend.js
@@ -26,7 +26,7 @@ import ColorLegend from 'components/common/color-legend';
 import {DIMENSIONS, CHANNEL_SCALES} from 'constants/default-settings';
 import {camelToTitle} from 'utils/utils';
 
-const StyledMapControlLegend = styled.div`
+export const StyledMapControlLegend = styled.div`
   padding: 10px 0 10px ${props => props.theme.mapControl.padding}px;
   font-size: 11px;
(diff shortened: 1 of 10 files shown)


src/layers/base-layer.js
```diff
@@ -310,6 +310,7 @@ export default class Layer {
    */
   getVisualChannelDescription(key) {
     // e.g. label: Color, measure: Vehicle Type
+    console.log(key)
```
macrigiuseppe: remove comment

src/components/map/map-legend.js
```diff
@@ -84,42 +84,76 @@ const propTypes = {
   layers: PropTypes.arrayOf(PropTypes.object)
 };
 
-const SingleColorLegend = ({layer, width}) => (
+const SingleColorLegend = ({width, color}) => (
   <ColorLegend
     scaleType="ordinal"
     displayLabel={false}
     domain={['']}
```
macrigiuseppe: I read about React performance improvements here: https://blog.opendigerati.com/taking-control-of-react-performance-4f21a15a19b5#483c

You can define a constant outside SingleColorLegend to avoid re-rendering

```
const emptyDomain = ['']

and use it as domain in ColorLegend for the domain property
```

Issue #517, opened 2019-05-20
Visualisation of points erroneous if radius_range_min is set to 0.


sidPhoenix17:
**Describe the bug**
If I select the visualisation as "Point" and fix the radius range from a radius_min to radius_max basis a specific column, then if radius_min = 0 , the points with lowest score end up having extremely large radius. But if I set radius_min=0.1, then the points with lowest score are sized relatively.

**To Reproduce**
To reproduce this behaviour, I have created a map with my own data.
1. Go to https://kepler.gl/demo/map?mapUrl=https://dl.dropboxusercontent.com/s/zhqhd6ro2w1000v/keplergl_iqibqu9.json
2. This map has location data in two cities - Mumbai & Delhi.
a. For Mumbai's layer, I have set radius_min = 0.1 - zoom in and you will be able to see a clear radius variation depending on the value of "d_score" - column name.
b.  For Delhi's layer, I have set radius_min = 0 - zoom in and you will be able to see all blocks with the value of d_score = 0 (lowest in this case) end up having an arbitrarily fixed radius. 

**Expected behavior**
I'd expect that if I set the radius_min = 0, then the point should be infinitesimally small and not hinder in the visualisation.

**Desktop (please complete the following information):**

OS: macOS 10.14.4
Browser: Chrome 73.0.3683.103

Please let me know if I have an incorrect understanding of the feature.

heshan0131:
This should be fixed, I see these 2 maps now. Looks correct to me.

Closing this issue, please reopen if there is anything else

<img width="727" alt="Screen Shot 2019-06-27 at 8 23 30 AM" src="https://user-images.githubusercontent.com/3605556/60278760-de24ce00-98b4-11e9-87f5-0abbd5d51af0.png">
<img width="624" alt="Screen Shot 2019-06-27 at 8 23 18 AM" src="https://user-images.githubusercontent.com/3605556/60278761-de24ce00-98b4-11e9-80cb-d52f94b25b72.png">

PR #551, opened 2019-06-11
[Bug] Fix radius rendering when value = 0

heshan0131:
- Pass default null value to `getRadius`
- Use `notNullorUndefined` to check whether value is `falsy`

Screen Shot 2019-06-10 at 4 45 37 PM
[__Before__]

Screen Shot 2019-06-10 at 4 56 44 PM
[__After__]


heshan0131:
#517

PR #552, opened 2019-06-12
[Docs] update user documentation with newer layers and features

yanhann10:
- Add hex bin, heat map, h3 layer to documentation

- Change geojson layer to polygon layer

### DIFF
--- a/docs/user-guides/c-types-of-layers/e-polygon.md
+++ b/docs/user-guides/c-types-of-layers/e-polygon.md
@@ -1,0 +1,20 @@
+# Polygon
+
+Polygon layers can display either paths or polygons in 2D or 3D with geoJSON inputs.
+
+![GeoJSON layer](https://d1a3f4spazzrp4.cloudfront.net/kepler.gl/documentation/image20.png "GeoJSON layer")
+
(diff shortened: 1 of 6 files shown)


docs/user-guides/c-types-of-layers/e-polygon.md
```diff
@@ -0,0 +1,20 @@
+# Polygon
+
+Polygon layers can display either paths or polygons in 2D or 3D with geoJSON inputs.
```
heshan0131: Polygon layer can display [GeoJSON](https://tools.ietf.org/html/rfc7946)
 features. GeoJSON supports the following geometry types: `Point`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString`, `MultiPolygon`.

kepler.gl supports all the GeoJSON types above. You can pass in either a single [`Feature`][features] or a [`FeatureCollection`][feature_collection]

[feature_collection]: https://tools.ietf.org/html/rfc7946#section-3.3
[features]: https://tools.ietf.org/html/rfc7946#section-3.2

yanhann10: Updated

Issue #553, opened 2019-06-12
Replace react-data-grid with react-window


macrigiuseppe:
As part of our package size investigation, we are going to replace react-data-grid with react-window to reduce bundle size and improve performance

More info: #499

PR #562, opened 2019-06-17
[Enhancement] Replace react-data-grid with react window

macrigiuseppe:
This pull request replaces react-data-grid with an implementation of react-window

Screen Shot 2019-06-17 at 10 24 52 AM

react-window

Solves #553


test/browser/components/data-table-model-test.js
```diff
@@ -59,60 +60,9 @@ test('Components -> DataTableModal.render', t => {
   );
 
   t.equal(wrapper.find(DataTableModal).length, 1, 'should render DataTableModal data');
-  t.equal(wrapper.find(ReactDataGrid).length, 1, 'should render ReactDataGrid data');
+  t.equal(wrapper.find(Autosizer).length, 1, 'should render Autosizer');
   t.equal(wrapper.find(DatasetTabs).length, 1, 'should render DatasetTabs');
   t.equal(wrapper.find(DatasetModalTab).length, 1, 'should render 1 dataset-modal-catalog');
-  t.equal(wrapper.find('.react-grid-Row').length, 10, 'should render 10 rows');
```
macrigiuseppe: React-Window doesn't provide subnested elements when running tests

examples/custom-theme/src/app.js
```diff
@@ -20,8 +20,8 @@
 
 import React, {Component} from 'react';
 import styled from 'styled-components';
-import window from 'global/window';
 import {connect} from 'react-redux';
+import AutoSizer from 'react-virtualized-auto-sizer';
```
heshan0131: Not I see `AutoSizer` being use everywhere, It feels  like kepler.gl is married to it and can't leave without `AutoSizer`. Our example should be a simple one that requires a minimum amount of dependencies. Something like `AutoSizer` is good to have in `demo-app`, but probably not a must in all other examples.

I would rather remove it and use good old school ` window.addEventListener`

test/browser/components/data-table-model-test.js
heshan0131: what do you mean by that?

If we only stop at `Autosizer` then there is no point running this test. The test here is to make sure different types of data render correctly. Please put this test back. The mis display of boolean value would have been caught by this test

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
...
-    // TODO: this should be all data
-    const rows = activeDataset.data;
-    const columns = activeDataset.fields
-      .map((field, i) => ({
-        ...field,
-        key: i,
-        headerRenderer: <FieldHeader {...field} />,
-        resizable: true,
-        formatter:
-          field.type === ALL_FIELD_TYPES.boolean ? BooleanFormatter : undefined
```
heshan0131: The `formatter` here is to make sure Boolean value renders correctly, without a specific formatter, boolean value can't be displayed

[on master]
Screen Shot 2019-06-21 at 12 26 32 AM

[in this pr]
Screen Shot 2019-06-21 at 12 25 24 AM

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
...
+  activeDatasetSelector = createSelector(
+    this.datasetsSelector,
+    this.dataIdSelector,
+    (datasets, dataId) => datasets[dataId]
+  );
+  columnsSelector = createSelector(
+    this.activeDatasetSelector,
+    dataset => dataset.fields.filter(({name}) => name !== '_geojson')
+  );
+  rowsSelector = createSelector(
```
heshan0131: save here, you don't need to create selector. We are not doing any operation on `dataset.data`, this selector simply returns `dataset.data`

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
...
+  headerGrid = React.createRef();
+
+  datasetsSelector = props => props.datasets;
+  dataIdSelector = props => props.dataId;
+  activeDatasetSelector = createSelector(
+    this.datasetsSelector,
+    this.dataIdSelector,
+    (datasets, dataId) => datasets[dataId]
+  );
+  columnsSelector = createSelector(
```
heshan0131: You don't need to call `createSelector` on `activeDatasetSelector` it is not derived data, it simply returns `datasets[id]`  This can just be
```js
activeDatasetSelector = props => props.datasets[props.dataId],
columnsSelector = createSelector(
    this.activeDatasetSelector,
    dataset => dataset.fields.filter(({name}) => name !== '_geojson')
  );
```

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
...
-              headerRowHeight={72}
-              columns={columns}
-              minColumnWidth={172}
-              minWidth={this.props.width}
-              minHeight={this.props.height - 65}
-              rowGetter={i => rows[i]}
-              rowHeight={48}
-              rowsCount={rows.length}
-            />
+        <Autosizer>
```
heshan0131: is `Autosizer` really necessary here?

With Autosizer, you still see white space on the right. The table should expand to entire with of the modal

Screen Shot 2019-06-21 at 12 48 53 AM

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
...
-  }
-
-  .react-grid-Grid {
-    border: 0;
-  }
-
-  .react-grid-Cell {
-    border-right: 0;
-    border-bottom: ${props => props.theme.panelBorderLT};
-    padding-left: 16px;
```
heshan0131: Left most of cell and header cell should have equal padding. Now it's too close to the edge
Screen Shot 2019-06-21 at 12 41 27 AM

src/components/modals/data-table-modal.js
```diff
@@ -30,183 +32,179 @@ import {Clock} from 'components/common/icons/index';
 // Breakpoints
 import {media} from 'styles/media-breakpoints';
 
-const ReactDataGrid = window.navigator ? require('react-data-grid/dist/react-data-grid.min') : null;
-
-let shouldPreventScrollBack = false;
-
-if (window.navigator && window.navigator.userAgent) {
```
heshan0131: This is to prevent scrolling left and right in the data table causes browser goes back to previous page. It would be extremely frustrating for someone to accidentally go back to the previous page and lose all his/her kepler.gl map. Please add this check back on `onMouseWheel`

ezgif com-optimize (14)

test/browser/components/data-table-model-test.js
macrigiuseppe: It looks like when running the tests, rows are not rendered because react-window requires the viewport of the browser to determine which rows are to be visualized

Issue #547, opened 2019-06-07
[Bug] Geojson failure when property contains JS object


heshan0131:
**Describe the bug**

When are geojson contains js object in the property, Kepler failed to render it in tooltip and data table

**To Reproduce**
1. download geojson at https://transit.land/api/v1/operators/o-9q8y-sfmta.geojson?per_page=false
2. upload to kepler
3. fail to render immediately

**Expected behavior**
Parse js objects as a string

matheusmonte:
@heshan0131  Can I work to solve this?

heshan0131:
YES! my thinking is when read in geojson from files, stringify properties that contain object. You can dig into [`geojson-processor`](https://github.com/keplergl/kepler.gl/blob/master/src/processors/data-processor.js#L414-L454)

matheusmonte:
Sure, I will work on this. Maybe on Saturday I send a PR solving this. Thanks

matheusmonte:
Hi @heshan0131 , I solve this. You're right, the problem occurs in geojson-processor , so I add a check type in properties and bug crash is avoid, but I'm not sure if this result is correct, you can see the render result in this image.

![Screen Shot 2019-06-18 at 13 59 18](https://user-images.githubusercontent.com/5808206/59708709-53b4de00-91d3-11e9-9dc7-ecba4673673c.png)

matheusmonte:
The code is in PR #564

heshan0131:
This is fixed, please reopen if this is anything else needed
Special thanks to @matheusmonte

PR #564, opened 2019-06-18
[Bug]Solve issue #547 avoid crash application

matheusmonte:
Add type check in geojson-processor to avoid crash when geojson contains an object in any properties. 

Signed-off-by: Matheus Monte <clark.monte1@gmail.com>

### DIFF
--- a/src/processors/data-processor.js
+++ b/src/processors/data-processor.js
@@ -422,11 +422,14 @@ export function processGeojson(rawData) {
   // getting all feature fields
   const allData = normalizedGeojson.features.reduce((accu, f, i) => {
     if (f.geometry) {
-      Object.keys(f.properties).forEach(key => {
-        if(typeof f.properties[key] === "object"){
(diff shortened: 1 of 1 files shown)


src/processors/data-processor.js
```diff
@@ -422,6 +422,11 @@ export function processGeojson(rawData) {
   // getting all feature fields
   const allData = normalizedGeojson.features.reduce((accu, f, i) => {
     if (f.geometry) {
+      Object.keys(f.properties).forEach(key => {
```
heshan0131: This will error if `properties` doesn't exist or not an object. I would do a `typeof f.properties === 'object' && f.properties !== null) check

matheusmonte: @heshan0131 Thanks for your review, I add this validation and update this PR

matheusmonte:
Hi @heshan0131 need something more in this PR? I don't get why netlify is failing , there's some problem in this code? 

Thanks for you review

PR #567, opened 2019-06-20
[Bug] add @deck.gl/experimental-layers back

yanhann10:
After pulling from latest master, rm node_modules and re run yarn. This error occurred:

Screen Shot 2019-06-20 at 1 01 38 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "umd"
   ],
   "dependencies": {
+    "@deck.gl/experimental-layers": "^6.4.10",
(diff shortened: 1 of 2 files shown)


1chandu:
`TileLayer` should be imported from `@deck.gl/geo-layers`, which it seems thats whats we are doin here : https://github.com/keplergl/kepler.gl/blob/e4fe29e94df3481c6bd35d80470f108a21e001f6/src/deckgl-layers/3d-building-layer/3d-building-layer.js#L22

Where are we using `experimental-layers`?

PR #569, opened 2019-06-22
Bump supercluster version

### LINKED ISSUE #524: Cluster layer goes blank
**Describe the bug**
As said the `Cluster Layer` goes blank when zoom out at small zoom level, the official demo can reproduce. 

Change the `earthquakes` demo point layer to cluster
image

 Then zoom out.
image

akre54:
Fixes #524 

As a bonus we're now on the same version as mapbox-gl!

### DIFF
--- a/src/deckgl-layers/layer-utils/cluster-utils.js
+++ b/src/deckgl-layers/layer-utils/cluster-utils.js
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import supercluster from 'supercluster';
+import Supercluster from 'supercluster';
 import memoize from 'lodash.memoize';
 
 export function getGeoJSON(data, getPosition) {
@@ -41,11 +41,9 @@ export function getGeoJSON(data, getPosition) {
 const clusterResolver = ({clusterRadius}) => `${clusterRadius}`;
 
 const getClusterer = memoize(({clusterRadius, geoJSON}) => {
-  return supercluster({
+  return new Supercluster({
     maxZoom: 20,
     radius: clusterRadius,
-    initial: () => ({points: []}),
-    map: props => props.data,
     reduce: (accumulated, props) => {
       if (props.points) {
         // avoid using spread to prevent max call stack exceeded error
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "redux": "^3.0.4",
     "redux-actions": "^2.2.1",
     "reselect": "^3.0.1",
-    "supercluster": "^3.0.0",
+    "supercluster": "^6.0.1",
     "type-analyzer": "0.2.1",
     "uber-licence": "^3.1.1",
     "viewport-mercator-project": "^6.0.0",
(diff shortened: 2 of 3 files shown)


src/deckgl-layers/layer-utils/cluster-utils.js
```diff
@@ -41,11 +41,9 @@ export function getGeoJSON(data, getPosition) {
 const clusterResolver = ({clusterRadius}) => `${clusterRadius}`;
 
 const getClusterer = memoize(({clusterRadius, geoJSON}) => {
-  return supercluster({
+  return new Supercluster({
     maxZoom: 20,
     radius: clusterRadius,
-    initial: () => ({points: []}),
```
heshan0131: this is a initial state for hovered poitns, and is used to calculate hover state. Is there a new prop that replacing it?

src/deckgl-layers/layer-utils/cluster-utils.js
```diff
@@ -41,11 +41,9 @@ export function getGeoJSON(data, getPosition) {
...
-    map: props => props.data,
```
heshan0131: why is this deleted without a replacement?

src/deckgl-layers/layer-utils/cluster-utils.js
akre54: v6.0.0 [removed the initial](https://github.com/mapbox/supercluster/releases/tag/v6.0.0) argument. It now uses the first item as a basis. The hovered points should still work, no?

src/deckgl-layers/layer-utils/cluster-utils.js
akre54: breaking change to the library: https://github.com/mapbox/supercluster/releases/tag/v6.0.0

src/deckgl-layers/layer-utils/cluster-utils.js
heshan0131: According to docs:
```
In most cases, you can simply remove the initial option and things will work. There are some exceptions where you'd need to update your map function to make up for lack of initial, but they should be rare.
```
We should then update the map function, see comment below

heshan0131: originally initial function add an empty array of points. then the reduce function will push props.data to the `points` array. 

We need the property `points` in each cluster to store all the points that it contains. It is used to render hover info. Please make sure they are propogated

src/deckgl-layers/layer-utils/cluster-utils.js
```diff
@@ -41,11 +41,9 @@ export function getGeoJSON(data, getPosition) {
...
     reduce: (accumulated, props) => {
```
heshan0131: you can probably add the initial points here
```
if (!Array.isArray(typeof accumulated.points)) {
  accumulated.points = [];
}
```

src/deckgl-layers/layer-utils/cluster-utils.js
akre54: It pulls the `points` property from the geoJSON object: https://github.com/keplergl/kepler.gl/blob/abe8e9b8e20d4f3b430c2c3c7d7350c838d49907/src/deckgl-layers/layer-utils/cluster-utils.js#L30

As far as I can tell it works correctly? I can see it changing the color of the hovered icon:

kepler-cluster-hover

If I'm missing something, how can I reproduce?

heshan0131: it should show total points correctly
Screen Shot 2019-06-24 at 2 05 56 PM

heshan0131: You should keep the `map: props => props.data` function

PR #582, opened 2019-06-26
Fix #579

darribas:
This patch replaces a print statement coded in Python 2 style with its Python 3 compatible equivalent.

PR #587, opened 2019-06-27
Fixed python3 compatibility and wrong variable in string format

itamar-otonomo:
The parenthesis-less print failed KeplerGl object initialization.


bindings/kepler.gl-jupyter/keplergl/keplergl.py
```diff
@@ -53,7 +53,7 @@ def data_to_json(data, manager):
         return None
     else:
         if type(data) is not dict:
-            print data
+            print(data)
```
heshan0131: #582 already fixed this. Please rebase

PR #595, opened 2019-07-02
[Feat] Create polygons - Geo-Operations 1

macrigiuseppe:
This pull requests introduce a new __Edit__ mode which will allow users to draw custom polygons on a map to be used as layer filter (to be implemented later).

polygons

We support the ability to create/move/delete rectangle or customd shape polygons,

Context menu and delete action
feature-context-menu

Selecting a feature and clicking on delete button will remove the feature.

Layers
Screen Shot 2019-07-18 at 12 37 06 AM

The current implementation supports adding new points to existing features

### Implementation details:
In order to implement the dropdown menu for editing mode i re-used the toolbar implementation currently used for sharing features.
I extracted the react component from sharing and moved into it's own file to be re-used in multiple both sharing and editing mode.

I created a new editor folder containing all react components for the editor mode, including a new context menu action panel.

The __Editor__ component which contains react-map-gl-draw is located in the same MapContainer as deck.gl layer to facilitate logic to switch from editor to read mode 

** There are still few edge cases to handle but the feature is pretty much integrated.


macrigiuseppe:
Depends on https://github.com/uber/nebula.gl/pull/254

macrigiuseppe:
@heshan0131 this is still WIP but it's touching a lof of files and I would love to have an early feedback

docs/user-guides/c-types-of-layers/e-polygon.md
```diff
@@ -1,6 +1,6 @@
-# Polygon
```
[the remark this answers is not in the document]
macrigiuseppe: Yeah, maybe a replace all side effect. I will make sure I revert the right ones

macrigiuseppe:
@heshan0131 Reverted all changes and fixed lint errors and warnings

src/components/map-container.js
```diff
@@ -173,86 +259,8 @@ export default function MapContainerFactory(MapPopover, MapControl) {
     };
 
     /* component render functions */
-    /* eslint-disable complexity */
-    _renderMapPopover() {
```
macrigiuseppe: i converted this method into a react component to take advantage of react render optimization

src/components/map/map-control.js
```diff
@@ -221,9 +237,11 @@ const MapControlPanel = ({children, header, onClick, scale = 1, isExport}) => (
     </StyledMapControlPanelHeader>
     <StyledMapControlPanelContent>{children}</StyledMapControlPanelContent>
   </StyledMapControlPanel>
-);
+));
+
+MapControlPanel.displayName = 'MapControlPanel';
 
-const MapLegendPanel = ({items, isActive, scale, toggleMenuPanel, isExport}) =>
+const MapLegendPanel = React.memo(({items, isActive, scale, onToggleMenuPanel, isExport}) =>
```
macrigiuseppe: React.memo aka PureComponent for stateless react

src/components/modals/export-data-modal.js
```diff
@@ -39,8 +39,7 @@ const propTypes = {
   onClose: PropTypes.func.isRequired,
   onChangeExportSelectedDataset: PropTypes.func.isRequired,
   onChangeExportDataType: PropTypes.func.isRequired,
-  onChangeExportFiltered: PropTypes.func.isRequired,
-  onChangeExportConfig: PropTypes.func.isRequired
```
macrigiuseppe: not used

macrigiuseppe:
@heshan0131 updated pull request, let me know what you think

src/components/common/action-panel.js
```diff
@@ -0,0 +1,162 @@
...
+import Checkbox from 'components/common/switch';
+
+const StyledItem = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  font-size: 12px;
+  line-height: 14px;
+  padding: 8px;
+  height: 32px;
```
heshan0131: `switchHeight` and `switchWidth` is in base.js, this should be added too

src/reducers/combined-updaters.js
```diff
@@ -182,3 +190,19 @@ export const addDataToMapUpdater = (state, {payload}) => {
 };
 
 export const addDataToMapComposed = addDataToMapUpdater;
+
+export const setFeaturesUpdater = (state, payload) => ({
+  ...state,
+  visState: visStateFeaturesUpdater(state.visState, payload),
```
heshan0131: we don't need a composed updater if an action is handled in each subreducer independently. You can just add `[SET_FEATURE]` action handlers to `visState` and `uiState` reducer
```
visState = {
  ....
  [ActionTypes.SET_FEATURE]: visStateFeaturesUpdater
}

uiState = {
  ....
  [ActionTypes.SET_FEATURE]: uiStateFeaturesUpdater
}
```

The time when we need to use combined updaters is when 1 action needs to be handled in multiple reducer is a certain sequence. e.g. in `updateVisData` we use layers created by `visState` to update viewport in `mapState1`

src/components/map-container.js
```diff
@@ -451,11 +486,22 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
+            layers={layers}
+            mapLayers={mapLayers}
+            mousePos={mousePos}
+          />
         </StyledMapContainer>
       );
     }
   }
 
+  MapContainer.displayName = 'MapContainer';
```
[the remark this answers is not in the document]
macrigiuseppe: eslint was complaining about all missing display names

src/components/common/toolbar-item.js
```diff
@@ -0,0 +1,43 @@
...
+    e.stopPropagation();
+    e.preventDefault();
+    onClick(e);
+  }}>
+    {icon}
+    <div className="save-export-dropdown__title">{label}</div>
+  </StyledDiv>
+));
+
+ToolbarItem.displayName = 'ToolbarItem';
```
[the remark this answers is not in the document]
macrigiuseppe: indeed an issue with React https://github.com/facebook/react/issues/14319.

I tried to use react-tools on my browser and this is what i have:
Screen Shot 2019-07-21 at 10 22 52 PM

src/components/common/toolbar-item.js
```diff
@@ -0,0 +1,43 @@
...
+import styled from 'styled-components';
+
+const StyledDiv = styled.div`
+  color: ${props => props.active ?
+  'white' : props.theme.textColor
+  };
+`;
+
+const ToolbarItem = React.memo(({active, className, icon, label, onClick}) => (
+  <StyledDiv active={active} className="save-export-dropdown__item" onClick={(e) => {
```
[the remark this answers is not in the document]
macrigiuseppe: done

src/components/common/action-panel.js
macrigiuseppe: done

src/components/common/toolbar-item.js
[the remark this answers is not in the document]
macrigiuseppe: same thing for ActionPanelItem:
Screen Shot 2019-07-21 at 10 32 20 PM

I wonder what could be the difference

src/components/map-container.js
```diff
@@ -53,10 +60,86 @@ const MAPBOXGL_RENDER = 'render';
...
 
 MapContainerFactory.deps = [
-  MapPopoverFactory, MapControlFactory
+  MapPopoverFactory,
+  MapControlFactory
 ];
 
 export default function MapContainerFactory(MapPopover, MapControl) {
+  /* eslint-disable complexity */
+  const MapTooltip = React.memo(({
```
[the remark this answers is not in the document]
macrigiuseppe: Done

src/components/map-container.js
```diff
@@ -425,23 +441,42 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
+            onSetEditorMode={uiStateActions.setEditorMode}
           />
-          <MapComponent
-            {...mapProps}
-            key="bottom"
-            ref={this._setMapboxMap}
-            mapStyle={mapStyle.bottomMapStyle}
-            getCursor={this.props.hoverInfo ? () => 'pointer' : undefined}
-            transitionDuration={TRANSITION_DURATION}
-            onMouseMove={this.props.visStateActions.onMouseMove}
```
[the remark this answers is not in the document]
macrigiuseppe: I think i accidentally removed it when i rebased. Is back now

src/components/map-container.js
```diff
@@ -53,10 +60,86 @@ const MAPBOXGL_RENDER = 'render';
...
+      position = this._getHoverXY(viewport, lngLat);
+    }
+    return (
+      <div>
+        <MapPopover
+          {...position}
+          layerHoverProp={layerHoverProp}
+          coordinate={interactionConfig.coordinate.enabled && ((pinned || {}).coordinate || coordinate)}
+          freezed={Boolean(clicked || pinned)}
+          onClose={this._onCloseMapPopover}
```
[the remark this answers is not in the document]
macrigiuseppe: done

src/components/map-container.js
```diff
@@ -53,10 +60,86 @@ const MAPBOXGL_RENDER = 'render';
...
+          layer
+        }
+      }
+    }
+
+    if (pinned || clicked) {
+      // project lnglat to screen so that tooltip follows the object on zoom
+      const viewport = new WebMercatorViewport(mapState);
+      const lngLat = clicked ? clicked.lngLat : pinned.coordinate;
+      position = this._getHoverXY(viewport, lngLat);
```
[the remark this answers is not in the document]
macrigiuseppe: done

src/components/kepler-gl.js
```diff
@@ -244,6 +244,8 @@ function KeplerGlFactory(
         mapboxApiAccessToken,
         mapboxApiUrl,
         mapState,
+        uiState,
+        visState,
```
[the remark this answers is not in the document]
macrigiuseppe: done

src/reducers/combined-updaters.js
macrigiuseppe: i initially started with independent actions and I was trying to see whether there was a performance improvement by using a composed action. But at the end i don't think it made any  different

macrigiuseppe: Done

macrigiuseppe:
@heshan0131 update with your feedback. Waiting on https://github.com/uber/nebula.gl/pull/260 as well for better interaction handler

heshan0131:
Selection works after yarn installed latest react-map-gl-draw

Selected polygon should show solid line instead of dash
Image Pasted at 2019-7-22 10-15

heshan0131:
After drawing polygon, select rectangle, draw and drag doesn't create rectangle
bug_rect

heshan0131:
The mouse cursor should be `crosshair` in draw mode. After selecting a feature, the mouse cursor should change to `move`, or `pointer` when a corner is hovered

macrigiuseppe:
> Selection works after yarn installed latest react-map-gl-draw
> 
> Selected polygon should show solid line instead of dash
> Image Pasted at 2019-7-22 10-15

@heshan0131 

Polygons are turned solid when selected

selected-solid-feature

macrigiuseppe:
> crosshair

@heshan0131 Updated cursor pointers

cursor-pointer

macrigiuseppe:
> After drawing polygon, select rectangle, draw and drag doesn't create rectangle
> bug_rect

@heshan0131 I think i know the reason. 

With react-map-gl-draw in order to create a rectangle you can't simply click and drag but you have to click first to define your first vertex, drag your mouse and click to define the opposite vertex. @xintongxia can you validate the above statement? thank you

macrigiuseppe:
Updated to the latest react-map-gl-draw

xintongxia:
> > After drawing polygon, select rectangle, draw and drag doesn't create rectangle
> > bug_rect
> 
> @heshan0131 I think i know the reason.
> 
> With react-map-gl-draw in order to create a rectangle you can't simply click and drag but you have to click first to define your first vertex, drag your mouse and click to define the opposite vertex. @xintongxia can you validate the above statement? thank you

@macrigiuseppe 

Yes, drawing rectangle flow is like the way you described.
Users have to click (start drawing a rectangle) then drag the mouse and click again (finish the rectangle).

macrigiuseppe:
@shan this is rebased was rebased with master and multiple dataset feature. New tests are added in the second pull requests: #678

src/components/common/checkbox.js
```diff
@@ -93,14 +93,16 @@ export default class Checkbox extends Component {
...
+
     const labelProps = {
       ...pick(this.props, ['checked', 'disabled', 'secondary']),
       htmlFor: this.props.id
     };
 
     const LabelElement = this.props.type === 'checkbox' ? StyledCheckboxInput : StyledSwitchInput;
     return (
-      <StyledCheckbox className="kg-checkbox">
+      <StyledCheckbox className={`kg-checkbox ${className}`}>
```
heshan0131: You can use `classnames` to handle the composition of classNames, to avoid blank spaces, it ignores falsy value
`className={classnames('kg-checkbox', className)}`, no need to call `const {className = ''} = this.props;`

src/components/common/icons/base.js
```diff
@@ -51,15 +51,18 @@ export default class Base extends Component {
     width: null,
     viewBox: '0 0 64 64',
     predefinedClassName: '',
-    className: ''
+    className: '',
```
heshan0131: setting `fill: 'currentColor'` here means if user pass in a style object, it will override `fill` prop. use `style.fill = 'currentColor';` in render simply adds it to user defined value

src/components/common/icons/rectangle.js
```diff
@@ -0,0 +1,51 @@
...
+    /** Set the height of the icon, ex. '16px' */
+    height: PropTypes.string,
+    predefinedClassName: PropTypes.string,
+    viewBox: PropTypes.string,
+    style: PropTypes.object
+  };
+
+  static defaultProps = {
+    height: '16px',
+    predefinedClassName: 'data-ex-icons-polygon',
```
heshan0131: `'data-ex-icons-rectangle'`

src/components/common/styled-components.js
```diff
@@ -374,8 +374,12 @@ export const StyledModalContent = styled.div`
  * This workaround will hide the error banner.
  */
 export const StyledMapContainer = styled.div`
-  .mapboxgl-map .mapboxgl-missing-css {
-    display: none;
+  .mapboxgl-map {
+    z-index: -1;
```
heshan0131: why `z-index: -1;` here?

src/components/common/toolbar.js
```diff
@@ -0,0 +1,84 @@
...
+      &:last-child {
+        border-bottom: 0;
+      }
+      ` : null
+    };
+  }
+`;
+
+const Toolbar = React.memo(({children, className, show, direction = 'row'}) => (
+  <StyledPanelDropdown className={`${className || ''} save-export-dropdown`} show={show} direction={direction}>
```
heshan0131: same here about using ` classnames`

src/components/map-container.js
```diff
@@ -459,6 +470,22 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
+              datasets={datasets}
+              editor={uiState.editor}
+              features={editor.features}
+              isEnabled={isEdit}
+              layers={layers}
+              onDeleteFeature={uiStateActions.deleteFeature}
+              onSelect={uiStateActions.setSelectedFeature}
+              onUpdate={visStateActions.setFeatures}
+              style={{zIndex: isEdit ? 0 : -1}}
+              onToggleFeatureLayer={visStateActions.toggleFeatureLayer}
```
heshan0131: I can't find `toggleFeatureLayer` is not defined in vis-state-actions, which causes this error:

Screen Shot 2019-12-05 at 3 39 25 PM

src/components/map-container.js
```diff
@@ -459,6 +470,22 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
+            <Draw
+              datasets={datasets}
+              editor={uiState.editor}
+              features={editor.features}
+              isEnabled={isEdit}
+              layers={layers}
+              onDeleteFeature={uiStateActions.deleteFeature}
+              onSelect={uiStateActions.setSelectedFeature}
+              onUpdate={visStateActions.setFeatures}
+              style={{zIndex: isEdit ? 0 : -1}}
```
heshan0131: making zIndex to -1 when !isEdit cauing deckgl overlay showing on top of the draw element
Screen Shot 2019-12-05 at 3 58 53 PM

src/components/map/map-control.js
```diff
@@ -162,12 +187,109 @@ const MapLegendPanel = ({layers, isActive, scale, toggleMenuPanel, isExport}) =>
...
+  position: absolute;
+  right: 32px;
+`;
+
+const MapDrawPanel = React.memo(({editor, isActive, onToggleMenuPanel, onSetEditorMode}) => {
+  return (
+    <div style={{position: 'relative'}}>
+      {isActive ? (
+        <StyledToolBar show={isActive} direction="column">
+          <ToolbarItem
```
heshan0131: Now that clicking on polygon automatically selectes it. Do we still need this to manually enter selection mode?

src/components/map/map-control.js
```diff
@@ -162,12 +187,109 @@ const MapLegendPanel = ({layers, isActive, scale, toggleMenuPanel, isExport}) =>
...
+            onClick={() => onSetEditorMode(EDITOR_MODES.DRAW_POLYGON)}
+            label="polygon"
+            icon={(<Polygon height="22px"/>)}
+            active={editor.mode === EDITOR_MODES.DRAW_POLYGON}
+          />
+          <ToolbarItem
+            onClick={() => onSetEditorMode(EDITOR_MODES.DRAW_RECTANGLE)}
+            label="rectangle"
+            icon={(<Rectangle height="22px"/>)}
```
heshan0131: A better way is to pass in `Rectangle` and render it in `<ToolbarItem><props.icon height="22px"></ToolbarItem>`

src/components/map/map-tooltip.js
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
heshan0131: Is this used anywhere?

src/reducers/combined-updaters.js
heshan0131: you are still creating `setFeaturesComposed` and `deleteFeatureComposed`.  this 2 actions don't need to be composed here. They can be handled independently inside each (visState, uiState) reducer...

src/components/map-container.js
heshan0131: oh i guess it's in geo-operation 2 pr?

src/components/common/icons/base.js
macrigiuseppe: The problem with style.fill = 'currentColor' is that i cannot override the default value;
I need a way to override the fill property.

we can do something like
```
style = {}
... } = this.props;

style.fill = style.fill || 'currentColor'
```

or I can create a helper to merge defaults values when no properties exist in the original object

src/components/common/styled-components.js
macrigiuseppe: this is due to an issue with deckgl where deckgl capture all mouse events despite of other existing layers and their z-index value. So I have to make sure deckgl and its container mapbox-gl has -1 when in editor mode in order to be able to use the mouse to draw polygons.

When in regular mode we have the following:
mapboxgl: -1
deckgl: 0
editor: -1

In editor mode:
mapboxgl: -1
deckgl: -1
editor: 0

src/components/map-container.js
macrigiuseppe: changing the z-index, in this case, will not fix the issue because all layers are rendered onto a canvas and the editor  is a regular html element.
In order to have the effect you describe we need to absolute position the editor container but in that case deck.gl will not receive mouse over and the tooltip will not show up

src/components/map-container.js
macrigiuseppe: it's in geo-operation 2 branch

src/components/map/map-control.js
macrigiuseppe: We need this for two reasons:
- enable drawing on the map
- the selection of a polygon only works when we are in editor mode. when we turn editor mode off we are not able to select polygons by clicking on it

src/components/map/map-control.js
macrigiuseppe: how would we handle all toolbarItems, each has a different icon.

Should we pass a default prop items with a list of button objects? something like this:

```
{
  onClick: () => onSetEditorMode(EDITOR_MODES.DRAW_POLYGON),
  label:'polygon',
  icon: Rectangle,
  activeCondition: EDITOR_MODES.DRAW_RECTANGLE
}
```

src/components/map/map-tooltip.js
macrigiuseppe: I forgot to add this to MapContainer. this will replace _renderMapPopover method

macrigiuseppe: I will remove it from this diff

src/reducers/combined-updaters.js
macrigiuseppe: The idea behing this actions is to be used the way we use addDataToMapComposed, directly accessing kepler.gl store rather than using actions

An application can call these composed updaters when it is directly kepler.gl and want to modify the features. I am not using them at all and since i changed the logic in the second PR i can get rid of them
i am not even using those actions anywhere, i removed them

macrigiuseppe:
@heshan0131 I found a more simple and elegant solution for z-index and mouse events.

The Draw component will be placed using absolute position over deck.gl (canvas). 

Draw
-----
Deck.gl

Modes
- Read: Draw pointer-events css property will set to 'none' so deck.gl will capture mouse events and display tooltip correctly.
- Edit:  Draw pointer-events css property will set to 'all', users will be able to draw and apply filters.

src/components/map-container.js
macrigiuseppe: I removed it from this PR

PR #622, opened 2019-07-22
Switch from callback refs to createRef

akre54:
`createRef` is the preferred way of interacting with the DOM in React. This updates all callback refs to use the newer API.

### DIFF
--- a/src/components/common/time-slider-marker.js
+++ b/src/components/common/time-slider-marker.js
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React, {Component} from 'react';
+import React, {Component, createRef} from 'react';
 import PropTypes from 'prop-types';
 import {scaleUtc} from 'd3-scale';
 import {select} from 'd3-selection';
@@ -79,6 +79,8 @@ export default class TimeSliderMarker extends Component {
     }
   }
 
+  xAxis = createRef();
+
   domainSelector = props => props.domain;
   widthSelector = props => props.width;
   scaleSelector = createSelector(
@@ -101,12 +103,8 @@ export default class TimeSliderMarker extends Component {
       .tickSize(8)
       .tickPadding(6);
 
-    const svg = select(this.svgContainer);
-
-    svg
-      .select('.x.axis')
-      .call(xAxis)
-      .selectAll('text');
+    select(this.xAxis.current)
+      .call(xAxis);
   }
 
   render() {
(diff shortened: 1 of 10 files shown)


src/components/map/layer-hover-info.js
```diff
@@ -17,6 +17,7 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+
```
akre54: I also ran lint

heshan0131:
@akre54 I rebased with master, build still seem to fail
https://deploy-preview-622--keplergl.netlify.com/demo

Can you make sure all the components you touched in this branch works?

CharlieMartell:
I am seeing some issues related to the ref assigned to the typeahead component I believe this will fix `Uncaught TypeError: this.entry.focus is not a function`. Would be great to get this in :)

akre54:
Sorry -- I made a change and missed that one.

The places changed in this PR are:
* File upload drag-n-drop
* File upload button
* Typeahead
* Range brush
* Range slider
* Slider
* Time slider
* Modal `rootNode`-- maybe we should use a [portal](https://reactjs.org/docs/portals.html) for this
* Map popover
* Plot container image capture -- currently only capturing map, no data overlay (related to #656 ?)

I went through these scenarios and they all seem fine with the exception of image capture, but would love another look.

macrigiuseppe:
> I am seeing some issues related to the ref assigned to the typeahead component I believe this will fix `Uncaught TypeError: this.entry.focus is not a function`. Would be great to get this in :)

We found out the problem was related to styled-components outdate version on the consumer app

akre54:
@macrigiuseppe @heshan0131 how are we looking for this?

macrigiuseppe:
> @macrigiuseppe @heshan0131 how are we looking for this?

Hey @akre54, can you rebase with the current version of master. @heshan0131 just fixed the export image feature and you will be able to test it the export image flow as well

src/components/common/time-slider-marker.js
```diff
@@ -115,11 +113,8 @@ export default class TimeSliderMarker extends Component {
         className="time-slider-marker"
         width={this.props.width}
         height={height}
-        ref={comp => {
-          this.svgContainer = comp;
-        }}
       >
-        <g className="x axis" transform="translate(0, 0)" />
+        <g ref={this.xAxis} transform="translate(0, 0)" />
```
heshan0131: Need the className here for style

heshan0131:
Seeing some error here:

1. File upload drag-n-drop
When drag-n-drop a file into kepler, like Add Data button again. File drag-n-drop disappear
Screen Shot 2019-08-09 at 1 48 11 PM

2. Time slider axis color should be white, not is black, missing className
Screen Shot 2019-08-13 at 12 24 04 PM

akre54:
Fixed! Thanks for pointing out

heshan0131:
Looks Good. Please commit with sign off to pass DCO

Issue #539, opened 2019-06-01
Export Image Refreshing Without Heatmap Layer


sleepybjr:
**Describe the bug**
When clicking on "Export Image", the preview image constantly refreshes without loading the heatmap layer onto it. When I am able to click download, there is no heatmap layer visible on the image.

**To Reproduce**
1. Upload latitude/longitude data.
2. Add a new heatmap layer using that data.
3. Reduce radius to < 6.
4. Select Export Image.
5. See error (reproduced at every refresh)
![image](https://user-images.githubusercontent.com/22383876/58743915-5d76cd00-83ef-11e9-9394-80b32c6a357f.png)

**Expected behavior**
When clicking on "Export Image", the visible heatmap layer will be visible on the preview and after downloading.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: windows 10
 - Browser: chrome, firefox, firefox dev edition
 - Version: latest for all

**Additional context**
Tried it on live demo site and cloned repo, occurred in both.

cschlessinger:
Hi! Wondering if there's an ETA for the fix?

macrigiuseppe:
We are currently looking into this issue

macrigiuseppe:
hi @cschlessinger we are currently looking into this issue

heshan0131:
This is fixed with release v1.1.2

Issue #466, opened 2019-04-05
Fix heatmap issue 


heshan0131:
**Describe the bug**
Only one heatmap can be created at a time**To Reproduce**

**Expected behavior**
User should be able to create multiple heatmap

PR #642, opened 2019-07-28
[Enhancement] Simplify map layer visible logic in splitMaps and deck, mapbox overlay renders

heshan0131:
- Simplify `splitMaps` structure to `{layers: {[id]: true | false}`, visible layers will have an entry, true or false value indicates their visibilities in each split map panel
- Add `layersToRender` selector to consolidate layer visibility logic
- Add `mapboxLayersToRender` selector to consolidate mapbox layer render logic


src/actions/vis-state-actions.js
```diff
@@ -404,22 +404,6 @@ export function toggleLayerForMap(mapIndex, layerId) {
...
 
-/**
- * Set layers to be visible in split map
- * @memberof visStateActions
- * @param {Number} mapIndex - index of the split map
- * @param {Array<string>} layerIds - array of layer ids
- * @returns {{type: ActionTypes.SET_VISIBLE_LAYERS_FOR_MAP, layerIndex: *, mapIndex: *}}
- * @public
- */
-export function setVisibleLayersForMap(mapIndex, layerIds) {
```
heshan0131: not used anywhere

src/components/map-container.js
```diff
@@ -108,7 +110,35 @@ export default function MapContainerFactory(MapPopover, MapControl) {
       }
     }
 
+    layersSelector = props => props.layers;
+    layerDataSelector = props => props.layerData;
+    mapLayersSelector = props => props.mapLayers;
+    layerOrderSelector = props => props.layerOrder;
+    layersToRenderSelector = createSelector(
```
heshan0131: single point of entry to determine which layer should be rendered based on `mapLayer` in splitMaps and `layer.shouldRenderLayer` logic

src/components/map-container.js
```diff
@@ -288,37 +302,27 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
       };
 
-      if (!this._shouldRenderLayer(layer, data, mapLayers)) {
-        return overlays;
-      }
-
-      let layerOverlay = [];
-
       // Layer is Layer class
-      if (typeof layer.renderLayer === 'function') {
```
heshan0131: This check is moved to `base-layer` class

heshan0131:
https://github.com/keplergl/kepler.gl/issues/539

Fixed:
Screen Shot 2019-07-28 at 10 40 01 PM

heshan0131:
https://github.com/keplergl/kepler.gl/issues/466

src/layers/heatmap-layer/heatmap-layer.js
```diff
@@ -150,26 +152,26 @@ class HeatmapLayer extends MapboxGLLayer {
...
 
-    (datasetId, isVisible, visConfig, weightField, weightDomain) => {
+    (datasetId, columns, visConfig, weightField, weightDomain) => {
 
-      const layer = {
+      return {
         type: 'heatmap',
         id: this.id,
-        source: datasetId,
+        source: `${datasetId}-${columns}`,
```
heshan0131: Setting source to `datasetId` plus `columns` so that multiple heatmap based on the same dataset but different column can be created

PR #649, opened 2019-08-02
[Enhancement] Use preserved state to apply keplerGlInit. when mint=false

heshan0131:
- When mint=false, kepler.gl will use the existing state instance. However, this also causes the `mapboxApiAccessToken` `mapboxApiUrl` and `mapStylesReplaceDefault` not being store in the reducer.
- We will call `keplerGlInit` on existing state as well. The `INIT` action is only listened by `map-style` reducer and is only used to store the 3 props above.

### DIFF
--- a/src/constants/default-settings.js
+++ b/src/constants/default-settings.js
@@ -254,6 +254,7 @@ export const SCALE_FUNC = {
   sqrt: require('d3-scale').scaleSqrt,
   point: require('d3-scale').scalePoint
 };
+
 export const ALL_FIELD_TYPES = keyMirror({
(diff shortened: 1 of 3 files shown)


heshan0131:
@Firenze11

src/constants/default-settings.js
```diff
@@ -254,7 +254,6 @@ export const SCALE_FUNC = {
   sqrt: require('d3-scale').scaleSqrt,
   point: require('d3-scale').scalePoint
 };
-
```
macrigiuseppe: i would keep the empty line here

PR #670, opened 2019-08-07
Add Log and Sqrt scale

akre54:
**Work in progress**

In this pull:
* Add `log` scale type and guard (see notes)
* `AggrColorScaleSelector` => `AggrScaleSelector`
* Add `jenks` scale type
* All scales live on a single object in default-settings.js. Ideally can be overridden if need be.

To do:
* The Jenks algorithm is very slow for datasets over a couple hundred points. I assume this is due to using 2d array-of-arrays. Investigate typed arrays or possibly a more generalized k-means algorithm (h/t @sdkennedy who pointed out that jenks is basically just k-means in 1d)
* Need to add height scales to other aggregate layers (e.g. hexagon layer)

Notes: 
* Because `log(0)` is undefined (`scaleLog` returns `NaN` when `domain[0]` is `0`, `Math.log(0)` returns `-Infinity`), we need to guard on this. Would love any feedback on a better solution.


src/deckgl-layers/layer-utils/utils.js
```diff
@@ -84,9 +82,18 @@ export function getColorScaleFunction(layer) {
     .range(layer.props.colorRange);
 }
 
+export function getElevationScaleFunction(layer) {
+  const elevationRange = layer.props.elevationRange;
+  const elevationDomain = layer.props.elevationDomain || layer.state.elevationValueDomain;
+
+  layer.state.sizeScaleFunc = getScaleFunctor(layer.props.sizeScale)()
```
akre54: This is confusing because `deck.gl` uses `elevationScale` to reference a scalar value that each height is multiplied by.

Most other places in Kepler `___Scale` refers to one of the `SCALE_TYPES`. Is there any way we can explicitly call out this difference?

heshan0131:
Using the San Francisco elevation contour dataset, try to set the color using jenkins, i am still seeing this `jenks.js:35 Uncaught TypeError: Cannot set property '1' of undefined` error
Screen Shot 2019-08-07 at 3 32 45 PM
Screen Shot 2019-08-07 at 3 32 39 PM

src/deckgl-layers/layer-utils/utils.js
heshan0131: __scale for value calculation in kepler.gl layer is stored in layer.config

__scale passed to deck.gl as multiplier in is store in layer.config.visConfig.

So at least they won't override each other

You can us `scaleScleType` or `elevationScaleType` as layer props to pass it to deck.gl

src/layers/base-layer.js
```diff
@@ -776,6 +778,14 @@ export default class Layer {
       case SCALE_TYPES.quantile:
         return getQuantileDomain(filteredIndexForDomain, indexValueAccessor, sortFunction);
 
+      case SCALE_TYPES.jenks:
+        // TODO: Is this right? Do we need to be more generic?
+        const nClasses = this.config.visConfig[visualChannel.range].colors.length;
```
heshan0131: this is fine for now

akre54:
I think I'm going to split out the Jenks implementation from this pull because it seems buggy. Can go into details on another PR

test/node/reducers/vis-state-test.js
```diff
@@ -384,7 +384,7 @@ test('#visStateReducer -> LAYER_TYPE_CHANGE.2', t => {
   );
   const newLayer2 = nextState2.layers[0];
   t.equal(newLayer2.config.sizeField, intField, 'should update sizeField');
-  t.equal(newLayer2.config.sizeScale, 'sqrt', 'should scale to linear');
+  t.equal(newLayer2.config.sizeScale, 'sqrt', 'should scale to sqrt');
```
akre54: Not sure why this changed @heshan0131 ....

src/utils/data-scale-utils.js
```diff
@@ -62,3 +62,15 @@ export function getLinearDomain(data, valueAccessor) {
...
   return range.map((d, i) => (d === undefined ? i : d));
 }
+
+/**
+ * return linear domain for an array of data. A log scale domain cannot contain 0
+ * @param {Array} data
+ * @param {function} valueAccessor
+ * @returns {Array} domain
+ */
+export function getLogDomain(data, valueAccessor) {
```
akre54: This is definitely a hack. Should we wrap the log scale instead?

heshan0131: No I think this is fine

test/node/reducers/vis-state-test.js
heshan0131: the comment in test is probably copied.

test/node/reducers/vis-state-test.js
```diff
@@ -421,7 +421,7 @@ test('#visStateReducer -> LAYER_TYPE_CHANGE.2', t => {
   );
   t.equal(
     newLayer3.config.sizeScale,
-    'linear',
```
heshan0131: this should be `linear`

heshan0131:
rebase with master so the travis-ci tests will run.
Also you could rename this PR to add `Log` and `Sqrt` Scale

PR #671, opened 2019-08-07
[FEAT] Add trip layer - 2

yanhann10:
- Add speed control
- Add documentation
- Refine animation slider logic
- Update in-layer and global config animation domain upon adding/removing data


yanhann10:
https://github.com/keplergl/kepler.gl/issues/397

src/layers/geojson-layer/geojson-utils.js
```diff
@@ -185,3 +186,84 @@ export function getGeojsonFeatureTypes(allFeatures) {
...
+
+export function containValidTime(data) {
+  const formattedData = [{...data}];
+  const analyzedType = Analyzer.computeColMeta(formattedData);
+  if (analyzedType.find(d => d.category !== 'TIME')) {
+    return false;
+  } else if (['x', 'X'].includes(analyzedType[0].format)) {
+    return 'unix';
+  }
+  return 'dateTime';
```
heshan0131: this function should return `analyzedType` which contain the `type` and `format`

src/layers/geojson-layer/geojson-utils.js
```diff
@@ -185,3 +186,84 @@ export function getGeojsonFeatureTypes(allFeatures) {
...
+}
+
+/**
+ * Get unix timestamp from animatable geojson for deck.gl trip layer
+ * @param {array} features array of geojson feature objects
+ * @returns {} unix timestamp
+ */
+export function dataToTimeStamp(features) {
+  const unixTime =
+    containValidTime(features[0].geometry.coordinates[0]) === 'dateTime'
```
heshan0131: You have many functions repeated here: `features.map(f =>  f.geometry.coordinates.map(coord =>`, `coord => (coord.length === 4 ? coord[3]`

Consider only call it once but apply a `getTimeStamp` functiion to it,

Also use returned `type` and `format` to convert timestamp.
```
  const analyzedType = containValidTime(features[0].geometry.coordinates[0]);
  const {format, type) = analyzedType;
  
  function getTimeVaule = (coord) => {
    if (coord.length === 4 && notNullorUndefined(coord[3])) {
      return typeof coord[3] === 'string'
        ? moment.utc(coord[3], format).valueOf()
        : format === 'x' ? coord[3] * 1000 : coord[3];
    }
  }

 return  features.map(f => f.geometry.coordinates.map(getTimeVaule))
```

src/reducers/vis-state-updaters.js
```diff
@@ -634,44 +634,125 @@ export const updateAnimationSpeedUpdater = (state, action) => ({
...
+});
+
+function updateAnimationDomain(state) {
+  // merge all animatable layer domain calculate the union
+  // state.layers.config.animation.domain
+  // take state.layer
+  const animatableLayers = state.layers.filter(l => l.config.animation.enabled);
+  // const domains = state.layers.map(l => l.config.animation.domain);
+
+  const minRange = [];
```
heshan0131: the whole thing can be implemented with 1 reduce function:
```
const mergedDomain = animatableLayers.reduce((accu, layer) => (
    [
      Math.min(accu[0], layer.config.animation.domain[0],
      Math.max(accu[1], layer.config.animation.domain[1]
    ]
  ), [+Infinity, -Infinity]);
```

src/reducers/vis-state-updaters.js
```diff
@@ -634,44 +634,125 @@ export const updateAnimationSpeedUpdater = (state, action) => ({
...
+ * @public
+ *
+ */
+
+export const enableLayerAnimationUpdater = (state, oldLayer) => {
+  // calculate animation domain for the layer enabled
+  // layer.config.animation.domain
+  // update newLayer into the state.layers
+  // merge all animatable layer domain calculate the union
+  const [minTs, maxTs] = getTimeAnimationDomainForTripLayer(
```
heshan0131: if you are not use `minTx` and `maxTs` separately, don't pull it out

`const domain = getTimeAnimationDomainForTripLayer`

src/reducers/vis-state-updaters.js
```diff
@@ -634,44 +634,125 @@ export const updateAnimationSpeedUpdater = (state, action) => ({
...
+    oldLayer[0],
+    state.datasets
+  );
+
+  oldLayer[0].updateLayerConfig({
```
heshan0131: Need to add updated `oldLayer` to state.layers

src/reducers/vis-state-updaters.js
```diff
@@ -1408,9 +1515,13 @@ export function updateAllLayerDomainData(state, dataId, newFilter) {
...
-  return {
+  const newState = {
     ...state,
     layers: newLayers,
     layerData: newLayerDatas
   };
+
+  // return newState;
+  // TODO: H
+  return updateAnimationDomain(newState);
```
heshan0131: No need to call it again here
`enableLayerAnimationUpdater` should call `updateAnimationDomain` already

src/utils/layer-utils/layer-utils.js
```diff
@@ -100,20 +100,38 @@ export function getLightSettingsFromBounds(bounds) {
...
-  // TODO: H factor in more input format beyond 3rd coord as ts
-  const timeField = dataContent
-    .map(d => d[geojsonFieldIdx].geometry.coordinates.map(coord => coord[3]))
-    .flat();
+  // TODO: factor in more input format beyond 3rd coord as ts
 
-  return extent(timeField);
+  const minTs = [];
+  const maxTs = [];
+  dataContent.map(d => {
```
heshan0131: Don't call `extent` 3 times. it's expensive
You can adapt the `reduce` function I wrote in `updateAnimationDomain` here

src/utils/layer-utils/layer-utils.js
```diff
@@ -100,20 +100,38 @@ export function getLightSettingsFromBounds(bounds) {
...
+    minTs.push(featureTimeRange[0]);
+    maxTs.push(featureTimeRange[1]);
+  });
+  return [extent(minTs)[0], extent(maxTs)[1]];
 }
 
-export function getTimeAnimationDomain(layer, datasets) {
-  // TODO: H factor in multiple layers
-  return getTimeAnimationDomainFoTripLayer(layer, datasets);
+export function getTimeAnimationDomain(layers, datasets) {
```
heshan0131: If this function used anywhere?

src/reducers/vis-state-updaters.js
yanhann10: thanks for reviewing. modified accordingly.

docs/user-guides/c-types-of-layers/k.trip.md
```diff
@@ -0,0 +1,58 @@
...
+In order to animate the path, the geometry needs to contain `LineString` feature type and the coordinates need to have 4 elements `[longitude, latitude, altitude, timestamp]`.
+
+```
+{
+  type: 'FeatureCollection',
+  features: [
+    {
+      geometry: {
+        type: 'LineString',
+        coordinates: [
```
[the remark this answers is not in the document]
heshan0131: Make sure your GeoJSON is the correct format according to https://geojson.org/
```
{
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [125.6, 10.1]
  },
  "properties": {
    "name": "Dinagat Islands"
  }
}
```

docs/user-guides/c-types-of-layers/k.trip.md
```diff
@@ -0,0 +1,59 @@
...
+
+![Trip layer](https://d1a3f4spazzrp4.cloudfront.net/kepler.gl/documentation/k-trip.gif 'Grid layer')
+Trip layer can display animated path.
+
+### How to use trip layer to animate path
+
+**Data format**
+Currently trip layer support a special `geoJSON` format where the coordinate `linestring` has a 4th element denoting timestamp.
+
+In order to animate the path, the `geoJSON` data needs to contain `LineString` in its feature geometry, and the coordinates in the `LineString` need to have 4 elements in the format of  `[longitude, latitude, altitude, timestamp]` with the last element being a valid unix timestamp.
```
heshan0131: unix timestamp in miliseconds? or seconds? You can link to
https://www.unixtimestamp.com/index.php

docs/user-guides/c-types-of-layers/k.trip.md
```diff
@@ -0,0 +1,59 @@
...
+          [-74.20987, 40.81765, 0, 1564184396],
+          [-74.20998, 40.81746, 0, 1564184409]
+        ]
+      }
+    }
+  ]
+}
+```
+
+Support for more data formats such as csv will be added in future releases.
```
heshan0131: Highlight this with 
__Note__

src/components/common/animation-control/animation-slider.js
```diff
@@ -127,9 +114,17 @@ const AnimationControlFactory = () => {
     };
 
     _nextFrame = () => {
-      const {domain} = this.props.animation;
-      this._currentStep > domain[1] - domain[0] - 1 ? 0 : (this._currentStep += 3);
-      this.onSlider1Change(this._currentStep);
+      const {currentTime, domain, speed} = this.props.animation;
```
heshan0131: `this.animation.speed` is a multiplier. Copy the logic from `time-range-slider`. This should be:
```
    const speed = ((domain[1] - domain[0]) / BASE_SPEED) * this.props.speed;
```

Also, the logic here is hard to follow, simplify it to:

```
// if nextTime exceed domain[1], set to back to domain[0]
const nextTime = currentTime + speed > domain[1] ? domain[0] : currentTime + speed;
this.props.playAnimation(nextTime);
```
That's it, you don't need a _pauseAnimation case

src/components/common/animation-control/animation-slider.js
```diff
@@ -141,9 +136,22 @@ const AnimationControlFactory = () => {
       this.setState({isAnimating: false});
     };
 
+    toggleSpeedControl = () => {
+      this.setState({showSpeedControl: !this.state.showSpeedControl});
+    };
+
+    _updateSpeed = speed => {
```
heshan0131: this function is redundant
You can just directly pass it to SpeedControl
```js
updateAnimationSpeed={this.props.updateSpeed}
```
Also why the actual action is called `updateAnimationSpeed` in `visState`, but the prop in `AnimationSlider` here is `updateSpeed` and the prop in `SpeedControl` is `updateAnmationSpeed`? Please name them consistently

src/components/common/animation-control/speed-control.js
```diff
@@ -0,0 +1,59 @@
...
+
+`;
+
+const StyledSpeedText = styled.div`
+  display: inline-block,
+  width: 27px
+`;
+
+class SpeedControl extends Component {
+  state = {};
```
heshan0131: You don't need `state = {}` here. Make it a `PureComponent`

```
const SpeedControl = ({onClick, updateAnimationSpeed, speed, showSpeedControl}) => (
    <StyledSpeedToggle>
    ...
    </StyledSpeedToggle>
);
```

src/layers/trip-layer/trip-layer.js
```diff
@@ -215,7 +200,8 @@ export default class TripLayer extends Layer {
 
     return {
       data: geojsonData,
-      getFeature,
+      getPath: d => d.geometry.coordinates.map(coord => coord.slice(0, 2)),
```
heshan0131: You can probably pass `d => d.geometry.coordinates`, no need to map it. the 4th element is not used in deck.gl. Let me know if this doesn't work

src/components/common/animation-control/animation-slider.js
yanhann10: for trip side it's now called updateLayerAnimationSpeed. for the filter side kept it the same as the previous to differentiate the 2 reducers

PR #677, opened 2019-08-10
[FEAT] Add trip layer - 3

yanhann10:
- Add info icon on how to format data for trip layer in side panel 
- New slider ui with animated timestamp overlap on visual and start/end time displayed


yanhann10:
https://github.com/keplergl/kepler.gl/issues/397

Issue #390, opened 2019-02-28
Allow filtering by geo boundaries


heshan0131:
-   Allow an option to ‘filter points by polygon’ in point layer, select polygons inside a Geojson layer or a polygon drew by the user to filter by.

-   Use turf.js to filter points inside or outside of the selected polygon.

-   Allow an option to filter arcs/lines by whether it begins or end inside the polygon.

macrigiuseppe:
@heshan0131 I am looking at turf.js, should we use https://turfjs.org/docs/#pointsWithinPolygon to filter points within a polygon?
or is there another method we should use?

macrigiuseppe:
The new polygon filter will be treated as a regular filter and added to Filters in vis-state. Because the each new polygon filter will required to be mapped against an existing layer, we need to properly create a connection between a filter and layer.

- We are going to add a new property __layerId__ in Filter to track connection between polygon filter and layers.
- We need to modify __filterData__ in order to apply the new filters.
- The current dataset will be filtered based on the polygon filter layer properties (shape and and layer columns)

seanosterday:
@macrigiuseppe I can see that the filter by drawn polygon works but was the filter by a geojson layer ever implemented? I am not seeing this in the demo app.

maxwell-oroark:
+1 to @seanosterday. Unclear what I can presently accomplish with the draw tool in the demo app. I created a shape and added a data layer to the "Geo" filter but when I go to export "filtered data" the # of rows in the dataset are exactly the same, regardless of how restrictive my polygon is

macrigiuseppe:
> @macrigiuseppe I can see that the filter by drawn polygon works but was the filter by a geojson layer ever implemented? I am not seeing this in the demo app.

@seanosterday we currently don't support filter by geojson layer

macrigiuseppe:
> export "filtered dat

@maxwell-oroark i have tried the export functionality after filtering points using the draw tool and it works. See below:


<img width="1680" alt="Screen Shot 2021-03-19 at 4 58 20 PM" src="https://user-images.githubusercontent.com/2554552/111841510-727a6400-88d4-11eb-9bff-97f31972bbdf.png">

PR #678, opened 2019-08-12
[Feat] Apply polygon filter - Geo-Operations 2

macrigiuseppe:
Implements #390 

#### Storing filters
I created a new property __layerId__ to store a reference to the mapped layer for the current polygon.
The polygon (feature) is stored in the filter __value__ property.
Feature filter is also deleted upon deletion of paired feature

#### Filtering
In order to filter out data correctly I integrated with [turf.js](https://turfjs.org/). In order to provide better flexibility and also be able to filter geojson elements I used [booleanWithin](https://turfjs.org/docs/#booleanWithin)
```javascript
import booleanWithin from '@turf/boolean-within';
```

#### Examples
geo-ops-1

geo-ops-2

Screen Shot 2019-12-10 at 4 25 38 PM

Added Feature type to filter header

Screen Shot 2020-01-11 at 7 46 10 PM

Enabled filter feature visibility toggle

toggle-feature

Split Map
Screen Shot 2020-01-13 at 2 55 55 PM

geo-filter-split-map

All features
apply-geo-filters


src/utils/filter-utils.js
```diff
@@ -99,6 +102,26 @@ export const TIME_ANIMATION_SPEED = [
   }
 ];
 
+export function generatePolygonFilter(layer, feature) {
+  return generateFilter({
+    dataId: layer.config.dataId,
+    layerId: layer.id,
+    fixedDomain: false,
```
heshan0131: I think Polygon filter should be `fixedDomain: true`, like time filter. It's weird when people select filter to a subset of point and its color changes.

src/utils/filter-utils.js
```diff
@@ -123,7 +146,10 @@ export function getDefaultFilter(dataId) {
     // plot
     plotType: PLOT_TYPES.histogram,
     yAxis: null,
-    interval: null
+    interval: null,
+
+    // polygon:
+    layerId: null
```
heshan0131: I think it's time we create unique properties for each filter. instead of adding everything to default filter.  I imagine we can store not only layerId, but also layer.columns, and converted polygon geometry by turf in it.

src/utils/filter-utils.js
```diff
@@ -489,6 +513,20 @@ export function isInRange(val, domain) {
...
+/**
+ * Determines whether a point is within the provided polygon
+ *
+ * @param point as input search [lat, lng]
+ * @param polygon Points must be within these (Multi)Polygon(s)
+ * @return {boolean}
+ */
+export function isInPolygon(point, polygon) {
+  const convertedPoint = turfPoint(point);
+  const convertedPolygon = turfPolygon(polygon.geometry.coordinates);
```
heshan0131: Here you convert filter.value every time. should only need to do it once. you can store convertedPolygon in polygon filter?

heshan0131: wouldn't it be more performant to http://turfjs.org/docs/#booleanPointInPolygon?

macrigiuseppe: > wouldn't it be more performant to http://turfjs.org/docs/#booleanPointInPolygon?

I am not sure booleanPointInPolygon is more performant. When we convert a point to a turf point, the underline representation of convertedPoint is a geojson object hence i am not sure about performance.
On the other hand booleanWithin is more flexible and we can pass any kind of shape (not just points) through geojson therefore we could use for different layers

macrigiuseppe:
@heshan0131 
New feature:
- Able to edit/move existing geo features:
edit-feature

Moving turfPolygon creation to filter creation code definitely improved performance

macrigiuseppe:
Filtering while modifying polygon
on-the-fly-filter

macrigiuseppe:
@heshan0131 Added new filter panel with data info:

Screen Shot 2019-08-29 at 4 48 45 PM

heshan0131:
Created a polygon, go to select mode, select it, then click on rectangle icon, it tries to draw a rectangle but also filtering the data by the intersection

Screen Shot 2019-09-04 at 4 34 01 PM

heshan0131:
When I delete a polygon by right click select 'delete' in the floating panel, The filter on the left side is still there. Once linked filter and the actual polygon should associate to each other, delete either will delete the other

Screen Shot 2019-09-04 at 4 52 46 PM

src/reducers/vis-state-updaters.js
```diff
@@ -1328,3 +1344,167 @@ export function updateAllLayerDomainData(state, dataId, newFilter) {
...
+/**
+ * Toggle feature as layer filter
+ * @memberof visStateUpdaters
+ * @param state
+ * @param {Object} payload
+ * @param {string} payload.featureId
+ * @param {Object} payload.layer
+ * @return {Object} nextState
+ */
+export function togglePolygonFilterUpdater(state, payload) {
```
heshan0131: do we need to store `feature` that is also a polygon filter in 2 places? It's making the logic really complicated. and you have to add this `filterData` everywhere in all feature update / delete calls.

I think as soon as someone toggle a polygon into a filter, you can move the `feature` into `polygonFilter.value`. So all the update / remove call on polygon filter can just be using `setFilter` `removeFilter`

Obviously, on the `Editor` side, you will have to assign different `onUpdate`, `onDeleteFeature` based on whether the `selectedFeatureId` is a filter or not. But I think this simplifies the logic in `visStateUpdater` so we don't add this many specific updaters just for polygon feature

macrigiuseppe: in filter.value we store the turf.js representation of the feature and not original one so in case a user wants to apply the same feature as filter to multiple layers we still need the original one.
Another reason. If a user has a feature and a filter mapped against and decide to delete/toggle the filter, the feature should still be present on the map

src/reducers/vis-state-updaters.js
```diff
@@ -270,6 +236,86 @@ function updateTextLabelPropAndValue(idx, prop, value, textLabel) {
...
+      return [
+        ...acc,
+        index
+      ];
+    }
+    return acc;
+  }, []);
+}
+
+function removeFeatureFilters(state, filtersIdx) {
```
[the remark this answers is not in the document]
macrigiuseppe: done

macrigiuseppe:
@heshan0131 Rebased with Master, adapted the logic to multiple dataset filters and added new tests.
I will work with the designer on the filter panel for geoperations

test/fixtures/test-hex-id-data.js
```diff
@@ -530,7 +530,13 @@ mergedH3Layer.config = {
   isVisible: true,
   highlightColor: [252, 242, 26, 255],
   isConfigActive: false,
-  colorField: fields[1],
```
macrigiuseppe: because we are now loading layers before filters, fields are not updated with filterProps.

src/reducers/vis-state-updaters.js
```diff
@@ -1007,8 +1042,8 @@ export const receiveMapConfigUpdater = (
 
   // reset config if keepExistingConfig is falsy
   let mergedState = !keepExistingConfig ? resetMapConfigUpdater(state) : state;
-  mergedState = mergeFilters(mergedState, filters);
   mergedState = mergeLayers(mergedState, layers);
+  mergedState = mergeFilters(mergedState, filters);
```
macrigiuseppe: in order for polygon filters to be loaded correctly we need load layers before filters

src/components/editor/index.js
```diff
@@ -120,38 +131,59 @@ class Draw extends Component {
...
+    const availableLayers = layers.filter(editorLayerFilter);
 
     return (
-      <StyledWrapper editor={editor} className={classnames('editor', className)} style={style}>
+      <StyledWrapper
+        editor={editor}
+        className={classnames('editor', className)}
+        style={style}
+      >
         <Editor
```
heshan0131: Screen Shot 2019-12-24 at 5 46 19 AM

Can you create a task in upgrade react-map-gl-drawer so this doesn't showup?

src/components/editor/index.js
```diff
@@ -120,38 +131,59 @@ class Draw extends Component {
...
-      features,
       layers,
+      filters,
+      onUpdate,
       style
     } = this.props;
+
     const {selectedFeature = {}} = editor;
+    const {lastPosition, showActions} = this.state;
+    const selectedFeatureId = (selectedFeature || {}).id;
```
heshan0131: `{selectedFeature = {}} = editor` is doing the same thing. No need to do it again here.

src/components/editor/index.js
```diff
@@ -120,38 +131,59 @@ class Draw extends Component {
...
+    const currentFilter = filters.find(f => f.value.id === selectedFeatureId);
```
heshan0131: is it possible where `f.value` is not defined?

test/fixtures/test-hex-id-data.js
heshan0131: that's going to be a problem. we need field.mappedValue to calculate layer domain

heshan0131:
nitpick
Screen Shot 2019-12-24 at 6 12 59 AM
when toolbar is horizontal, the border-right is 1. when it is vertical, the border-right should be 0

src/components/common/action-panel.js
```diff
@@ -102,35 +103,49 @@ export const ActionPanelItem = React.memo(({
...
+          type="checkbox"
+          checked={Boolean(isActive)}
+          id={`switch-${label}`}
+          secondary
+          label={label}
+        />
+      ) : (
+        <span className="label">{label}</span>
+      )}
+      {children && children.length ? ( <div className="label-icon">
```
heshan0131: why not wrap them in 1 div
```js
     {children && children.length ?
      <div>
        <div className="label-icon">
          <ArrowRight height="16px" />
        </div>
        <div className="nested-group">
        {React.Children.map(children, renderChildren)}
        </div>
        </div>
      : null}
```

src/components/common/action-panel.js
```diff
@@ -102,35 +103,49 @@ export const ActionPanelItem = React.memo(({
...
+  style}) => {
+
+  const onClickCallback = useCallback(event => {
+    event.preventDefault();
+    event.stopPropagation();
+    onClick()
+  }, [onClick]);
+
+  return (
+    <StyledItem
```
heshan0131: Also nitpick:
the `:last` item should have `border-bottom: 0`

src/reducers/vis-state-updaters.js
heshan0131: see above comment

src/components/common/field-selector.js
```diff
@@ -19,26 +19,40 @@
...
       {showToken ? (
-        <div style={{display: 'inline-block', margin: '0 4px 0 0'}}>
+        <StyledToken>
           <FieldToken type={value.type} />
-        </div>
+        </StyledToken>
       ) : null}
-      <span className={classList.listItemAnchor}>{displayOption(value)}</span>
-    </div>
+      <span className={classnames(classList.listItemAnchor, value.type, displayOption(value))}>
```
heshan0131: is it neccessary to add `displayOption(value)` to `className`?  what if it is a very long string

src/components/common/item-selector/item-selector.js
```diff
@@ -31,25 +31,7 @@ import Typeahead from './typeahead';
...
-    return item;
-  }
-
-  if (typeof item === 'undefined' || item === null) {
-    return [];
-  }
-
-  return [item];
-}
+import {toArray} from 'utils/utils';
```
heshan0131: nice

src/components/common/styled-components.js
```diff
@@ -81,6 +83,17 @@ export const PanelLabel = styled.label.attrs({
   text-transform: capitalize;
 `;
 
+export const PanelValue = styled.label.attrs({
```
heshan0131: is this used anywhere?

src/components/common/styled-components.js
```diff
@@ -540,3 +553,33 @@ export const MapControlButton = styled.div.attrs({
     color: ${props => props.theme.secondaryBtnActColor};
   }
 `;
+
+export const StyledFilterPanel = styled.div`
```
heshan0131: `StyledFilterPanel` is also defined in `filters/components` and `side-panel/filter-panel/filter-panel`.  Can you delete duplicates and keep them all in one place? I don't think this should be a generic `styled-components` I would keep it under `side-panel`

src/components/common/styled-components.js
```diff
@@ -540,3 +553,33 @@ export const MapControlButton = styled.div.attrs({
...
+  margin-bottom: 12px;
+  border-radius: 1px;
+
+  .filter-panel__filter {
+    margin-top: 24px;
+  }
+`;
+
+export const StyledFilterHeader = styled(StyledPanelHeader)`
```
heshan0131: this is only used in `filter-panel-header`, if it's not used in other places, I would keep it there

src/components/editor/feature-action-panel.js
```diff
@@ -58,9 +60,13 @@ export class FeatureActionPanel extends PureComponent {
...
       layers,
+      currentFilter,
+      onToggleLayer,
       onDeleteFeature
     } = this.props;
 
+    const {layerId = []} = (currentFilter || {});
+
     return (
       <StyledActionsLayer className={className} position={position}>
```
heshan0131: instead of passing this to styled-component. this should just be in component `style={{top: x, left: y}}` handled by React. It is not recommended to pass props that constantly change to styled-component, because it needs to inject to css whenever prop changes.

src/components/editor/index.js
```diff
@@ -44,17 +44,28 @@ const StyledWrapper = styled.div`
   cursor: ${props => props.editor.mode === EDITOR_MODES.EDIT ? 'pointer' : 'crosshair'};
 `;
 
+const AVAILABLE_LAYERS = [
```
heshan0131: Shouldn't this be in `default-settings`?

src/components/filters/filter-panels/range-filter-panel.js
```diff
@@ -0,0 +1,112 @@
...
+        setFilter(idx, 'dataId', value),
+      [idx, setFilter]
+    );
+
+    const onSetFilter = useCallback(value =>
+        setFilter(idx, 'value', value),
+      [idx, setFilter]);
+
+    return (
+      <>
```
heshan0131: should this be a <div>?

src/utils/filter-utils.js
```diff
@@ -369,6 +435,74 @@ export function filterData(dataset, filters) {
...
+    const val = field ? data[field.tableFieldIndex - 1] : null;
+    return filter.value.includes(val)
+  },
+  [FILTER_TYPES.select]: (data, filter, index, field) => {
+    const val = field ? data[field.tableFieldIndex - 1] : null;
+    return val === filter.value
+  },
+  // layers contain only layers for the current dataset
+  [FILTER_TYPES.polygon]: (data, filter, index, field, layers) => {
+    if (!(layers || layers.length === 0)) {
```
heshan0131: This function is called on every data point, `currentLayers` should be passed into this function. Instead of creating it on every pass.

src/utils/filter-utils.js
```diff
@@ -369,6 +435,74 @@ export function filterData(dataset, filters) {
...
+    }
+
+    // determine which layers to apply the filter on
+    const currentLayers = filter.layerId.map(id =>
+        layers.find(l => l.id === id)
+      // we may get null value because filter.layerId may contain layers from other datasets
+    ).filter(l => Boolean(l));
+
+    return currentLayers.every(layer => {
+      const points = getLayerPointsToFilter(layer, data);
```
heshan0131: Each layer has a method `getPositionAccessor` which will return a position accessor, you can get point position. no need for the `getLayerPointsToFilter` function
```
const getPosition = layer.getPositionAccessor();
const points = getPosition(data);
```

src/utils/filter-utils.js
```diff
@@ -309,7 +375,7 @@ export function getFieldDomain(data, field) {
  * @param {Object} dataset to perform the filter on
  * @param {Object[]} filters list of filters to use against dataset
  */
-export function filterData(dataset, filters) {
+export function filterData(dataset, filters, layers) {
```
heshan0131: in this function, we can create a `filterFunctors(field, dataset.id, filter, layers)` instead of a `filtersToFields` with all the pre compyting before passing data point to it.  filterFunction is a function that takes `(data, index) => true || false`, it shouldn't does any computation rather than get the value from data and compare it to filter.value. which means by passing `field, dataset.id, filter, layers` to `getFilterFunction` we on longer need to pass it to `isDataMatchFilter`

implementation as below:

```js
  const appliedFilters = filters.filter(d => shouldApplyFilter(d, dataset.id));

  // Map filter against current dataset field
  const filterFunctors = appliedFilters.reduce((acc, filter) => {
    const fieldIndex = getDatasetFieldIndexForFilter(dataset, filter);
    const field = fieldIndex !== -1 ? fields[fieldIndex] : null;

    return {
      ...acc,
      [filter.id]: getFilterFunction(field, dataset.id, filter, layers)
    }
  }, {});


function getFilterFunction(field, dataId, filter, layers) {
  // field could be null
  let valueAccessor = data => field ? data[field.tableFieldIndex - 1] : null;

  switch (filter.type) {
    case FILTER_TYPES.range:
      return data => {
        return isInRange(valueAccessor(data), filter.value)
      };
    case FILTER_TYPES.multiSelect:
      return data => filter.value.includes(valueAccessor(data));
    case FILTER_TYPES.select:
      return data => valueAccessor(data) === filter.value;
    case FILTER_TYPES.timeRange:
      if (field && field.filterProp && Array.isArray(field.filterProp.mappedValue)) {
        valueAccessor = (data, index) => field.filterProp.mappedValue[index];
      } else {
        valueAccessor = data => timeToUnixMilli(valueAccessor(data), field.format);
      }
      return (data, index) => isInRange(valueAccessor(data, index), filter.value);
    case FILTER_TYPES.polygon:
      if (!(layers || layers.length === 0)) {
        return data => true;
      }

      const layerFilterFunctions = filter.layerId
        .map(id => layers.find(l => l.id === id))
        .filter(l => l && l.config.dataId === dataId)
        .map(layer => getPolygonFilterFunctor(layer, filter));

      return data => layerFilterFunctions.every(filterFunc => filterFunc(data));
    default:
      return data => true;
  }
};
```

src/utils/filter-utils.js
```diff
@@ -344,15 +410,15 @@ export function filterData(dataset, filters) {
       // generate 2 sets of
       // filter data used to calculate layer Domain
       const matchForDomain = dynamicDomainFilters.every(filter => {
-        return isDataMatchFilter(d, filter, i, filtersToFields[filter.id]);
+        return isDataMatchFilter(d, filter, i, filtersToFields[filter.id], layers);
```
heshan0131: no longer needed, simply call 
```js
 const matchForDomain = dynamicDomainFilters
        .every(filter => filterFunctors[filter.id](d, i));
```

src/utils/filter-utils.js
```diff
@@ -344,15 +410,15 @@ export function filterData(dataset, filters) {
...
       });
 
       if (matchForDomain) {
         accu.filteredIndexForDomain.push(i);
 
         // filter data for render
         const matchForRender = fixedDomainFilters.every(filter =>
-          isDataMatchFilter(d, filter, i, filtersToFields[filter.id])
+          isDataMatchFilter(d, filter, i, filtersToFields[filter.id], layers)
```
heshan0131: no longer needed, simply call 
```js
 const matchForDomain = dynamicDomainFilters
        .every(filter => filterFunctors[filter.id](d, i));
```

src/utils/filter-utils.js
```diff
@@ -369,6 +435,74 @@ export function filterData(dataset, filters) {
   return {data: filtered, filteredIndex, filteredIndexForDomain};
 }
 
+/**
+ * This returns the points to evaluate in polygon filter
+ * @param layer
+ * @param data
+ * @return {*[][]}
+ */
+export const getLayerPointsToFilter = (layer, data) => {
```
heshan0131: each layer has a `.getPositionAccessor` method, here you can precalculate `getPosition` and pass data to it

```js
export const getPolygonFilterFunctor = (layer, filter) => {
  const getPosition = layer.getPositionAccessor();

  switch (layer.type) {
    case LAYER_TYPES.arc:
    case LAYER_TYPES.line:
      return data => {
        const pos = getPosition({data});
        return pos.every(Number.isFinite) && [
          [pos[0], pos[1]],
          [pos[3], pos[4]]
        ].every(point => isInPolygon(point, filter.value));
      }
    default:
      return data => {
        const pos = getPosition({data});
        return pos.every(Number.isFinite) && isInPolygon(pos, filter.value);
      }
  }
};
```

src/utils/filter-utils.js
```diff
@@ -379,32 +513,9 @@ export function filterData(dataset, filters) {
...
-    case FILTER_TYPES.multiSelect:
-      return filter.value.includes(val);
-
-    case FILTER_TYPES.select:
-      return filter.value === val;
-
-    default:
-      return true;
-  }
+export function isDataMatchFilter(data, filter, i, field, layers = null) {
```
heshan0131: after mapping filter to `filterFunction`, we dont' need this function anumore

src/components/editor/index.js
macrigiuseppe: Done: https://github.com/keplergl/kepler.gl/issues/877

src/components/editor/index.js
macrigiuseppe: that's true

macrigiuseppe:
> nitpick
> Screen Shot 2019-12-24 at 6 12 59 AM
> when toolbar is horizontal, the border-right is 1. when it is vertical, the border-right should be 0

Screen Shot 2019-12-24 at 5 47 09 PM

src/components/common/field-selector.js
macrigiuseppe: I can actually remove those, they may have been leftovers from previous versions

src/components/common/styled-components.js
macrigiuseppe: leftover and removed

src/components/common/styled-components.js
macrigiuseppe: i removed the code from here. I think i have started creating some generic components and stopped at some point

src/components/editor/index.js
macrigiuseppe: I will move it to default-settings but i will rename it into
```
EDITOR_AVAILABLE_LAYERS
```

src/components/map-container.js
```diff
@@ -468,22 +479,20 @@ export default function MapContainerFactory(MapPopover, MapControl) {
...
-                and fewer updates when we switch from edit to read mode
-              */}
             <Draw
               datasets={datasets}
-              editor={uiState.editor}
-              features={editor.features}
+              editor={editor}
+              filters={this.polygonFilters(this.props)}
               isEnabled={isEdit}
               layers={layers}
```
heshan0131: Should we show layer that is currently hidden? It's probably better to only show visible layer. You can achieve that by passing `layersToRender`, it is a `{[layer.id]: layer}` Object

src/processors/data-processor.js
```diff
@@ -203,6 +203,7 @@ function cleanUpFalsyCsvValue(rows) {
  * Process uploaded csv file to parse value by field type
  *
  * @param {Array<Array>} rows
+ * @param geoFieldIdx geo field index
```
heshan0131: it should be {Number}

src/reducers/vis-state-updaters.js
```diff
@@ -111,12 +120,19 @@ disableStackCapturing();
 const visStateUpdaters = null;
 /* eslint-enable no-unused-vars */
 
-export const defaultAnimationConfig = {
+export const defaultAnimationConfig ={
```
heshan0131: missing space

src/reducers/vis-state-updaters.js
```diff
@@ -183,10 +199,8 @@ export const INITIAL_VIS_STATE = {
   // time in unix timestamp (milliseconds) (the number of seconds since the Unix Epoch)
   animationConfig: defaultAnimationConfig,
 
-  editor: {
-    // GEO FEATURES (Shapes)
-    features: []
-  }
+  editor: DEFAULT_EDITOR
+
```
heshan0131: empty line

src/reducers/vis-state-updaters.js
```diff
@@ -535,8 +553,8 @@ export function setFilterUpdater(state, action) {
     return state;
   }
 
-  // ENsuring backward compatibility
-  const datasetIds = Array.isArray(dataId) ? dataId : [dataId];
+  // Ensuring backward compatibility
+  let datasetIds = Array.isArray(dataId) ? dataId : [dataId];
```
heshan0131: use `toArray`?

src/reducers/vis-state-updaters.js
```diff
@@ -799,10 +830,18 @@ export const removeFilterUpdater = (state, action) => {
...
     ...state,
-    datasets: applyFiltersToDatasets(dataId, state.datasets, newFilters),
-    filters: newFilters
+    datasets: applyFiltersToDatasets(dataId, state.datasets, newFilters, state.layers),
+    filters: newFilters,
+    editor: {
+      ...state.editor,
+      features: filter.type !== FILTER_TYPES.polygon ?
+        state.editor.features : state.editor.features.filter(f => f.id !== filter.value.id),
+      selectedFeature: null
```
heshan0131: if `filter.type !== FILTER_TYPES.polygon ` should `selectedFeature` be set to null?
Looks like you only need to update editor if ``filter.type === FILTER_TYPES.polygon` so why not
```js

  const newState = {
    ...state,
    datasets: applyFiltersToDatasets(dataId, state.datasets, newFilters, state.layers),
    filters: newFilters,
    editor:  filter.type === FILTER_TYPES.polygon ? {
      ...state.editor,
      features: state.editor.features.filter(f => f.id !== filter.value.id),
      selectedFeature: null
    } : state.editor
  };
```

src/reducers/vis-state-updaters.js
```diff
@@ -1524,3 +1530,215 @@ export function updateAnimationDomain(state) {
...
+  const filtersIdx = getFilterIdxByFeatureId(newState, selectedFeature.id);
+
+  const filter = state.filters[filtersIdx];
+
+  const newFilter = updatePolygonFilter(filter, feature);
+
+  const newFilters = Object.assign([...state.filters], {[filtersIdx]: newFilter});
+
+  // no filters found
+  if (filtersIdx === -1) {
```
heshan0131: shouldn't this be moved to right after `const filtersIdx = getFilterIdxByFeatureId(newState, selectedFeature.id);`

src/reducers/vis-state-updaters.js
```diff
@@ -1524,3 +1530,215 @@ export function updateAnimationDomain(state) {
...
+export function setFeaturesUpdater(state, {features = []}) {
+
+  const lastFeature = features.length && features[features.length - 1];
+
+  let newState = {
+    ...state,
+    editor: {
+      ...state.editor,
+      features,
+      mode: lastFeature && lastFeature.properties.isClosed ?
```
heshan0131: what is `isClosed`? and why when `isClosed` is true editor mode should be EDIT?

src/reducers/vis-state-updaters.js
```diff
@@ -1524,3 +1530,215 @@ export function updateAnimationDomain(state) {
...
+  }
+
+  // TODO: check if the feature has changed
+  const feature = features.find(f => f.id === selectedFeature.id);
+
+  if (!feature) {
+    return newState;
+  }
+
+  const filtersIdx = getFilterIdxByFeatureId(newState, selectedFeature.id);
```
heshan0131: shouldn't this be `filterIdx`?

src/utils/filter-utils.js
```diff
@@ -840,3 +969,51 @@ export function getDatasetFieldIndexForFilter(dataset, filter) {
...
+}
+
+/**
+ * Updates polygon filter
+ * @param filter
+ * @param feature
+ * @return {object} Filter
+ */
+export function updatePolygonFilter(filter, feature) {
+  const polygon = turfPolygon(feature.geometry.coordinates);
```
heshan0131: a turfPolygon is a geojson Polygon feature, same as the `feature` object react-map-gl-draw returns. Here you dn't need to call  `turfPolygon` and merge them. this works:
```js
export function updatePolygonFilter(filter, feature) {
  // const polygon = turfPolygon(feature.geometry.coordinates);
  return {
    ...filter,
    // we merge both turf and feature properties into one
    value: {
      id: feature.id,
      ...feature
    }
  }
}
```

src/reducers/vis-state-merger.js
```diff
@@ -111,13 +110,24 @@ export function mergeFilters(state, filtersToMerge) {
   // flatten all filter dataIds
   const datasetsToFilter = uniq(flattenDeep(merged.map(f => f.dataId)));
 
-  const filtered = applyFiltersToDatasets(datasetsToFilter, updatedDatasets, updatedFilters);
+  const filtered = applyFiltersToDatasets(datasetsToFilter, updatedDatasets, updatedFilters, state.layers);
```
heshan0131: Can you add a test case to `vis-state-merger-test.js` with a map saved with a polygon filter, some other filter and a layer with `colorField`, then loaded it back to test this?

src/utils/filter-utils.js
macrigiuseppe: this looks way better, thank you

src/utils/filter-utils.js
macrigiuseppe:
```
const matchForRender = fixedDomainFilters
        .every(filter => filterFunctors[filter.id](d, i));
```

src/utils/filter-utils.js
macrigiuseppe: had to change few things in the above code because it was causing the following error:
```
not ok 1 RangeError: Maximum call stack size exceeded
  ---
    operator: error
    expected: |-
      undefined
    actual: |-
      [RangeError: Maximum call stack size exceeded]
    at: Test.die (/Users/gmacri/Documents/projects/kepler.gl/node_modules/tape-catch/index.js:56:10)
    stack: |-
      RangeError: Maximum call stack size exceeded
          at valueAccessor (/Users/gmacri/Documents/projects/kepler.gl/src/utils/filter-utils.js:418:25)
          at valueAccessor (/Users/gmacri/Documents/projects/kepler.gl/src/utils/filter-utils.js:418:49)
```
@ line: ```valueAccessor = data => timeToUnixMilli(valueAccessor(data), field.format);```

The overall result is way better

src/components/map-container.js
macrigiuseppe: layersToRender container {[layer.id]: true|false}.
It looks like the current pattern is to pass both __layers__ and __layersToRender__ so I need to pass both layers and layers to render

src/reducers/vis-state-updaters.js
macrigiuseppe: isClosed is a property that comes from react-map-gl-draw. It means the feature we are currently drawing is completed, e.g. while drawing a polygon we finally draw all lines.

The abose logic allows the app to detect when the current feature is being completed and switch from POLYGON/RECTANGLE mode to EDIT mode so users can immediately select the latest created feature.

select-feature

src/actions/vis-state-actions.js
```diff
@@ -508,6 +508,76 @@ export function setFeatures(features) {
...
+}
+
+/**
+ * Set the current feature to be edited/deleted
+ * @memberof visStateActions
+ * @param {Object} feature
+ * @param {string} feature.selectedFeatureId
+ * @return {{selectedFeatureId: *, type: ActionTypes.SET_SELECTED_FEATURE}}
+ */
+export function setSelectedFeature({selectedFeatureId}) {
```
heshan0131: Seems like all the places where `onSelect` are called in `editor.js`, you are only passing `{selectedFeatureId: null | id}`,  why not unwrap the argument to be `selectedFeatureId`?

src/components/common/vertical-toolbar.js
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components';
+import Toolbar from './toolbar';
+
+const VerticalToolbar = styled(Toolbar)`
+  flex-direction: column;
+  padding: 0 24px;
+  
+  .toolbar-item {
+    padding: 16px;
```
heshan0131: vertical toolbar `toolbar-item` padding should be `16px 0`. Because toolbar already has side padding

src/components/map/map-control.js
```diff
@@ -240,16 +242,22 @@ const Toggle3dButton = React.memo(({dragRotate, onTogglePerspective}) => (
...
 Toggle3dButton.displayName = 'Toggle3dButton';
 
-const StyledToolBar = styled(Toolbar)`
+const StyledToolbar = styled(VerticalToolbar)`
   position: absolute;
   right: 32px;
 `;
 
-const MapDrawPanel = React.memo(({editor, isActive, onToggleMenuPanel, onSetEditorMode}) => {
+const MapDrawPanel = React.memo(({
```
heshan0131: In split map, is it possible to only show `MapDrawPanel` in the split map clicked?  Currently, it is shown in both, which is a little weird.
e,g, in `mapControls.mapDraw` we can save the `splitMap` index [true, false], instead of `true | false`. However, this means in `ui-state` reducer has to listen to `TOGGLE_SPLIT_MAP` and update mapDraw

Screen Shot 2020-01-07 at 9 18 34 PM

src/reducers/vis-state-updaters.js
```diff
@@ -311,6 +325,10 @@ function updateTextLabelPropAndValue(idx, prop, value, textLabel) {
   return newTextLabel;
 }
 
+function getFilterIdxByFeatureId(state, featureId) {
+  return state.filters.findIndex(f => f.type === FILTER_TYPES.polygon && f.value.id === featureId);
```
heshan0131: what if `f.value` is `null`?

src/reducers/vis-state-merger.js
```diff
@@ -111,13 +110,24 @@ export function mergeFilters(state, filtersToMerge) {
...
+  const features = updatedFilters.reduce((acc, f) => {
+    return f.type === FILTER_TYPES.polygon ? [...acc, f.value] : acc;
+  }, []);
 
   return {
     ...state,
     filters: updatedFilters,
     datasets: filtered,
-    filterToBeMerged: unmerged
+    filterToBeMerged: unmerged,
```
heshan0131: now that we merge filter twice, `filterToBeMerged` should be `[...state.filterToBeMerged, ...unmerged]`

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+
+  const newFilters = Object.assign([...state.filters], {[filterIdx]: newFilter});
+
+  newState = {
+    ...state,
+    datasets: applyFiltersToDatasets(newFilter.dataId, newState.datasets, newFilters, newState.layers),
+    filters: newFilters
+  };
+
+  return updateAllLayerDomainData(newState, Object.keys(newState.datasets));
```
heshan0131: dataId here should be `newFilter.dataId`, and you should pass `newFilter` to this function as well.
Because we only need to update the layer domains / layerdata associated with newly filtered dataset. and skipped the layers haven't been affected by it's parent datasets.

Also if the filter has fixedDomain, we don't need to update layer.domain, only need to update layerData

```
return updateAllLayerDomainData(newState, newFilter.dataId, newFilter);
```

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+  const filterIdx = getFilterIdxByFeatureId(newState, selectedFeature.id);
+
+  // no filters found
+  if (filterIdx === -1) {
+    return newState;
+  }
+
+  const filter = state.filters[filterIdx];
+
+  const newFilter = updatePolygonFilter(filter, feature);
```
heshan0131: updatePolygonFilter in a nutshell updates filter.value, and all the code following it is to apply the filter.
You should be able to call  `setFilterUpdater` here with updated value, instead of all the lines below.
```
const featureInValue = {...feature, id: feature.id};
return setFilterUpdater(newState, {idx: filterIdx, prop: 'value', value: featureInValue})
```

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+ * @param {Object} payload
+ * @param {string} payload.featureId
+ * @param {Object} payload.layer
+ * @return {Object} nextState
+ */
+export function togglePolygonFilterUpdater(state, payload) {
+  const {layer, featureId} = payload;
+  const {dataId} = layer.config;
+  const filterIdx = getFilterIdxByFeatureId(state, featureId);
+  let datasetIds = toArray(dataId);
```
heshan0131: you don't need `toArray` here because layer.config.dataId is always a single id, layer is only associated with 1 dataset.

src/utils/filter-utils.js
```diff
@@ -18,13 +18,18 @@
...
 // THE SOFTWARE.
 
-import moment from 'moment';
 import {ascending, extent, histogram as d3Histogram, ticks} from 'd3-array';
 import keyMirror from 'keymirror';
+import get from 'lodash.get';
+import booleanWithin from '@turf/boolean-within';
+import {point as turfPoint} from '@turf/helpers';
 import {ALL_FIELD_TYPES} from 'constants/default-settings';
 import {maybeToDate, notNullorUndefined, unique} from './data-utils';
```
heshan0131: add `timeToUnixMilli` here

src/utils/filter-utils.js
```diff
@@ -304,12 +368,74 @@ export function getFieldDomain(data, field) {
...
+    case LAYER_TYPES.line:
+      return data => {
+        const pos = getPosition({data});
+        return pos.every(Number.isFinite) && [
+          [pos[0], pos[1]],
+          [pos[3], pos[4]]
+        ].every(point => isInPolygon(point, filter.value));
+      };
+    default:
+      return data => {
```
heshan0131: not all getPosition from layer returns point, i think for other layer type we should simply return `data => true`

src/utils/filter-utils.js
```diff
@@ -304,12 +368,74 @@ export function getFieldDomain(data, field) {
...
+    case FILTER_TYPES.range:
+      return data => isInRange(valueAccessor(data), filter.value);
+    case FILTER_TYPES.multiSelect:
+      return data => filter.value.includes(valueAccessor(data));
+    case FILTER_TYPES.select:
+      return data => valueAccessor(data) === filter.value;
+    case FILTER_TYPES.timeRange:
+      const mapppedValue = get(field, ['filterProp', 'mappedValue']);
+      const accessor = Array.isArray(mapppedValue) ?
+        (data, index) => field.filterProp.mappedValue[index]
```
heshan0131: You already have `mapppedValue` defined
`(data, index) => mappedValue[index] : data => timeToUnixMilli(valueAccessor(data), field.format);`

src/reducers/vis-state-updaters.js
heshan0131: Again on this point. 1. turf representation and the original feature is the same. 2. if a user deletes a polygon filter on map, the feature should not be present on the map. I don't think we should store them in 2 places. we would have to implement `getFilterIdxByFeatureId`, `currentFeature = editorFeatures.find(feat => feat.id === featureId);` just to get each one.

src/actions/vis-state-actions.js
```diff
@@ -508,6 +508,76 @@ export function setFeatures(features) {
...
 
+/**
+ * It will apply the provide feature as filter to the given layer.
+ * If the given feature is already applied as filter to the layer it  will remove it from  the filter list
+ * @memberof visStateActions
+ * @param {Object} featureId feature to use as filter base for layer
+ * @param {Object} layer
+ * @return {{feature: *, type: ActionTypes.TOGGLE_POLYGON_FILTER, layer: *}}
+ */
+export function togglePolygonFilter(layer, featureId) {
```
heshan0131: the naming of this action is a bit blurry. it sounds like this is to hide and show polygon filter. If layer is passed along can it be `setPolygonFilterLayer` ?

src/actions/vis-state-actions.js
macrigiuseppe: Initially I thought I was going to propagate the full feature

src/reducers/vis-state-updaters.js
macrigiuseppe: actually it turns out if we call setFilterUpdater we don't need to use updatePolygonFilter and we can simply do the following:
```
return setFilterUpdater(newState, {idx: filterIdx, prop: 'value', value: feature});
```

src/reducers/vis-state-updaters.js
macrigiuseppe: no longer needed because we are now using setFilterUpdater

src/reducers/vis-state-merger.js
macrigiuseppe: this change is breaking a lot of tests. need to dig a little bit

macrigiuseppe: this may not work because if we consider the following scenario:
- a filter is added to filterToBeMerged list
- a new dataset is added that matches such filter, therefore, we try to merge filters again
- when we merge filter, ```state.filterToBeMerged``` has one element and unmerged has 0, therefore, the new state ```filterToBeMerged``` property will have again one element because of ```[...state.filterToBeMerged, ...unmerged]```

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+});
+
+/**
+ * Update editor features
+ * @memberof visStateUpdaters
+ * @param {Object} state `visState`
+ * @param {[Object]} features to store
+ * @return {Object} nextState
+ */
+export function setFeaturesUpdater(state, {features = []}) {
```
heshan0131: regarding to the comments in this function
see [this commit](https://github.com/keplergl/kepler.gl/commit/7f927840ab976305eee3200d3aedb903ab60c0df#diff-3c42abf857bb61c5f6578c72771405d2R1556)

src/reducers/vis-state-updaters.js
heshan0131: I tested out in [this commit](https://github.com/keplergl/kepler.gl/commit/7f927840ab976305eee3200d3aedb903ab60c0df#diff-3c42abf857bb61c5f6578c72771405d2R1658)

src/schemas/vis-state-schema.js
```diff
@@ -648,7 +648,11 @@ export const filterPropsV1 = {
       name: null,
       type: null
     }
-  })
+  }),
+  fixedDomain: null,
```
heshan0131: fixedDomain is a property generated from `getDefaultFilter` we don't need to save it to the config

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+
+/**
+ * Delete existing feature from filters
+ * @memberof visStateUpdaters
+ * @param {Object} state `visState`
+ * @param {string} selectedFeatureId feature to delete
+ * @return {Object} nextState
+ */
+export function deleteFeatureUpdater(state, {featureId}) {
+  if (!featureId) {
```
heshan0131: pass feature to be deleted instead of `featureId`,
https://github.com/keplergl/kepler.gl/commit/7f927840ab976305eee3200d3aedb903ab60c0df#diff-3c42abf857bb61c5f6578c72771405d2R1615

src/reducers/vis-state-updaters.js
```diff
@@ -1514,3 +1526,215 @@ export function updateAnimationDomain(state) {
...
+}
+
+/**
+ * Set the current selected feature
+ * @memberof uiStateUpdaters
+ * @param {Object} state `uiState`
+ * @param {[Object]} features to store
+ * @return {Object} nextState
+ */
+export const setSelectedFeatureUpdater = (state, {selectedFeatureId}) => ({
```
heshan0131: instead of save the id in reducer, why not save the entire feature, that way it easier to literate over its property, also nicer to pass the entire feature around when deleting it
https://github.com/keplergl/kepler.gl/commit/7f927840ab976305eee3200d3aedb903ab60c0df#diff-3c42abf857bb61c5f6578c72771405d2R1600

src/components/editor/index.js
```diff
@@ -24,7 +24,8 @@ import styled from 'styled-components';
 import {Editor} from 'react-map-gl-draw';
```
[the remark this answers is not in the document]
heshan0131: also made some changes to the Event key
https://github.com/keplergl/kepler.gl/commit/7f927840ab976305eee3200d3aedb903ab60c0df#diff-37098adfde85b215e32b27850c960d0fR43

src/reducers/vis-state-merger.js
macrigiuseppe: @heshan0131 ^

src/reducers/vis-state-updaters.js
macrigiuseppe: the main reason is because react-map-gl-draw provide a feature.id when a selection is performed

src/components/map/map-control.js
macrigiuseppe: We may have a similar problem with the legend being displayed in both maps at the sam time
hmm do you think a lot of people will be using the editor mode in split map view?
We could disable the editor when in split mode

heshan0131: displaying legend and map control panel in 2 places is fine actually. It allows people to compare whats different in 2 split map. But allow drawying in split maps is useful. you can imagine people want to draw in one map, and see how the other map is filtered. 

For example. showing begin trip in one map, and dropoff in another. Drawing in begin trip to see how dropoff map changes...

src/reducers/vis-state-merger.js
heshan0131: in cases where filterA is regular filter, filterB is layer based filter.  both are waiting for the dataset to be loaded. when we run mergeFilter once, we save filterA to unmerged. and when we run it again, we clear out unmerged, and save filterB to unmerged. It's wrong, because we are losing filterA

src/components/map/map-control.js
macrigiuseppe: ok, that makes sense

src/reducers/vis-state-merger.js
macrigiuseppe: got it. Need to review the logic and understand why a lot of tests are failing

src/reducers/vis-state-updaters.js
macrigiuseppe: no longer valid

macrigiuseppe:
@heshan0131 
I added polygon type in filter panel header. I added the prefix __Geo__ because the previous one (Polygon) would have created a non optimal user experience since for non-rectangle features we would have displayed ```Polygon - Polygon```. See below

Screen Shot 2020-01-11 at 7 46 10 PM

Added toggle visibility for filter feature. Initially, I was going to rename onToggleAnimation to something more generic to be reused in different filters but I think it's better to have different actions in case we want to extend filter panel in the future so I decided to create a new __toggleFilterFeature__ action. See below

apply-geo-filters

test/node/schemas/vis-state-schema-test.js
```diff
@@ -102,7 +102,8 @@ test('#visStateSchema -> v1 -> save load filters', t => {
     value: [1474606800000, 1474617600000],
     enlarged: true,
     plotType: 'histogram',
-    yAxis: null
+    yAxis: null,
+    fixedDomain: true
```
heshan0131: don't need to save `fixedDomain` prop

src/utils/filter-utils.js
```diff
@@ -134,33 +143,66 @@ export function getDefaultFilter(dataId) {
...
+
+  if (!layer) {
+    return failed;
+  }
+
+  return {
+    filter: {
+      ...filter,
+      freeze: true,
+      fieldIdx: [0]
```
heshan0131: why `fieldIdx` needs to be [0]? if there is no field selected, can't it be `[]`?

src/schemas/vis-state-schema.js
heshan0131: ping

src/components/filters/filter-panels/multi-select-filter-panel.js
```diff
@@ -0,0 +1,112 @@
...
+        setFilter(idx, 'dataId', value),
+      [idx, setFilter]
+    );
+
+    const onSetFilter = useCallback(value =>
+        setFilter(idx, 'value', value),
+      [idx, setFilter]);
+
+    return (
+      <>
```
heshan0131: <div>?

src/components/filters/filter-panels/new-filter-panel.js
```diff
@@ -0,0 +1,94 @@
...
+      [idx, setFilter]
+    );
+
+    const onSourceDataSelector = useCallback(value =>
+        setFilter(idx, 'dataId', value),
+      [idx, setFilter]
+    );
+
+    return (
+      <>
```
heshan0131: <div className="new-filter-panel">

heshan0131: `<div className="new-filter-panel">`

src/components/filters/filter-panels/polygon-filter-panel.js
```diff
@@ -0,0 +1,102 @@
...
+
+    const onSetLayers = useCallback(value =>
+        setFilter(idx, 'layerId', value),
+      [setFilter]);
+
+    const isVisible = get(filter, ['value', 'properties', 'isVisible'], true);
+    const featureType = get(filter, ['value', 'properties', 'renderType'], true);
+
+    return (
+      <>
```
heshan0131: `<div className="polygon-filter-panel">

src/components/filters/filter-panels/range-filter-panel.js
```diff
@@ -0,0 +1,112 @@
...
+        setFilter(idx, 'dataId', value),
+      [idx, setFilter]
+    );
+
+    const onSetFilter = useCallback(value =>
+        setFilter(idx, 'value', value),
+      [idx, setFilter]);
+
+    return (
+      <div>
```
heshan0131: `<div className="range-filter-panel">`

src/components/filters/filter-panels/single-select-filter-panel.js
```diff
@@ -0,0 +1,113 @@
...
+        setFilter(idx, 'dataId', value),
+      [idx, setFilter]
+    );
+
+    const onSetFilter = useCallback(value =>
+        setFilter(idx, 'value', value),
+      [idx, setFilter]);
+
+    return (
+      <>
```
heshan0131: same here

src/components/filters/filter-panels/range-filter-panel.js
```diff
@@ -0,0 +1,112 @@
...
+        <FilterPanelHeader
```
heshan0131:
```js
const onFieldSelector = useCallback(field =>
        setFilter(idx, 'name', field.name),
      [idx, setFilter]
    );

    const onSourceDataSelector = useCallback(value =>
        setFilter(idx, 'dataId', value),
      [idx, setFilter]
    );

    const onSetFilter = useCallback(value =>
        setFilter(idx, 'value', value),
      [idx, setFilter]);

<PanelHeader>
   <FieldSelector/>
</PanelHeader>
<StyledFilterContent>
    <SourceDataSelecto/>
</StyledFilterContent>
```

Has been repeated in every `xx-filter-panel` component, except polygon filter. Repeating them 5 times, doesn't make sense. Why not make a component `FilterPanelWithFieldSelect` and use it to wrap other filters?

src/components/filters/filter-panels/polygon-filter-panel.js
macrigiuseppe: i assigned ```<div className="polygon-filter-panel">```

src/components/filters/filter-panels/range-filter-panel.js
macrigiuseppe: I think I decided to go down this way because both polygon and time range filters have panel actions and by simply using children couldn't be recreated but I think we can create the common component and
- panel actions: will be passed as an array of object: `
``{onClick: enlargeFilter, tooltip:: 'Time Playback, iconComponent: Clock, onSelect: 
onFieldSelector}```
- the actual field component, e.g.TimeRangeFilter, will be passed as one of the children

macrigiuseppe:
Fixed map control padding:
Screen Shot 2020-01-17 at 2 13 11 PM

Fixed map export
Screen Shot 2020-01-17 at 2 16 04 PM

src/actions/vis-state-actions.js
```diff
@@ -378,6 +378,18 @@ export function enlargeFilter(idx) {
   };
 }
 
+/**
+ * SHow/hide filter feature on map
+ * @param {Number} idx - index of filter feature to show/hide
+ * @return {{type: ActionTypes.TOGGLE_FEATURE_LAYER, idx: idx}}
```
heshan0131: return `TOGGLE_FILTER_FEATURE`

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,76 @@ export function setFeatures(features) {
...
+    layer,
+    feature
+  };
+}
+
+/**
+ * Set the current feature to be edited/deleted
+ * @memberof visStateActions
+ * @param {Object} feature
+ * @param {string} feature.selectedFeatureId
```
heshan0131: remove `@param {string} feature.selectedFeatureId`

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,76 @@ export function setFeatures(features) {
   }
 }
 
+/**
+ * It will apply the provide feature as filter to the given layer.
+ * If the given feature is already applied as filter to the layer it  will remove it from  the filter list
+ * @memberof visStateActions
+ * @param {Object} featureId feature to use as filter base for layer
```
heshan0131: JSDoc param should be updated

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,76 @@ export function setFeatures(features) {
...
+  return {
+    type: ActionTypes.SET_SELECTED_FEATURE,
+    feature
+  }
+}
+
+/**
+ * Delete the given feature
+ * @memberof visStateActions
+ * @param {string} featureId if of the feature
```
heshan0131: same here

src/components/map/map-control.js
```diff
@@ -305,18 +319,21 @@ const MapControlFactory = () => {
...
+      onToggleEditorVisibility: PropTypes.func.isRequired,
       top: PropTypes.number.isRequired,
 
       // optional
       readOnly: PropTypes.bool,
       scale: PropTypes.number,
       mapLayers: PropTypes.object,
-      editor: PropTypes.object
+      editor: PropTypes.object,
+      indexMap: PropTypes.number
```
heshan0131: `indexMap` is defined, but `index` is passed in

src/components/map/map-control.js
```diff
@@ -305,18 +319,21 @@ const MapControlFactory = () => {
...
     };
 
     static defaultProps = {
       isSplit: false,
-      top: 0
+      top: 0,
+      indexMap: 0
```
heshan0131: `index` instead of `indexMap`

src/components/map/map-control.js
macrigiuseppe: actually I noticed we were already passing mapIndex hence we don't need either index now indexMap

src/components/map/map-control.js
macrigiuseppe: replace into mapIndex=0

macrigiuseppe:
@heshan0131 updated

src/actions/vis-state-actions.js
```diff
@@ -378,6 +378,18 @@ export function enlargeFilter(idx) {
   };
 }
 
+/**
+ * SHow/hide filter feature on map
```
heshan0131: * @memberof visStateActions

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
   }
 }
 
+/**
+ * It will apply the provide feature as filter to the given layer.
+ * If the given feature is already applied as filter to the layer it  will remove it from  the filter list
+ * @memberof visStateActions
+ * @param {Object} feature
```
heshan0131: layer first, feature second

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
   }
 }
 
+/**
+ * It will apply the provide feature as filter to the given layer.
+ * If the given feature is already applied as filter to the layer it  will remove it from  the filter list
```
heshan0131: If the given feature is already applied as filter to the layer, it will __remove the layer__ from the filter

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
...
+    layer,
+    feature
+  };
+}
+
+/**
+ * Set the current feature to be edited/deleted
+ * @memberof visStateActions
+ * @param {Object} feature
+ * @return {{selectedFeatureId: *, type: ActionTypes.SET_SELECTED_FEATURE}}
```
heshan0131: `@return {{feature: *, type: ActionTypes.SET_SELECTED_FEATURE}}`

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
...
+    type: ActionTypes.SET_SELECTED_FEATURE,
+    feature
+  }
+}
+
+/**
+ * Delete the given feature
+ * @memberof visStateActions
+ * @param {Object} feature
+ * @return {{type: ActionTypes.DELETE_FEATURE, featureId: *}}
```
heshan0131: `@return {{type: ActionTypes.DELETE_FEATURE, feature: *}}`

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
...
+ */
+export function deleteFeature(feature) {
+  return {
+    type: ActionTypes.DELETE_FEATURE,
+    feature
+  };
+}
+
+/** Set the map mode
+ * @memberof main
```
heshan0131: `@memberof visStateActions`
`@return {{type: ActionTypes. SET_EDITOR_MODE, mode: *}}`

src/actions/vis-state-actions.js
```diff
@@ -508,6 +520,75 @@ export function setFeatures(features) {
...
+ */
+export function setEditorMode(mode) {
+  return {
+    type: ActionTypes.SET_EDITOR_MODE,
+    mode
+  };
+}
+
+/**
+ * Toggle editor layer visibility
```
heshan0131: `@memberof visStateActions`

examples/demo-app/src/components/map-control/map-control.js
```diff
@@ -182,7 +182,7 @@ function SampleMapPanel(props) {
 
 const CustomMapControl = props => (
   <StyledMapControlOverlay>
-    {props.currentSample ? <SampleMapPanel {...props} /> : null}
+    {!props.isExport && props.currentSample ? <SampleMapPanel {...props} /> : null}
```
macrigiuseppe: we don't show info panel when export map

src/components/map-container.js
```diff
@@ -439,23 +462,25 @@ export default function MapContainerFactory(MapPopover, MapControl) {
       return (
         <StyledMapContainer style={MAP_STYLE.container}>
           <MapControl
+            index={index}
```
heshan0131: we already pass in`mapIndex={this.props.index}`, no need to pass in`index={index}`

src/components/map/map-control.js
heshan0131: delete `indexMap: PropTypes.number`

src/components/map/map-control.js
heshan0131: did you push? it'still showing `indexMap: 0`

src/components/map-container.js
macrigiuseppe: yeah, I already removed that one

src/components/map/map-control.js
```diff
@@ -351,7 +367,8 @@ const MapControlFactory = () => {
         onToggleMapControl,
         editor,
         scale,
-        readOnly
+        readOnly,
+        index
```
heshan0131: delete index

src/components/map/map-control.js
```diff
@@ -414,10 +431,11 @@ const MapControlFactory = () => {
           {mapDraw.show ? (
             <ActionPanel key={4}>
               <MapDrawPanel
-                isActive={mapDraw.active}
+                isActive={mapDraw.active && mapDraw.activeMapIndex === index}
```
heshan0131: `mapDraw.active && mapDraw.activeMapIndex === mapIndex`

src/components/map/map-control.js
macrigiuseppe: yeah done

src/components/map/map-control.js
macrigiuseppe: same here

cperbost:
> @heshan0131 Added new filter panel with data info:
> 
> Screen Shot 2019-08-29 at 4 48 45 PM

Hi @macrigiuseppe , I can't find how to display the "count" (of points inside the polygon) on the layer section. Is does not seem to be implemented in version 2.2.0. Did you remove it?

sulemanbadr:
@cperbost did you able to fidn this "count" feature?

PR #683, opened 2019-08-13
[Bug] Fixed issue with geojson fields

### LINKED ISSUE #679: [Bug] Table view crashes when geojson dataset is being display
https://kepler.gl/demo/nyc_census

Datatable view crashes

macrigiuseppe:
Fixes #679 

The current implementation doesn't have the ability to display a subset of the row properties so I had to create a way to omit some of the properties.

679-table-issue

### DIFF
--- a/src/components/common/datagrid.js
+++ b/src/components/common/datagrid.js
@@ -23,6 +23,7 @@ import PropTypes from 'prop-types';
 import {MultiGrid} from 'react-virtualized';
 import styled, {withTheme}from 'styled-components';
 import {createSelector} from 'reselect';
+import classnames from 'classnames';
 import FieldToken from 'components/common/field-token';
 import {Clock} from 'components/common/icons/index';
 import {parseFieldValue} from 'utils/data-utils';
@@ -156,16 +157,17 @@ function DataGridFactory(
     _cellRenderer = ({columnIndex, key, rowIndex, style}) => {
       const {columns, rows} = this.props;
 
-      const isLast = columnIndex === columns.length - 1;
-
-      // rowIndex -1 because data rows start rendering at index 1 and we normalize back using the -1 param
-      const className = `${rowIndex === 0
-        ? `${isLast ? 'last' : ''} header-${columnIndex}`
-        : `${isLast ? 'last' : ''} row-${rowIndex-1}`} column-${columnIndex}`;
+      const isLast = columnIndex === columns.length - 1
 
       const type = columns[columnIndex].type;
+      
+      // rowIndex -1 because data rows start rendering at index 1 and we normalize back using the -1 param
+      const className = classnames({
+        last: isLast,
+        [`header-${columnIndex}`]: rowIndex === 0,
+        [`row-${rowIndex-1}} column-${columnIndex}`]: rowIndex > 0
+      });
 
-      // console.error(className);
       return (
         <div key={key} style={style} className={className}>
           {rowIndex === 0
--- a/src/components/common/datagrid.js
+++ b/src/components/common/datagrid.js
@@ -22,9 +22,11 @@ import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {MultiGrid} from 'react-virtualized';
 import styled, {withTheme}from 'styled-components';
+import {createSelector} from 'reselect';
 import FieldToken from 'components/common/field-token';
 import {Clock} from 'components/common/icons/index';
-import {parseFieldValue} from '../../utils/data-utils';
+import {parseFieldValue} from 'utils/data-utils';
+import {ALL_FIELD_TYPES} from 'constants';
 
 const DataGridWrapper = styled.div`
   .ReactVirtualized__Grid:focus,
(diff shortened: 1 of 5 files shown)


src/components/modals/data-table-modal.js
```diff
@@ -90,8 +90,19 @@ function DataTableModalFactory(DataGrid) {
     const activeDataset = datasets[dataId];
     const rows = activeDataset.data;
 
-    const columns = activeDataset.fields
-      .filter(({name}) => name !== '_geojson');
+    const {columns, dataColumnMap} = activeDataset.fields.reduce((acc, field, index) => {
+      const valid = field.name !== '_geojson';
```
heshan0131: We can add a function to parse `_geojson` field instead of filter them out. 
I think it's useful to display it, especially when people uploaded csv contains `WKT` and `Feature`. Put it in `data-utils` maybe, so it can be used in other places

```js
const parseFieldValue = (value, type) => {

  if (!notNullorUndefined(value)) {
    return '';
  }

  switch (type) {
    case ALL_FIELD_TYPES.boolean:
      return String(value);
    case ALL_FIELD_TYPES.geojson:
      return typeof value === 'string' ? value :
      isPlainObject(value) ? JSON.stringify(value) :
      Array.isArray(value) ? `[${String(value)}]` : '';
    default:
      return value;
  }
}
```

src/components/common/datagrid.js
```diff
@@ -90,9 +90,13 @@ const StyledCell = styled.div`
   height: 100%;
 `;
 
-const CellTemplate = ({name, type}) => {
-  const value = type === ALL_FIELD_TYPES.boolean ? String(name) : name;
-  return (<StyledCell title={value}><span>{value}</span></StyledCell>);
+const CellTemplate = ({value, type}) => {
+  const content = type === ALL_FIELD_TYPES.boolean ? String(value) : value;
+  return (
+    <StyledCell title={content}>
```
heshan0131: why `title`?

src/components/common/datagrid.js
```diff
@@ -90,9 +90,13 @@ const StyledCell = styled.div`
   height: 100%;
```
heshan0131: when `span` is in `StyledCell` text-overflow css needs to be applied to `span`, instead of the parent `div`
```
  span {
    overflow: scroll;
    text-overflow: ellipsis;
    white-space: pre-wrap;
    word-wrap: break-word;
    width: 100%;
  }
```

src/components/common/datagrid.js
```diff
@@ -125,12 +129,19 @@ function DataGridFactory(
 
       const type = columns[columnIndex].type;
 
+      let content = null;
+      if (rowIndex) {
+        const {dataColumnMap = {}} = this.props;
+        const dataIndex = dataColumnMap[columnIndex] !== null || dataColumnMap[columnIndex] !== undefined
+          ? dataColumnMap[columnIndex] : columnIndex;
+        content = (<Cell className={`cell ${type}`} value={rows[rowIndex - 1][dataIndex]} type={type} />);
```
heshan0131: `className` here will not be displayed, because you are not rendering it in `Cell`

src/components/common/datagrid.js
heshan0131: Also maybe add left padding to the first cell
```
const DataGridWrapper = styled.div`
   .....other css
  .ReactVirtualized__Grid {
    .column-0 .cell{
      padding-left: 24px;
    }
  }
```

src/components/common/datagrid.js
```diff
@@ -90,9 +90,13 @@ const StyledCell = styled.div`
...
+const CellTemplate = ({value, type}) => {
+  const content = type === ALL_FIELD_TYPES.boolean ? String(value) : value;
+  return (
+    <StyledCell title={content}>
+      <span>{content}</span>
+    </StyledCell>
+  );
 };
 
 export const CellFactory = () => CellTemplate;
```
heshan0131: this could cause compile and uglify error

```js
export const CellFactory = () => {
const CellTemplate = ({value, type}) => {}
return CellTemplate
}

```

src/components/common/datagrid.js
macrigiuseppe: title will add the tooltip effect on mouse hover

src/components/common/datagrid.js
macrigiuseppe: fixed

src/components/common/datagrid.js
macrigiuseppe: fixed

macrigiuseppe:
@heshan0131 updated with your comments and made a small change to make geojson columns easier to read.

### Non-geojson columns
679-table-issue

### Geojson Columns
679-table-issue

src/components/common/datagrid.js
```diff
@@ -115,29 +148,42 @@ function DataGridFactory(
...
+      columns.some(c => c.type === ALL_FIELD_TYPES.geojson)
+    );
+
     _cellRenderer = ({columnIndex, key, rowIndex, style}) => {
       const {columns, rows} = this.props;
 
+      const isLast = columnIndex === columns.length - 1;
+
       // rowIndex -1 because data rows start rendering at index 1 and we normalize back using the -1 param
       const className = `${rowIndex === 0
```
heshan0131: you can use `classnames` to construct this, easier to follow
```js
      const className = classnames(`column-${columnIndex}`, {
        last: isLast,
        [`header-${columnIndex}`]: rowIndex === 0,
        [`row-${rowIndex-1}}`]: rowIndex > 0
      });```

macrigiuseppe: definitely better

Issue #684, opened 2019-08-13
Deck.gl Versioning Issue


treyhakanson:
**Describe the bug**
The `deck.gl` version is set to `^7.1.5` in the `package.json`, and the lockfile specifies `7.1.5`. So, when cloning and running this repository as-is, everything works properly. However, when using `yarn add kepler.gl` in an external project, `deck.gl@7.2.1` is installed. This leads to the following error upon attempting to load data (occurs within `deck.gl`):

```
Uncaught TypeError: Cannot assign to read only property 'mousePosition' of object '[object Object]'
```

**To Reproduce**
This can be simulated in this repository by removing the lockfile:
1. Clone repo, navigate into directory
2. `rm yarn.lock`
3. `yarn --ignore-engines`
4. `yarn start`
5. Go to web browser, attempt to load a dataset
6. See error in console

**Expected behavior**
`package.json` should target only compatible versions of `deck.gl`, allowing for `yarn add` to external projects.

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser: N/A
 - Version: 1.1.3
 - Yarn Version: 1.15.2
 - Node Version: 11.14.0

treyhakanson:
For those looking for a fix in the meantime, you can add the following to your `package.json`:

```json
"resolutions": {
    "deck.gl": "7.1.5"
 }
```

heshan0131:
We will lock it to 7.1.5 in kepler.gl.

treyhakanson:
Awesome, thanks @heshan0131! I went ahead and opened a PR for this to save you some time:
https://github.com/keplergl/kepler.gl/pull/688

ianhansborough:
Thanks for the workaround @treyhakanson! @heshan0131 any idea when can we expect an official patch for this?

heshan0131:
keplerl.gl@1.1.6 is released with this

PR #688, opened 2019-08-17
[Bug] Lock deck.gl to version 7.1.5

treyhakanson:
Locking `deck.gl` to version `7.1.5`  as per this issue: https://github.com/keplergl/kepler.gl/issues/684#issuecomment-522102726

Signed-off-by: treyhakanson <treyhakanson@gmail.com>


package.json
```diff
@@ -83,7 +83,7 @@
     "d3-hexbin": "^0.2.2",
     "d3-request": "^1.0.6",
     "d3-scale": "^1.0.6",
-    "deck.gl": "^7.1.5",
+    "deck.gl": "7.1.5",
```
heshan0131: let's lock luma.gl as well. Since they come in pairs

treyhakanson: Done; locked to the version that install with this version of deck.gl (7.1.1, as already specified)

PR #699, opened 2019-08-26
[Feat] Add Trip Layer - Final

heshan0131:
trip_layer_sf


src/components/common/animation-control/floating-time-display.js
```diff
@@ -0,0 +1,186 @@
...
+import React, {PureComponent} from 'react';
+import styled from 'styled-components';
+import moment from 'moment';
+import {createSelector} from 'reselect';
+import {Minus} from 'components/common/icons';
+import {DEFAULT_TIME_FORMAT} from 'constants/default-settings';
+
+const StyledTimeDisplay = styled.div`
+  background-color: ${props => props.theme.panelBackground};
+  border-radius: 32px;
```
macrigiuseppe: past of the theme?

src/components/common/animation-control/floating-time-display.js
```diff
@@ -0,0 +1,186 @@
...
+  bottom: ${props => `calc(100% + ${props.theme.bottomPanelGap}px)`};
+  color: ${props => props.theme.titleTextColor};
+  display: flex;
+  height: 64px;
```
macrigiuseppe: same

src/components/common/animation-control/playback-controls.js
```diff
@@ -0,0 +1,86 @@
...
+
+function AnimationPlaybacksFactory() {
+  const AnimationPlaybacks = ({
+    isAnimatable,
+    isAnimating,
+    buttonStyle,
+    pauseAnimation = () => {},
+    updateAnimationTime = () => {},
+    startAnimation = () => {},
+    buttonHeight = '18px'
```
macrigiuseppe: constant

src/components/common/animation-control/animation-control.js
```diff
@@ -0,0 +1,197 @@
...
+                maxValue={domain ? domain[1] : 1}
+                value1={currentTime}
+                onSlider1Change={this.onSlider1Change}
+                enableBarDrag={true}
+              />
+            </SliderWrapper>
+            <StyledDomain className="animation-control__time-domain">
+              <span>{moment.utc(domain[1]).format(defaultTimeFormat)}</span>
+            </StyledDomain>
+            <div style={{marginRight: '-10px'}}>
```
macrigiuseppe: constant for the value of the margin. 
Another improvement would be to define the entire object as constant rather than inline

src/components/common/animation-control/floating-time-display.js
```diff
@@ -0,0 +1,186 @@
...
+`;
+
+const TimeDivider = () => (
+  <StyledHorizontalBar>
+    <Minus height="12px" />
+  </StyledHorizontalBar>
+);
+
+const TimeDisplayRow = ({timeValues = []}) => (
+  <div style={{display: 'flex'}}>
```
macrigiuseppe: nit: defining styling object as contants rather than inline would save re-render cycles

macrigiuseppe:
I love this:

Screen Shot 2019-09-19 at 3 36 29 PM

We should do the same thing for other features

macrigiuseppe:
@heshan0131 based on the logic and data demo, when we have both time filter and trip layer, only trip layer animation control will be visible unless the trip layer is hidden, is that correct?

macrigiuseppe:
nit: I found a potential css issue with the filter dropdown:
datetime-filter

If the name of the field is too long, the trash and clock icon is pushed to the right

src/components/common/animation-control/animation-control.js
```diff
@@ -0,0 +1,197 @@
...
+    };
+
+    render() {
+      const {currentTime, domain, speed} = this.props.animationConfig;
+      const {showSpeedControl} = this.state;
+
+      return (
+        <BottomWidgetInner className="bottom-widget--inner">
+          <AnimationWidgetInner className="animation-widget--inner">
+            <div style={{marginLeft: '-10px'}}>
```
macrigiuseppe: constant

src/components/common/slider/slider.js
```diff
@@ -40,7 +40,7 @@ const StyledRangeSlider = styled.div`
 
 const SliderWrapper = styled.div`
   flex-grow: 1;
-  margin-top: ${props => (props.isRanged ? 0 : 10)}px;
+  margin-top: ${props => (props.isRanged ? 0 : 12)}px;
```
macrigiuseppe: constant

src/components/filters/time-widget.js
```diff
@@ -85,37 +73,21 @@ const TopSectionWrapper = styled.div`
...
-  height: 24px;
-  margin-right: 4px;
-  text-align: center;
-  width: 24px;
-  line-height: 24px;
-
-  :hover {
-    cursor: pointer;
+    .animation-control__speed-slider {
+      right: calc(0% - 48px);
```
macrigiuseppe: constant

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+} from 'layers/geojson-layer/geojson-utils';
+
+/**
+ * Parse geojson from string
+ * @param {array} geojson feature object values
+ * @returns {boolean} whether the geometry coordinates has length of 4
+ */
+export function coordHasLength4(samples) {
+  let hasLength4 = true;
+  for (let i = 0; i < samples.length; i += 1) {
```
macrigiuseppe:
```javascript
const FIELDS_NUMBER = 4;
return samples.every(sample => sample..geometry.coordinates.every(c => c.length === FIELDS_NUMBER));
```

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+ * Check if geojson features are trip layer animatable by meeting 3 conditions
+ * @param {array} features array of geojson feature objects
+ * @returns {boolean} whether it is trip layer animatable
+ */
+export function isTripGeoJsonField(allData = [], field) {
+  if (!allData.length) {
+    return false;
+  }
+  const getValue = d => d[field.tableFieldIndex - 1];
+  const maxCount = 10000;
```
macrigiuseppe: define this constant outside the method

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const features = sampleRawFeatures.map(parseGeoJsonRawFeature);
+
+  let isTrip = false;
+  const featureTypes = getGeojsonFeatureTypes(features);
+
+  // condition 1: contain line string
+  const hasLineString = Boolean(featureTypes.line);
+  if (!hasLineString) {
+    return isTrip;
```
macrigiuseppe: rather than using isTrip variable, can't you just return false

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  // condition 2:sample line strings contain 4 coordinates
+  const HasLength4 = coordHasLength4(features);
+  if (!HasLength4) {
+    return isTrip;
```
macrigiuseppe: same here

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  // condition 3:the 4th coordinate of the first feature line strings is valid time
+  const tsHolder = features[0].geometry.coordinates.map(
+    coord => coord[3]
+  );
+
+  const hasValidTime = containValidTime(tsHolder);
+  if (hasValidTime) {
```
macrigiuseppe: and here you can do the following:
```javascript
const hasValidTime = containValidTime(tsHolder);
return hasValidaTime;
```

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+      ? f.geometry.coordinates.map(getTimeValue)
+      : null
+  );
+
+  const animationDomain = getAnimationDomainFromTimestamps(dataToTimeStamp);
+
+  return {dataToTimeStamp, animationDomain};
+}
+
+function findMinFromSorted(list = []) {
```
macrigiuseppe: can you use __find__ to find the value you are interested in?
```
return list.find(value => notNullorUndefined(value))
```

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+  while (i < list.length) {
+    if (notNullorUndefined(list[i])) {
+      return list[i]
+    }
+    i++;
+  }
+  return null;
+}
+
+function findMaxFromSorted(list = []) {
```
macrigiuseppe: here you can take advantage of inverse:
```javascript
return list.reverse().find(value => notNullorUndefined(value));
```

src/layers/trip-layer/trip-utils.js
```diff
@@ -0,0 +1,188 @@
...
+}
+
+export function getAnimationDomainFromTimestamps(dataToTimeStamp = []) {
+  return dataToTimeStamp.reduce(
+    (accu, tss) => {
+      accu[0] = Math.min(accu[0], findMinFromSorted(tss));
+      accu[1] = Math.max(accu[1], findMaxFromSorted(tss));
+      return accu;
+    },
+    [Number(Infinity), -Infinity]
```
macrigiuseppe: I don't think you need to use Number. I tested with the console and:
```
Number(Infinity) === Infinity
true
```

Otherwise, if you want to use Number, it's better to be consistent and use it for both values
```javascript
[Number(Infinity), Number(-Infinity)]
```

heshan0131:
I was able to get 2 time playback showing at the same time. Can you try reproduce it?

Screen Shot 2019-09-20 at 1 54 12 PM


Also fixed time filter display

Screen Shot 2019-09-20 at 1 54 05 PM

src/components/filters/time-widget.js
heshan0131: don't think we have to use constant in css everywhere, this is only so that the slider will align with the edge

src/layers/trip-layer/trip-utils.js
heshan0131: find and for loop is more performative, it exists at the first unmatched encounter.

src/layers/trip-layer/trip-utils.js
heshan0131: not sure if it's neccessary, this only used here

src/layers/trip-layer/trip-utils.js
heshan0131: not sure if it's neccessary, this only used here

src/layers/trip-layer/trip-utils.js
heshan0131: the list can be really long, when it's sorted, starting from the end  and immediately return when found

src/layers/trip-layer/trip-utils.js
heshan0131: as the name suggested, this is looping through a sorted array, starting from index = 0 and immediately exist if anything is found

src/layers/trip-layer/trip-utils.js
macrigiuseppe: "every" also exits as first non-matched element

src/layers/trip-layer/trip-utils.js
macrigiuseppe: "find" also returns immediately as the first element is found

macrigiuseppe:
> I was able to get 2 time playback showing at the same time. Can you try reproduce it?
> 
> Screen Shot 2019-09-20 at 1 54 12 PM
> 
> Also fixed time filter display
> 
> Screen Shot 2019-09-20 at 1 54 05 PM

yes i can

src/layers/trip-layer/trip-utils.js
heshan0131: 👍

src/layers/trip-layer/trip-utils.js
heshan0131: 👍

src/layers/trip-layer/trip-utils.js
macrigiuseppe: actually i was reading some performance reports and it looks liek reverse is not as performant as doing a while loop. Good with this

macrigiuseppe:
Looks good. Also tested export and import map

Issue #658, opened 2019-08-05
Disable side panel when embedding map as html/json


guyts:
**Is your feature request related to a problem? Please describe.**
I am trying to export a visual I created to embed on a blogpost. But I can't find an option to export without the side panel to embed as the dynamic map alone.

**Describe the solution you'd like**
There will be an export without the side panel, such that whomever it is shared with, cannot edit, export or change the map, but only interact with the map.

**Describe alternatives you've considered**
I tried to find a field or parameter in the html / json of the export to disable the panel, but couldn't spot it.

omegadeep10:
Hi, we ran into a similar issue. I wanted to embed a map on our conference area TV, without the sidebar. We are using the single file HTML version.

Note that Kepler already supports loading a "read Only" version of the map, you just have to pass the option in the `initialState` object as part of the reducer. I'm not a react person, so a lot of this goes over my head. I used the provided example and modified it slightly based on the params I saw in the Kepler docs.

We were able to configure to kepler to load a readOnly version of the map if we passed a param titled `readOnly` in the URL. The relevant code is shown here:
![image](https://user-images.githubusercontent.com/14183626/62952598-f8a9fc80-bdb9-11e9-88fa-54f4a27811f8.png)

The full file is attached. It will not work because it references URLs and files only available on our internal network.

Hope this helps!
[example.kepler.zip](https://github.com/keplergl/kepler.gl/files/3497587/example.kepler.zip)

guyts:
Thanks @omegadeep10 ! I actually managed a similar fix as well (not a react person myself too), but forgot to update here! 
For reference, here's my fix:
https://stackoverflow.com/questions/57364050/disable-side-panel-when-exporting-kepler-gl-map/57381970#57381970

guyts:
Accidentally closed ! But if anyone wants to do this as enhancement, leaving open.

heshan0131:
Adding `readOnly: true` to `keplerGlReducer.initialState` is the simple solution here. 
We can also add an option to kepler.gl export map modal whether to disable side-panel.

PR #709, opened 2019-09-03
[Feat] HTML Export: provide read only mode

macrigiuseppe:
This PR implements #658.

Created new UI to pick between READ and EDIT mode when exporting html Map (see below).

Screen Shot 2019-09-03 at 5 19 49 PM

Result:
Screen Shot 2019-09-03 at 6 28 37 PM

#### READ
export-read-mode

#### EDIT
export-edit-mode


src/reducers/combined-updaters.js
```diff
@@ -75,7 +75,6 @@ const combinedUpdaters = null;
 
 export const isValidConfig = config => isObject(config) && isObject(config.config) && config.version;
 export const defaultAddDataToMapOptions = {
-  readOnly: false,
```
macrigiuseppe: I noticed if the application customizes kepler initial state with readOnly: true, the custom properties is lost because of this setting

src/reducers/combined-updaters.js
```diff
@@ -155,10 +154,11 @@ export const addDataToMapUpdater = (state, {payload}) => {
   mergedState = {
     ...mergedState,
     uiState: {
+      ...mergedState.uiState,
       ...toggleModalUpdater(mergedState.uiState, {payload: null}),
-      readOnly: options.readOnly
+      ...((options.readOnly !== undefined && options.readOnly !== null) ? {readOnly: options.readOnly} : {})
```
heshan0131: `options.hasOwnProperty('readOnLy')`?

macrigiuseppe: done

Issue #719, opened 2019-09-06
[Bug] Generate new layer should based on a more sophisticated logic


heshan0131:
**Describe the bug**
We only compare `mergedState.layers.length === previousState.layers.length` to decide whether to find some default layers. 

This triggers a edge case bug: When add data with a preset config to a map that already has layers. and current map contains same number of layers as the preset config. After merging the state with preset layer config, kepler wtill thinks no new layers have been merged, therefore create default layers.

**Expected behavior**
We should be comparing whether new layers depends on current new datasets are merged in before adding default ones

@Firenze11

PR #721, opened 2019-09-10
[Bug] check new layers based on new dataset id

heshan0131:
- Bug: when add data with a preset config to a map that already has layers.  If `oldState.layers.length` happens to equal to `mergedState.layers.length`, kepler thinks no new layers has been merged, therefore create default layers.
- Fixe: Check if any new layers are merged from config based on new dataset id
#719 

### DIFF
--- a/test/node/reducers/vis-state-test.js
+++ b/test/node/reducers/vis-state-test.js
@@ -677,26 +677,6 @@ test('#visStateReducer -> UPDATE_VIS_DATA.1 -> No data', t => {
 test('#visStateReducer -> UPDATE_VIS_DATA.2 -> to empty state', t => {
   const oldState = InitialVisState;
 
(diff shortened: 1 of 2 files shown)


test/node/reducers/vis-state-test.js
```diff
@@ -673,22 +674,28 @@ test('#visStateReducer -> UPDATE_VIS_DATA.1', t => {
...
-      info: {id: 'smoothie', label: 'exciting dataset 1'}
-    })
-  );
-
-  t.deepEqual(
-    Object.keys(testState1.datasets),
-    ['smoothie'],
-    'should save data to smoothie'
-  );
+  // const testState1 = reducer(
```
macrigiuseppe: If no longer necessary, delete these tests otherwise re-enable them

PR #725, opened 2019-09-10
[Bug] Fixed env variable webpack injection when running samples independently

macrigiuseppe:
Samples were missing env variables in webpack configuraiton

### DIFF
--- a/examples/demo-app/webpack.config.js
+++ b/examples/demo-app/webpack.config.js
@@ -70,10 +70,11 @@ const CONFIG = {
     historyApiFallback: true
   },
 
-  // Optional: Enables reading mapbox and dropbox client token from environment variable
   plugins: [
     new webpack.EnvironmentPlugin([
-      'MapboxAccessToken'
+      'MapboxAccessToken',
+      'DropboxClientId',
+      'MapboxExportToken'
     ])
   ]
 };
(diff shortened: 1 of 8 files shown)


heshan0131:
why do we need to add `DropboxId` to none demo-app examples?

macrigiuseppe:
> why do we need to add `DropboxId` to none demo-app examples? saving to dropbox is not a core feature, it shouldn't matter to code examples.

@heshan0131 

my mistake I added to all of them but I meant to add them the demo-app sample.

Note: if we run the samples from the root folder npm start:demo-app, we are already looking for those environment variables because we use the common webpack config (webpack.config.local.js).

The demo app has all features (dropbox and export map) and if we want to run it as standalone we should be able to read all variables. If dropbox id and mapbox export token variables are not passed, core functionalities will be working as always.

The PR will enable reading the env variables when demo-app is running as standalone (from its own folder)

examples/demo-app/webpack.config.js
```diff
@@ -70,10 +70,11 @@ const CONFIG = {
     historyApiFallback: true
   },
 
-  // Optional: Enables reading mapbox and dropbox client token from environment variable
```
heshan0131: Should add this to README.md in demo-app

Issue #729, opened 2019-09-13
React 16.9 - componentWillMount and componentWillReceiveProps will not be supported in future React


heshan0131:
**Is your feature request related to a problem? Please describe.**
<img width="776" alt="Screen Shot 2019-09-12 at 5 11 10 PM" src="https://user-images.githubusercontent.com/3605556/64829218-59c31c80-d580-11e9-963e-b1ced692611c.png">

macrigiuseppe:
This requires styled-components upgrade to newest react version.

PR #745, opened 2019-09-23
[Enhancement] React 17: replace componentWillReceiveProps and componentWillMount

heshan0131:


heshan0131:
https://github.com/keplergl/kepler.gl/issues/729

examples/open-modal/src/app.js
```diff
@@ -32,8 +32,8 @@ import SavedMap from './components/saved-map';
 const MAPBOX_TOKEN = process.env.MapboxAccessToken; // eslint-disable-line
 
 class App extends Component {
-  componentWillReceiveProps(nextProps) {
-    if (!this.props.keplerGl.bar && nextProps.keplerGl.bar) {
+  componentDidUpdate(prevProps) {
+    if (!this.props.keplerGl.bar && prevProps.keplerGl.bar) {
```
chucksellick: Looks like you broke the the logic here, since nextProps became prevProps it should be the other way around: `this.props.keplerGl.bar && !prevProps.keplerGl.bar`

src/components/common/item-selector/typeahead.js
```diff
@@ -138,13 +191,21 @@ export default class Typeahead extends Component {
...
+
+    return {searchResults};
+  }
+
   constructor(props) {
     super(props);
 
     this.state = {
-      searchResults: this.getOptionsForValue(
+      searchResults: getOptionsForValue(
```
chucksellick: Surely not necessary to do this here now since it is handled by `getDerivedStateFromProps`?

heshan0131: 👍

examples/open-modal/src/app.js
heshan0131: 👍

PR #753, opened 2019-09-30
Applied the repo's eslint/prettier formatting

chrisbrainerd:
Added a `prettier-fix` script to the package.json and then ran it. - no other changes.

Getting this over with will make it easier to maintain over time 🥳

I'd also personally love to add in a pre-commit hook to run that script but that's something I wouldn't want to do without socializing a bit too.


chrisbrainerd:
Missed the outstanding PR doing the same 🙃

Issue #592, opened 2019-06-30
Saved HTML file keeps flashing when opening with browser


HLiuUS:
**Describe the bug**
When the map contains a filter, the map keeps flashing when opened with a browser

**To Reproduce**
1. Created a map within Jupyter Notebook
2. Add a filter 
3. Call .save_to_html method to save to HTML

**Environment (please complete the following information):**
 - Python version: python3
 - keplergl Widget version: 
[test-fake1.zip](https://github.com/keplergl/kepler.gl/files/3343163/test-fake1.zip)

0.1.0.a7

rubenseijioto:
I'm having the same issue and for it's happening when I add a timestamp filter, other types works fine for me.

HLiuUS:
> I'm having the same issue and for it's happening when I add a timestamp filter, other types works fine for me.

Exactly the same thing happened to me.

heshan0131:
@HLiuUS What browser (chrome, safari, firefox) and operating system (mac, windows) are you on?
mind attach the html?

HLiuUS:
@heshan0131 It was on Windows 10 and I tried with both Chrome and Edge, same issue. You can find a html file in the zip file I uploaded.

heshan0131:
@macrigiuseppe will look into this

Bwsutton1990:
> I'm having the same issue and for it's happening when I add a timestamp filter, other types works fine for me.

This seems to be the same for me. It stops flashing when I remove it from the map by selecting "Time Playback" in the filter section. But starts again when I select it again.

macrigiuseppe:
Based on my findings it looks like the browser page keeps on showing/hiding scrollbars and creates a flashing effect.

This is the recording

https://drive.google.com/file/d/1V_kQWjZD8L4Qgtn1bEbBBHX5YYXAS_1b/view?usp=sharing

@Bwsutton1990 @HLiuUS can you please confirm that's the actual problem you are experiencing?

If not, can you please provide a recording of your issue?

Thank you

Bwsutton1990:
Hi @macrigiuseppe 

I cant access the file, but having looked at the flashing more closely, that seems to make sense.

Thanks
Ben

HLiuUS:
Hi @macrigiuseppe, I cant access the file either, but it seemscorrect based on your description.

macrigiuseppe:
@Bwsutton1990 @HLiuUS i updated the link. Let me know

Bwsutton1990:
[...]

PR #756, opened 2019-10-01
[Bug] [Jupyter] fix flashing html export when open in window

heshan0131:
- Upgrade to `kepler.gl@1.1.8`
- Remove AutoSizer, user React hooks to handle resize


heshan0131:
#592

Issue #739, opened 2019-09-20
Allow users to create a filter and operate across multiple datasets


macrigiuseppe:
Implement filter logic to be applied across multiple datasets

PR #773, opened 2019-10-07
[Feat] Multiple datasets per filter

macrigiuseppe:
Issue: #739 
This PR implements the business logic to apply a single filter to multiple datasets.

Changes:
Filter structure has been updated with the following:
```
dataId: string -> [string]
fieldIdx: string -> [string]
```

Each field used to filter on will store the filter props computed for the given filter/dataset combination in order to support operation such as time filtering.

Few new helpers have been created to support filtering operations:
```
applyFilterFieldName: used when the filter name is updated to apply the right changes to the both filters and datasets, including field props
applyFiltersToDatasets
getDatasetIndexForFilter
getDatasetFieldIndexForFilter
```


src/reducers/vis-state-merger.js
```diff
@@ -493,60 +499,3 @@ export function validateLayerWithData(
 
   return newLayer;
 }
-
```
macrigiuseppe: moved to filter utils

src/schemas/vis-state-schema.js
```diff
@@ -648,7 +648,8 @@ export const filterPropsV1 = {
       name: null,
       type: null
     }
-  })
+  }),
+  fieldIdx: null
```
macrigiuseppe: this is used to export filedIdx into the exported configuration

src/utils/filter-utils.js
```diff
@@ -275,9 +374,11 @@ export function filterData(data, dataId, filters) {
  * @param {Object[]} data
  * @param {Object} filter
  * @param {number} i
+ * @param {field} field containing values to test data against. This is used only when
+ * testing timestamp filters
  * @returns {Boolean} - whether value falls in the range of the filter
  */
-export function isDataMatchFilter(data, filter, i) {
+export function isDataMatchFilter(data, filter, i, field) {
   const val = data[filter.fieldIdx];
```
heshan0131: `filter.fieldIdx` should be one of the `fieldIdx` you are filtering based on. or simply use `field.tableFieldIndex - 1`

`const val = data[field.tableFieldIndex - 1];`

src/utils/filter-utils.js
```diff
@@ -288,9 +389,10 @@ export function isDataMatchFilter(data, filter, i) {
       return isInRange(val, filter.value);
 
     case FILTER_TYPES.timeRange:
-      const timeVal = filter.mappedValue
-        ? filter.mappedValue[i]
-        : moment.utc(val).valueOf();
+      const timeVal = field && field.filterProp && field.filterProp.mappedValue
```
heshan0131: `Array.isArray(field.filterProp.mappedValue)` is a safer then `field.filterProp.mappedValue`

src/utils/filter-utils.js
```diff
@@ -288,9 +389,10 @@ export function isDataMatchFilter(data, filter, i) {
...
+        ? field.filterProp.mappedValue[i]
+        : moment.utc(data[field.tableFieldIndex - 1]).valueOf();
```
heshan0131: use `val` here

src/utils/filter-utils.js
```diff
@@ -563,3 +664,183 @@ export function getDefaultFilterPlotType(filter) {
...
+
+/**
+ *
+ * @param datasetIds list of dataset ids
+ * @param datasets
+ * @param filters
+ * @return {{[p: string]: *}}
+ */
+export function applyFiltersToDatasets(datasetIds, datasets, filters) {
+  const dataIds = Array.isArray(datasetIds) ? datasetIds : [datasetIds];
```
heshan0131: Could collect a unique list of `datasetId` based on filters passed. So that we don't need to filter through all datasets

src/reducers/vis-state-merger.js
```diff
@@ -52,43 +51,49 @@ export function mergeFilters(state, filtersToMerge) {
...
+
+      const validatedFilter = datasetIds.reduce((acc, d) => {
+        const updatedFilter = validateFilterWithData(datasets[d], filter);
+        if (updatedFilter) {
+          return acc ? {
+            ...acc,
+            // merge props from every single dataset/field
+            ...mergeFilterProps(acc, updatedFilter)
+          } : updatedFilter;
+        }
```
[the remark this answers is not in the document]
macrigiuseppe: in this scenario, we update the accumulator only when updatedFilter is not null. The else case is not applicable

src/utils/filter-utils.js
```diff
@@ -563,3 +664,183 @@ export function getDefaultFilterPlotType(filter) {
...
+
+    // if no field with same name is found, move to the next datasets
+    if (fieldIndex === -1) {
+      throw new Error(`fieldIndex not found. Dataset must contain a property with name: ${fieldName}`);
+    }
+
+    // TODO: validate field type
+    const field = fields[fieldIndex];
+    const filterProps = getFilterProps(allData, field);
+    const filterDatasetIndex = getDatasetIndexForFilter(datasets[dataIdentifier], filter);
```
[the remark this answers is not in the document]
macrigiuseppe: yeah, nice catch

src/utils/filter-utils.js
```diff
@@ -563,3 +664,183 @@ export function getDefaultFilterPlotType(filter) {
...
+ * @param fieldName
+ * @return {object} {filter, datasets}
+ */
+export function applyFilterFieldName(filter, datasets, fieldName) {
+  const {dataId} = filter;
+  if (!dataId || !dataId.length) {
+    return filter;
+  }
+
+  return dataId.reduce((acc, dataIdentifier) => {
```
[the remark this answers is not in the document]
macrigiuseppe: at this point dataId should always be an array

src/utils/filter-utils.js
macrigiuseppe: nice

src/reducers/vis-state-updaters.js
```diff
@@ -783,13 +762,7 @@ export const removeFilterUpdater = (state, action) => {
...
   const newState = {
     ...state,
-    datasets: {
-      ...state.datasets,
-      [dataId]: {
-        ...state.datasets[dataId],
-        ...filterData(state.datasets[dataId].allData, dataId, newFilters)
-      }
-    },
+    datasets: applyFiltersToDatasets(dataId, state.datasets, newFilters),
```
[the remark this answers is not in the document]
macrigiuseppe: we only filter the dataset with dataId as id value, we could change the method of applyFiltersToDatasets to accept a list of datasets to filter and newFilters

src/schemas/vis-state-schema.js
macrigiuseppe: converted name into a list

src/reducers/vis-state-merger.js
```diff
@@ -52,43 +51,49 @@ export function mergeFilters(state, filtersToMerge) {
...
   filtersToMerge.forEach(filter => {
-    // match filter.dataId with current datesets id
-    // uploaded data need to have the same dataId with the filter
-    if (datasets[filter.dataId]) {
-      // datasets is already loaded
-      const validateFilter = validateFilterWithData(
-        datasets[filter.dataId],
-        filter
-      );
+    const datasetIds = Object.keys(datasets).reduce((acc, datasetId) => ([
```
heshan0131: There is one scenario when 1 filter applies to multiple dataset ['a', 'b']. however, only 'a' is loaded at the moment. We should handle that scenario. in your current implementation, this filter will be sent to unmerged. because `validateFilterWithData`

heshan0131: why use reduce when this is just a simple filter function?
`datasetIds = Object.keys(datasets).filter(datasetId => shouldApplyfilter(filter, datasetId))`

src/reducers/vis-state-merger.js
```diff
@@ -52,43 +51,49 @@ export function mergeFilters(state, filtersToMerge) {
...
+      ...acc,
+      ...(shouldApplyfilter(filter, datasetId) ? [datasetId] : [])
+    ]), []);
+
+    if (datasetIds.length) {
+
+      // for each dataset data mapped against the current filter are valid
+      const validatedFilter = datasetIds.reduce((acc, d) => {
```
heshan0131: There is a scenario when 1 filter applies to multiple datasets ['a', 'b']. however, only 'a' is loaded at the moment. We should handle that scenario. in your current implementation, this filter will be merged with only 1 loaded dataset. I think we should send the filter to `unmerged` and wait until all datasets is loaded to merge it.



That's why I was asking what is the return value here

src/reducers/vis-state-merger.js
```diff
@@ -52,43 +51,49 @@ export function mergeFilters(state, filtersToMerge) {
...
-      [dataId]: {
-        ...datasets[dataId],
-        ...filterData(datasets[dataId].allData, dataId, updatedFilters)
-      }
-    }),
-    datasets
-  );
+
+  // flatten all filter dataIds
+  const datasetsToFilter = uniq(merged.reduce((acc, f) => ([
```
heshan0131: A more readabble version of this would be

`
import flattenDeep from 'lodash.flattendeep`
const datasetsToFilter = uniq(flattenDeep(merged.map(f => f.dataId)))
`

src/reducers/vis-state-updaters.js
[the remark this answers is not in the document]
heshan0131: Yes, but if your filter is applied to ['a' and 'b'] , and you changed the field for `a`, you don't need to re-filter `b`

test/node/utils/filter-utils-test.js
```diff
@@ -208,9 +211,10 @@ function testGetTimeFieldDomain(rows, allFields, t) {
   );
 }
 
-function testIsTimeDataMatchFilter(rows, t) {
+function testIsTimeDataMatchFilter(rows, fields, t) {
+
   const timeStringFilter = {
-    fieldIdx: 0,
+    fieldIdx: [ 0 ],
```
heshan0131: space in bracket

test/node/utils/filter-utils-test.js
```diff
@@ -219,19 +223,20 @@ function testIsTimeDataMatchFilter(rows, t) {
...
-    isDataMatchFilter(rows[15], timeStringFilter),
+    isDataMatchFilter(rows[15], timeStringFilter, 15, fields[timeStringFilter.fieldIdx[0]]),
     false,
     `${rows[15][0]} should be outside the range`
   );
 
+
   const epochFilter = {
-    fieldIdx: 4,
+    fieldIdx: [ 4 ],
```
heshan0131: same here

src/reducers/vis-state-merger.js
macrigiuseppe: We will apply a filter only if all datasets are present

macrigiuseppe: I updated the code a little bit and made sure we validate each filter against each dataset present in dataId

src/reducers/vis-state-updaters.js
```diff
@@ -582,15 +573,10 @@ export function setFilterUpdater(state, action) {
...
-      [dataId]: {
-        ...newState.datasets[dataId],
-        ...filterData(allData, dataId, newState.filters)
-      }
-    }
+    datasets: applyFiltersToDatasets(dataId, newState.datasets, newState.filters)
   };
 
+  // dataId is an array
   newState = updateAllLayerDomainData(newState, dataId, newFilter);
```
heshan0131: if a filter is applied to dataId ['a', 'b'], and only the field name of dataset 'a' is updated. you don't neet to filter both 'a' and 'b'. You should only apply filter to dataset at `datasetIdx`

src/reducers/vis-state-updaters.js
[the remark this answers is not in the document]
heshan0131: what about this comment here?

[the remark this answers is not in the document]
macrigiuseppe: In this method, we only remove existing filters and we need to recompute data for those datasets that were mapped against the deleted filter.
When we call applyFiltersToDatasets passing dataId we only process filters that were mapped against the deleted filter.

On the other hand, I made some changes to setFilterUpdater to only process the appropriate dataset based on the setFilter action props, code here: https://github.com/keplergl/kepler.gl/pull/773/files#diff-3c42abf857bb61c5f6578c72771405d2R580

src/utils/filter-utils.js
```diff
@@ -59,6 +58,15 @@ export const PLOT_TYPES = keyMirror({
   lineChart: null
 });
 
+export const FILTER_UPDATER_PROPS = keyMirror({
+  dataId: null,
+  name: null
+});
+
+export const LIMITED_FILTER_EFFECT_PROPS = keyMirror({
```
macrigiuseppe: For this list of properties we can apply filter side effects to only one dataset

PR #775, opened 2019-10-08
[Bug] Filter invalid H3 IDs

isaacbrodsky:
Similarly to how the point layer filters invalid coordinates, the H3 layer behaves strangely if given datasets that have invalid hexagon IDs (an easy test is the ID `0`).

### DIFF
--- a/src/layers/h3-hexagon-layer/h3-hexagon-layer.js
+++ b/src/layers/h3-hexagon-layer/h3-hexagon-layer.js
@@ -23,7 +23,7 @@ import memoize from 'lodash.memoize';
 import Layer from '../base-layer';
 import {GeoJsonLayer, H3HexagonLayer} from 'deck.gl';
 import EnhancedColumnLayer from 'deckgl-layers/column-layer/enhanced-column-layer';
-import {getVertices, getCentroid, idToPolygonGeo} from './h3-utils';
+import {getVertices, getCentroid, idToPolygonGeo, h3IsValid} from './h3-utils';
 import H3HexagonLayerIcon from './h3-hexagon-layer-icon';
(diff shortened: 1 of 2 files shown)


src/layers/h3-hexagon-layer/h3-hexagon-layer.js
```diff
@@ -224,7 +224,7 @@ export default class HexagonIdLayer extends Layer {
 
     allData.forEach((d, index) => {
       const id = getHexId(d);
-      if (typeof id !== 'string' || !id.length) {
+      if (typeof id !== 'string' || !id.length || !h3IsValid(id)) {
```
heshan0131: if `h3IsValid(id)` will cover `id !== 'string' || !id.length` case, then we don't need to check the other 2

isaacbrodsky: Internally, it calls https://github.com/uber/h3-js/blob/master/lib/h3core.js#L84 which looks like it should account for those - if the string is of the wrong length it should return something like `[0, 0]` which will be invalid.

src/layers/h3-hexagon-layer/h3-hexagon-layer.js
```diff
@@ -224,7 +224,7 @@ export default class HexagonIdLayer extends Layer {
 
     allData.forEach((d, index) => {
       const id = getHexId(d);
-      if (typeof id !== 'string' || !id.length) {
+      if (!h3IsValid(id)) {
```
nrabinowitz: This should also remove the need for the `id &&` checks on L233-L234 below.

isaacbrodsky: Ok, also updated

Issue #652, opened 2019-08-02
More scale options


heshan0131:
**Problem**
kepler.gl now only supports quantile and quantize scale

**Describe the solution you'd like**
 Add log, sqrt and Jenkins natural break scale

heshan0131:
#670

PR #809, opened 2019-10-28
[Feat] add stdev and variance aggregators

jefemagril:
Make use of D3 deviation and variance functions to add stdev and variance options to the list of available data aggregators.

This fails `yarn test-node`, however master also fails with the same error.


heshan0131:
#652

PR #835, opened 2019-11-17
[Bug] Fixed delete dataset action

macrigiuseppe:
Fixed issue when deleting dataset.

### DIFF
--- a/src/components/bottom-widget.js
+++ b/src/components/bottom-widget.js
@@ -97,7 +97,7 @@ export default function BottomWidgetFactory(TimeWidget, AnimationControl) {
             updateAnimationSpeed={visStateActions.updateLayerAnimationSpeed}
           />
         ) : null}
-        {enlargedFilterIdx > -1 ? (
+        {enlargedFilterIdx > -1 && Object.keys(datasets).length > 0 ? (
           <TimeWidget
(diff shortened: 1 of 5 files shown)


src/components/bottom-widget.js
```diff
@@ -97,7 +97,7 @@ export default function BottomWidgetFactory(TimeWidget, AnimationControl) {
             updateAnimationSpeed={visStateActions.updateLayerAnimationSpeed}
           />
         ) : null}
-        {enlargedFilterIdx > -1 ? (
+        {enlargedFilterIdx > -1 && Object.keys(datasets).length > 0 ? (
```
heshan0131: this doesn't make sense, how come there is filter but no dataset? what could have happened is when deleting a dataset, the associated filter was not deleted

macrigiuseppe: @heshan0131 you are right. In fact the root cause was related to the filter not being deleted correctly which I fixed https://github.com/keplergl/kepler.gl/pull/835/files#diff-3c42abf857bb61c5f6578c72771405d2R910

I put this as an extra check

heshan0131: Oh nice. Could you add a test case for this?

PR #853, opened 2019-12-03
[Enhancement] Decouple components and create new factories

macrigiuseppe:


src/constants/default-settings.js
```diff
@@ -694,3 +694,6 @@ export const MAP_CONFIG_DESCRIPTION = 'Map config will be included in the Json f
 export const BASE_SPEED = 600;
 export const DEFAULT_TIME_FORMAT = 'MM/DD/YY HH:mm:ssa';
 export const SPEED_CONTROL_RANGE = [0, 10];
+
+// No Op
+export const NO_OP = () => {};
```
heshan0131: do we really need to add a N0_OP to default settings? feel like it's a little overkill, adding one line of import every file to use no_op probably not that pretty

src/components/modal-container.js
```diff
@@ -34,7 +34,7 @@ import DataTableModalFactory from './modals/data-table-modal';
 import LoadDataModalFactory from './modals/load-data-modal';
 import ExportImageModalFactory from './modals/export-image-modal';
 import ExportDataModalFactory from './modals/export-data-modal';
-import ExportMapModalFactory from './modals/export-map-modal';
+import ExportMapModalFactory from './modals/export-map-modal/export-map-modal';
```
heshan0131: you should also export `ExportMapModalFactory` from `components/index.js`, so people can replace it with custom one

jonsadka:
This is awesome 🙏

src/constants/default-settings.js
macrigiuseppe: this was removed

Issue #848, opened 2019-11-27
[Bug] not table to render and export map as image using ClusterLayer


littlebirdalice:
When click on export image, it either shows blank preview or basic map preview without plotted points - I've tried both on safari and chrome and retried many times after restarting browser.
The data I used is very small - 9.5 MB. 

<img width="1354" alt="Screen Shot 2019-11-27 at 1 01 21 AM" src="https://user-images.githubusercontent.com/16776117/69710447-479f4780-10b4-11ea-9d05-d3b6944dc482.png">

macrigiuseppe:
@littlebirdalice would you be able to share your dataset?
I tested using our examples and it works.

macrigiuseppe:
There is an issue with the cluster layer. We are going to fix it.

macrigiuseppe:
fixed

Issue #836, opened 2019-11-18
Cluster map: total points keep the same count when zoom in to a sub cluster


LeslieXong:

macrigiuseppe:
@LeslieXong please provide clear details and steps to reproduce your issue

macrigiuseppe:
@heshan0131 duplicated of #843

PR #855, opened 2019-12-07
[Bug] fix cluster point count, cluster layer failed to render on export image

heshan0131:
- Pass mapState to deck.gl layer prop to avoid width=0, height=0 at initial rendering
- Correct reduce and map function to accumulate points in each cluster


heshan0131:
Screen Shot 2019-12-06 at 9 05 01 PM

heshan0131:
#848 
#836 
#524

PR #856, opened 2019-12-09
[Bug] Remove sqrt, log from default color aggregation for count

heshan0131:
- using Sqrt, log scale in color legend causes app to crash
- Remove sqrt, log from default color aggregation for count

### DIFF
--- a/examples/demo-app/src/app.js
+++ b/examples/demo-app/src/app.js
@@ -124,7 +124,7 @@ class App extends Component {
     //   window.setTimeout(this._showBanner, 3000);
     // }
     // load sample data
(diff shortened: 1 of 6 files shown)


examples/demo-app/src/app.js
```diff
@@ -167,7 +167,7 @@ class App extends Component {
   _loadSampleData() {
     this._loadPointData();
     // this._loadGeojsonData();
-    this._loadTripGeoJson();
+    // this._loadTripGeoJson();
```
akre54: intentional?

heshan0131: this is just for local testing

PR #869, opened 2019-12-19
Fix typo `newDateEntries` -> `newDataEntries`

jonsadka:


jonsadka:
See https://github.com/keplergl/kepler.gl/pull/870

PR #871, opened 2019-12-19
(bugfix) Ensure all colors returned from get3DBuildingColor are RGB arrays

jonsadka:
`<ColorBlock />` expects a color array (https://github.com/keplergl/kepler.gl/blob/master/src/components/side-panel/layer-panel/color-selector.js#L35) but is receiving a string, thus causing a crash.

Screen Shot 2019-12-19 at 3 14 17 PM

Error:
Screen Shot 2019-12-19 at 3 14 30 PM

Signed-off-by: Jon Sadka <jonsadka@uber.com>

### DIFF
--- a/src/reducers/map-style-updaters.js
+++ b/src/reducers/map-style-updaters.js
@@ -189,7 +189,7 @@ function findLayerFillColor(layer) {
 function get3DBuildingColor(style) {
   // set building color to be the same as the background color.
   if (!style.style) {
-    return DEFAULT_BLDG_COLOR;
+    return hexToRgb(DEFAULT_BLDG_COLOR);
   }
 
   const backgroundLayer = (style.style.layers || []).find(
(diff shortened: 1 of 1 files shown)


src/reducers/map-style-updaters.js
```diff
@@ -189,7 +189,7 @@ function findLayerFillColor(layer) {
 function get3DBuildingColor(style) {
   // set building color to be the same as the background color.
   if (!style.style) {
-    return DEFAULT_BLDG_COLOR;
+    return hexToRgb(DEFAULT_BLDG_COLOR);
```
heshan0131: hmm, maybe it's easier to save `DEFAULT_BLDG_COLOR` as a [r, g, b]?

You also need to convert it on line 206

jonsadka: I initially thought about that as well but `DEFAULT_BLDG_COLOR ` at line 206 is converted to rgb at line 212; seemed like it made more sense to leave it as a hex since this function is operation on other hex values

PR #876, opened 2019-12-24
[Enhancement] added coordinate to tooltip export configuration

### LINKED ISSUE #846: [Bug] Interactions: Coordinate setting is not saved to the Map Config
**Describe the bug**
The state of the Coordinate setting under Interactions is not saved in the Map Config.  The Tooltip and Brush settings are saved, but the Coordinate is not.

**To Reproduce**
1. Enable Coordinate under the Interactions settings
2. Select Share and Export Map
3. Select the JSON format
4. Look at the Map Config
5. Under "interactionConfig" the only saved settings are for "tooltip" and "brush".  There is nothing related to Coordinate.

Tested using Kepler.gl version 1.1.11 on https://kepler.gl/demo/

**Expected behavior**
The Coordinate setting should be saved to the Map Config like the other Interaction Config settings and that Coordinate setting should be likewise be applied when loading a map from a JSON Map Config.

**Related Issue**
#504

macrigiuseppe:
Fixes #846 


src/styles/base.js
```diff
@@ -206,7 +206,7 @@ export const modalPadding = '10px 0';
 export const modalLateralPadding = '72px';
 export const modalPortableLateralPadding = '36px';
 
-export const modalOverLayZ = 1000;
+export const modalOverLayZ = 1001;
```
macrigiuseppe: this makes sure the export modal dialog is fully visible when the tooltip is also active

src/reducers/vis-state-merger.js
```diff
@@ -213,13 +213,15 @@ export function mergeInteractions(state, interactionToBeMerged) {
...
         ...state.interactionConfig[key],
         enabled,
-        config: pick(
-          {
-            ...state.interactionConfig[key].config,
-            ...configToMerge
-          },
-          Object.keys(state.interactionConfig[key].config)
-        )
+        ...(state.interactionConfig[key].config ? {
```
macrigiuseppe: coordinate entry doesn't have a config property

src/tasks/tasks.js
```diff
@@ -24,7 +24,7 @@ import console from 'global/console';
 
 export const LOAD_FILE_TASK = taskCreator(
   ({fileBlob, info, handler, processor}, success, error) => {
-    console.time('loadFileTask')
+    console.time('loadFileTask');
```
heshan0131: delete

src/tasks/tasks.js
```diff
@@ -36,7 +36,7 @@ export const LOAD_FILE_TASK = taskCreator(
           // result has both datasets and info
           // TODO: I think we should pass info to the handler and return
           // the same format back from the file handler
-          console.timeEnd('loadFileTask')
+          console.timeEnd('loadFileTask');
```
heshan0131: delete

test/node/schemas/vis-state-schema-test.js
```diff
@@ -143,7 +143,8 @@ test('#visStateSchema -> v1 -> save load interaction', t => {
     brush: {
       enabled: false,
       size: 0.5
-    }
+    },
+    coordinate: { enabled: false }
```
heshan0131: remove space in brackats

src/reducers/vis-state-merger.js
heshan0131: can you add tests in vis-schema to test for
1. coordinates are saved and correctly loaded
2. older config without coordinates will result in loading error

macrigiuseppe: @heshan0131 are you sure old configuration should fail? Would it be better if we handle them gracefully and default to __coordinate enabled: false__?

macrigiuseppe: let's make sure old configs are not triggering loading errors

PR #883, opened 2020-01-03
License 2020

macrigiuseppe:


actions.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2020 Uber Technologies, Inc.
```
macrigiuseppe: @heshan0131 should we use __Uber Technologies__ still? or UCF?

heshan0131: yea use Uber

Issue #905, opened 2020-01-14
[Feat] Provide save map UI components in kepler.gl


heshan0131:
To provide UI components for save map to Backend storage. kepler.gl will add UI component needed to render:

1. Additional action item and dropdown in side panel header
2. A save map dialog to input map title, description and thumbnail

Backend storage feature tracking here:
https://github.com/keplergl/kepler.gl/issues/804

![Screen Shot 2020-01-14 at 11 35 34 AM](https://user-images.githubusercontent.com/3605556/72376292-90da4380-36c2-11ea-8fa6-f99262ddcd31.png)
![Screen Shot 2020-01-14 at 11 36 19 AM](https://user-images.githubusercontent.com/3605556/72376293-90da4380-36c2-11ea-80a9-7f2431b48233.png)

PR #906, opened 2020-01-14
[Feat] UI for save map to backend storage

heshan0131:
https://github.com/keplergl/kepler.gl/issues/905

Save map modal allows select provider, input of map title and description. it will also generate a thumbnail
Screen Shot 2020-02-07 at 12 13 49 PM

Changes to side panel header, add another cloud storage icon + dropdown
Screen Shot 2020-01-14 at 11 36 19 AM

Added share map url modal
share-dropbox-url


examples/demo-app/src/app.js
```diff
@@ -356,6 +356,7 @@ class App extends Component {
                   width={width}
                   height={height - (showBanner ? BannerHeight : 0)}
                   onSaveMap={this._isCloudStorageEnabled() && this._toggleCloudModal}
+                  onSaveToStorage={() => {}}
```
macrigiuseppe: what would be the difference between onSaveMap and onSaveToStorage?
It feels like they should perform the same operation, should we merge it?

we could use onSaveToStorage and pass it to cloud provider interaction

src/components/common/styled-components.js
```diff
@@ -368,6 +378,58 @@ export const StyledModalContent = styled.div`
...
+  }
+`;
+
+export const StyledModalSection = styled.div.attrs({
+  className: 'modal-section'
+})`
+  margin-bottom: 32px;
+
+  .modal-section-title {
+    font-weight: 500;
```
macrigiuseppe: ${props => props.theme.inputFontWeight}

src/components/common/styled-components.js
```diff
@@ -368,6 +378,58 @@ export const StyledModalContent = styled.div`
...
+  .modal-section-subtitle {
+    color: ${props => props.theme.subtextColorLT};
+  }
+
+  input {
+    margin-top: 8px;
+  }
+
+  ${media.portable`
+    margin-bottom: 24px;
```
macrigiuseppe: should create constant values?

src/components/common/styled-components.js
```diff
@@ -368,6 +378,58 @@ export const StyledModalContent = styled.div`
...
+  `};
+`;
+
+export const StyledModalInputFootnote = styled.div.attrs({
+  className: 'modal-input__footnote'
+})`
+  display: flex;
+  justify-content: flex-end;
+  color: ${props => props.error ? props.theme.errorColor : props.theme.subtextColorLT};
+  font-size: 10px;
```
macrigiuseppe: inputFontSizeSmall

src/components/index.js
```diff
@@ -73,7 +69,7 @@ export {default as LayerHoverInfoFactory} from './map/layer-hover-info';
 export {default as CoordinateInfoFactory} from './map/coordinate-info';
 
 // // modal container factories
-export {default as ModalDialoFactory} from './modals/modal-dialog';
+export {default as ModalDialogFactory} from './modals/modal-dialog';
```
macrigiuseppe: Jon fixed this already: https://github.com/keplergl/kepler.gl/pull/911

src/components/modal-container.js
```diff
@@ -134,99 +134,24 @@ export default function ModalContainerFactory(
...
       const {uiState} = this.props;
       const {format} = uiState.exportMap;
-
-      const downloader = {
-        [EXPORT_MAP_FORMATS.HTML]: this._onExportHTMLMap,
-        [EXPORT_MAP_FORMATS.JSON]: this._onExportJSONMap
-      }[format];
-
-      downloader && downloader();
+      (format === EXPORT_MAP_FORMATS.HTML ? exportHtml : exportJson)(
```
macrigiuseppe: if we add new formats this wouldn't scale

src/components/modals/save-map-modal.js
```diff
@@ -0,0 +1,144 @@
...
+        mapW: this.props.thumbWidth,
+        mapH: this.props.thumbHeight
+      });
+    }
+
+    _onChangeInput = (key, limit, e) => {
+      const {
+        target: {value}
+      } = e;
+      // if (value.length <= limit) {
```
macrigiuseppe: remove comments

src/components/modals/save-map-modal.js
```diff
@@ -0,0 +1,144 @@
...
+        <StyledSaveMapModal>
+          <StyledModalContent className="save-map-modal-content">
+            <StyledModalVerticalPanel>
+              <StyledModalSection className="save-map-modal-name">
+                <div className="modal-section-title">Name*</div>
+                <div>
+                  <InputLight
+                    type="text"
+                    value={mapInfo.title}
+                    onChange={e => this._onChangeInput('title', LIMIT, e)}
```
macrigiuseppe: do we need to update map info for every single character, can't we simply pass all info when we submit?

examples/demo-app/src/app.js
heshan0131: for backward compatibility. `onSaveMap` is currently called to share map with URL (by dropbox), `onSaveToStorage` is the newly added API which could be called by Carto - provider with a storage option. It's important to keep both, and later rename `onSaveMap` to `onShareMapURL` so that apps using the onSaveMap won't bread. Also we provide 2 methods to save / share, with an URL or directly opload assets to a storage.

src/utils/export-utils.js
```diff
@@ -0,0 +1,212 @@
...
+export const DEFAULT_JSON_NAME = 'keplergl.json';
+export const DEFAULT_DATA_NAME = 'kepler-gl';
+
+const defaultResolution = RESOLUTION_OPTIONS.find(
+  op => op.id === RESOLUTIONS.ONE_X
+);
+
+const defaultRatio = RATIO_OPTIONS.find(op => op.id === RATIOS.FOUR_BY_THREE);
+
+export function calculateExportImageSize({mapW, mapH, ratio, resolution}) {
```
heshan0131: @jonsadka

src/components/modals/export-image-modal.js
```diff
@@ -21,20 +21,15 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import ImagePreview from 'components/common/image-preview';
 
-import {calculateExportImageSize} from 'utils/export-image-utils';
```
heshan0131: @jonsadka

heshan0131:
`KeplerGl` component now takes in 3 addition props:

- `cloudProviders <Array>` array of providers
- `onExportToCloudSuccess <Function>` additional action to invoke upon upload file to cloud success, here you can call actions such as push to location url. 
- `onExportToCloudError <Function>` additional action to invoke when upload file to cloud errored

Cloud Provider Methods:
`hasPrivateStorage()`:  returns `true` or `false` - whether private storage is supported. if return true, tile will be shown in save map modal
`hasSharingUrl()`:  returns `true` or `false` - whether share map via url is enabled, if return true, tile will be shown in share map url modal 
`uploadFile({mapData, blob, fileName, isPublic})`: invokes when share map url or save map triggered

- **`mapData`** - json represencation of datasets, config and map info 
```js
map: {
  datasets: [{…}]
  config: {version: "v1", config: {…}},
  info: {
    app: "kepler.gl", 
    created_at: "Wed Feb 05 2020 15:27:06 GMT-0800 (Pacific Standard Time)"
  }
},
info: {
  title: ""
  description: ""
  id: ""
},
thumbnail: <png file blob>
```

- **blob** - json file blob, used to directly upload to file-based cloud system (dropbox or google drive)

- **fileName** - Auto generated, or from user input

- **isPublic** - privacy setting when storing map

``

src/actions/provider-actions.js
```diff
@@ -0,0 +1,66 @@
...
+);
+
+export const exportFileError = createAction(
+  ActionTypes.EXPORT_FILE_ERROR,
+  ({error, provider, onError}) => ({error, provider, onError})
+);
+
+export const saveToCloudSuccess = createAction(
+  ActionTypes.SAVE_TO_CLOUD_SUCCESS
+)
```
macrigiuseppe: missing __;__

src/tasks/tasks.js
```diff
@@ -67,3 +67,28 @@ export const LOAD_MAP_STYLE_TASK = taskCreator(
...
+
+/**
+ * task to upload file to cloud provider
+ */
+export const EXPORT_FILE_TO_CLOUD_TASK = Task.fromPromise(
+
+  ({provider, payload}) => provider.uploadFile(payload),
+
+  'EXPORT_FILE_TO_CLOUD_TASK'
+)
```
macrigiuseppe: missing __;__

src/tasks/tasks.js
```diff
@@ -67,3 +67,28 @@ export const LOAD_MAP_STYLE_TASK = taskCreator(
...
+
+/**
+ *  task to dispatch a function as a task
+ */
+export const ACTION_TASK = Task.fromCallback(
+  (_, cb) => cb(),
+
+  'ACTION_TASK'
+)
```
macrigiuseppe: missing __;__

src/reducers/provider-state-updaters.js
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * This method will export the current kepler config file to the chosen cloud proder
+ * add returns a share URL
+ *
+ * @param {*} state
+ * @param {*} action
+ */
+export const exportFileToCloudUpdater = (state, action) => {
+  const {mapData, provider, isPublic, onSuccess, onError, closeModal} = action.payload;
+  console.log(mapData);
```
macrigiuseppe: remove this

src/components/modal-container.js
```diff
@@ -134,101 +145,54 @@ export default function ModalContainerFactory(
...
-      const {uiState} = this.props;
-      const {userMapboxToken, exportMapboxAccessToken, mode} = uiState.exportMap[EXPORT_MAP_FORMATS.HTML];
+    _exportFileToCloud = ({provider, isPublic, closeModal}) => {
+      const toSave = exportMap(this.props);
+      this.props.providerActions.exportFileToCloud({
+        mapData: toSave,
+        provider,
+        isPublic,
+        closeModal,
+        onSuccess: this.props.onExportToCloudSuccess,
```
macrigiuseppe: we are using these two callbacks (onSuccess, onError) from within redux using trunk.

Is this a better pattern than using redux to set attributes in the store and upon detecting new values in the view we call onSuccess or onError?

src/components/map/map-control.js
```diff
@@ -261,26 +261,30 @@ const MapDrawPanel = React.memo(({
           <ToolbarItem
             onClick={() => onSetEditorMode(EDITOR_MODES.EDIT)}
             label="select"
-            icon={(<CursorClick height="22px"/>)}
+            iconHeight="22px"
```
macrigiuseppe: are you using this value in toolbarItem?

macrigiuseppe: unless we need the height, i like the previous version where we just pass the icon. More concise

src/components/modal-container.js
```diff
@@ -78,6 +79,10 @@ const LoadDataModalStyle = css`
   top: 60px;
 `;
 
+const DefaultStyle = css`
+  max-width: 960px;
```
macrigiuseppe: theme value?

src/components/modals/cloud-tile.js
```diff
@@ -0,0 +1,182 @@
...
+  // cloud provider class
+  cloudProvider,
+  // function to take after login or logout
+  onSetCloudProvider,
+  // whether provider is selected as currentProvider
+  isSelected,
+  // whether user has logged in
+  isConnected
+}) => {
+  const userName =
```
macrigiuseppe: we can improve performance by avoiding recomputing for every rendering

```
import {useMemo} from 'react';
const userName = useMemo(() =>  
  typeof cloudProvider.getUserName === 'function'
      ? cloudProvider.getUserName()
      : null
, [cloudProvider.getUserName]);
```

src/components/modals/cloud-tile.js
```diff
@@ -0,0 +1,182 @@
...
+    typeof cloudProvider.getUserName === 'function'
+      ? cloudProvider.getUserName()
+      : null;
+  const onClickConnect =
```
macrigiuseppe: similar thing here:

```
import {useCallback} from 'react';

const onClickConnect = useCallback(() => typeof onConnect === 'function'
      ? onConnect
      : () => cloudProvider.login(() => onSetCloudProvider(cloudProvider.name))
,[onConnect, cloudProvider.login, cloudProvider.name, onSetCloudProvider]);

```

src/components/modals/cloud-tile.js
```diff
@@ -0,0 +1,182 @@
...
+    typeof onConnect === 'function'
+      ? onConnect
+      : () => cloudProvider.login(() => onSetCloudProvider(cloudProvider.name));
+
+  const onClickLogout =
```
macrigiuseppe: same here

src/constants/action-types.js
```diff
@@ -127,9 +128,10 @@ const ActionTypes = keyMirror({
   REMOVE_NOTIFICATION: null,
 
   // uiState > export image
-  SET_RATIO: null,
-  SET_RESOLUTION: null,
-  TOGGLE_LEGEND: null,
+  // SET_RATIO: null,
+  // SET_RESOLUTION: null,
+  // TOGGLE_LEGEND: null,
```
macrigiuseppe: is this correct?

src/reducers/provider-state-updaters.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  if (typeof provider.uploadFile !== 'function') {
+    Console.error(
+      `uploadFile is not a function of Cloud provider: ${provider.name}`
+    );
+    return state;
+  }
+
+  // TODO: do we need to always create a fileBlob?
```
macrigiuseppe: I double checked the API (https://dropbox.github.io/dropbox-sdk-js/Dropbox.html#filesUpload) and based on the type accepted by the API we may be able to pass any object: https://dropbox.github.io/dropbox-sdk-js/global.html#FilesCommitInfo

macrigiuseppe: I agree the provider should take care of handling the json the way it wants

src/reducers/provider-state-updaters.js
```diff
@@ -0,0 +1,200 @@
...
+ * @param {*} state
+ * @param {*} action
+ */
+export const exportFileSuccessUpdater = (state, action) => {
+  const {response, provider, onSuccess, closeModal} = action.payload;
+
+  const newState = {
+    ...state,
+    isLoading: false,
+    // TODO: do we always have to store this?
```
macrigiuseppe: i think we should store this since each provider may want to display information upon a successful operation. Since data may be dynamic we may have to define a set of standard properties each provider should provider upon operation completion

src/reducers/provider-state-updaters.js
```diff
@@ -0,0 +1,200 @@
...
+      folderLink: response.folder_link
+    }
+  };
+
+  const tasks = [
+    createActionTask(onSuccess, {response, provider}),
+    closeModal && ACTION_TASK().map(_ => saveToCloudSuccess())
+  ].filter(d => d);
+
+  return tasks.length ? withTask(newState, tasks) : newState;
```
macrigiuseppe: doesn't __tasks__ always have at least on element due to __createActionTask(onSuccess, {response, provider})__?

src/reducers/ui-state.js
```diff
@@ -36,9 +36,7 @@ const actionHandler = {
   [ActionTypes.ADD_NOTIFICATION]: uiStateUpdaters.addNotificationUpdater,
   [ActionTypes.REMOVE_NOTIFICATION]: uiStateUpdaters.removeNotificationUpdater,
 
-  [ActionTypes.SET_RATIO]: uiStateUpdaters.setRatioUpdater,
-  [ActionTypes.SET_RESOLUTION]: uiStateUpdaters.setResolutionUpdater,
-  [ActionTypes.TOGGLE_LEGEND]: uiStateUpdaters.toggleLegendUpdater,
```
macrigiuseppe: why removing toggle legend?

src/utils/map-info-utils.js
```diff
@@ -18,17 +18,21 @@
...
 
-import {SaveMapFactory} from 'kepler.gl/components';
+import {MAP_INFO_CHARACTER} from 'constants/default-settings';
 
-export const CustomSaveMapFactory = () => {
-  const SaveMap = SaveMapFactory();
-  SaveMap.defaultProps = {
-    ...SaveMap.defaultProps,
-    label: 'Share Public URL'
+export function getDefaultMapInfo() {
```
macrigiuseppe: this is a simple constant

src/components/modal-container.js
heshan0131: `onSuccess` and `onError` are action handlers. kepler.gl core uses tasks, instead of thunk. it creates a task that dispatches the custom actions if user-defined

src/components/modals/cloud-tile.js
heshan0131: we need to specifically recomputing because cloudProvider is a class stored in the store, changing it will not cause react re render

src/components/modals/cloud-tile.js
heshan0131: why makes it hard to read, while the performance improvement here is so minimum

src/constants/action-types.js
heshan0131: these actions are deleted, all handled by one: `SET_EXPORT_IMAGE_SETTING`

src/reducers/provider-state-updaters.js
heshan0131: `createActionTask` should return null, if `onSuccess` is not defined

src/reducers/ui-state.js
heshan0131: because all these 3 actions are behaving the same and will trigger recalculation of imageSize. they are merged into 1 below

src/tasks/tasks.js
heshan0131: fixed

src/constants/action-types.js
macrigiuseppe: what about TOGGLE_LEGEND?

src/components/modals/cloud-tile.js
macrigiuseppe: because we are calling getUserName every time this component renders, we could memoize its value by using useMemo

src/components/modal-container.js
macrigiuseppe: got it.

src/components/modals/cloud-tile.js
heshan0131: `useMemo will only recompute the memoized value when one of the dependencies has changed`. and if you use the `getUserName` function as a dependency, the functino itself will never change. this won't work.

Also `username` is a string and most likely a constant stored in provider. when `getUsername` gets called every time, it will return the same value.

macrigiuseppe: I can see the problem when username is populated after the provider class is initiated,e.g. async all to get username information.
Otherwise, if both getUserName and the actual value of the username were to be initiated at the same time and never change, e.g. in case of a constant, this would still work.
We need to make sure we support the async scenario as well, therefore, you are right on not using useMemo

src/components/modals/cloud-tile.js
macrigiuseppe: it will avoid StyledTileWrapper to re-render everytime we render cloud tile

src/components/modals/cloud-tile.js
heshan0131: right, username is only assigned when the user successfully logged in

PR #922, opened 2020-01-22
CARTO Provider for Backend Storage

neokore:
This feature allows loading/saving private and public maps using CARTO.

Related to https://github.com/keplergl/kepler.gl/issues/804


heshan0131:
I am testing out the branch locally. Currently seeing issue:
1. Connection not secured when run `npm start`, due to innitialze carto cloud provider on app start
2. If user doesn't provide a CartoClientId, the app won't start. this is not acceptable. We won't be able to provide a client id for all kepler developers to run the demo app locally

neokore:
Sorry @heshan0131, the second issue was fixed earlier, but maybe it got lost in the refactor for the new UI, I'll fix it asap.
About the secure connection, using a dummy HTTPS connection just for developing doesn't have to cause any problem and right now it is needed for OAuth, but if you want us to change this behavior, just let us know what method do you prefer.

neokore:
@heshan0131. I've found that the crash was produced by an uncaught exception, now it is fixed and works without setting a clientId. Also I've added a check to show only properly initialized providers.

neokore:
I've fixed all conflicts with this branch and added a HTTPS entry for development called `start:https` for OAuth providers that need it.

examples/demo-app/src/actions.js
```diff
@@ -365,27 +416,121 @@ export function exportFileToCloud(providerName) {
...
+          const responseUrl = cloudProvider.getMapPermalink
+            ? cloudProvider.getMapPermalink(response.url, false)
+            : getMapPermalink(response.url, false)
+          dispatch(push(responseUrl));
         }
-        dispatch(setPushingFile(false, {filename: file.name, status: 'success', metadata: response}));
+        dispatch(setPushingFile(false, {
+          filename: file.name,
+          status: 'success',
+          metadata: response,
```
heshan0131: Is it necessary for kepler to store the entire response here? looks like kepler only needs the `shareUrl` and `folderLink` to display. And it sends the url back to `provider.getMapPermalink`. Why don't the provider store `shareUrl` and `folderLink` when request succeed, and kepler only call `provider.getMapPermalink()` to render it. 

If this makes sense. I will implement it in my pr

examples/demo-app/src/app.js
```diff
@@ -278,14 +294,38 @@ class App extends Component {
     });
   };
 
+  _hasBackendProviderActive = () => {
```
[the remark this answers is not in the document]
neokore: You're right. I've used this method before to check the selected backend provider status before showing the save modal. If this method returned `false` it showed the provider selector modal. Now it cannot be used from `app.js` so feel free to move it to the core or simply remove it.

examples/demo-app/src/actions.js
neokore: This reponse object only has a `url` field from CARTO provider (same as `shareUrl`) but I kept the entire response in case of other providers need it.
Anyway, I agree with your proposal, we already save last saved/loaded visualization and offer a method called `getCurrentVisualization` to get it, but also offer a `getMapPermalinkFromParams` method to be able to create a permalink to any vis from the load map modal or from query params.

heshan0131:
Trying to login my carto account to test saving. got this screen:
Screen Shot 2020-02-05 at 11 27 24 AM

The url in the popup window is 
`https://kepler.gl/demo#error=invalid_request&error_description=The+redirect_uri+must+match+the+redirect_uri+param+used+in+the+authorization+request&state`

Is there anything else I need to setup in my carto account to test this feature?

neokore:
Hi @heshan0131! Probably you just need to test it using HTTPS server. There is already a `start:https` script in the project's package.json file.

heshan0131:
@neokore 
Another PR created for loading maps with provider.loadMaps function, supported by kepler core.
https://github.com/keplergl/kepler.gl/pull/947

I used the dropbox provider as demo how `provider.loadMap` function `@params` and `@return`

examples/demo-app/src/cloud-providers/carto-provider.js
```diff
@@ -0,0 +1,352 @@
...
+        if (currentUsername && currentUsername===username) {
+          const cs = await this._carto.getCustomStorage();
+          visualization = await cs.getVisualization(mapId);
+        }
+      } else {
+        visualization = await this._carto.PublicStorageReader.getVisualization(username, mapId);
+      }
+
+      if (!visualization) {
+        throw {target: {status: 404, responseText: `Can't find map with ID: ${mapId}`}};
```
heshan0131: this error will be caught on line 231, and pass to `_maangeErrors`, in order to get the correct message, this should be throw ```new Error(`Can't find map with ID: ${mapId}`)```

src/components/common/icons/database.js
```diff
@@ -0,0 +1,50 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import Base from './base';
+
+export default class Database extends Component {
```
heshan0131: seems like this file is not used anymore

examples/demo-app/src/utils/routes.js
```diff
@@ -44,6 +44,7 @@ export function buildAppRoutes(Component) {
       <Route key="demo" path="demo">
         <IndexRoute component={Component} />
         <Route path="map" component={Component} />
+        <Route path="map/:provider" component={Component} />
```
heshan0131: Duplicate routes here

neokore:
We've simplified the PR in another branch here: https://github.com/keplergl/kepler.gl/pull/985

PR #933, opened 2020-01-30
[Bug] fix hexagon layer hover cause app crash

heshan0131:


src/components/map/layer-hover-info.js
```diff
@@ -115,14 +115,14 @@ const CellInfo = ({data, layer}) => {
 
 const LayerHoverInfoFactory = () => {
   const LayerHoverInfo = props => {
-    const {data, layer, fieldsToShow} = props;
+    const {data, layer} = props;
 
-    if (!data || !layer || !fieldsToShow.length) {
```
macrigiuseppe: was this the reason of the crash?

Issue #929, opened 2020-01-29
heat map feature not working 


aloncohen1:
Heat maps can't be generated at all. all other features seems to work. (for example points, grid, etc)
Attached is the data sample I tried to load for the heat map.

[data.csv.zip](https://github.com/keplergl/kepler.gl/files/4126950/data.csv.zip)

macrigiuseppe:
fixed

PR #934, opened 2020-01-30
[Bug] fix heatmap not displayed when there is no filter

heshan0131:
- do not assign mapbox style filter when there is none


heshan0131:
Fixed https://github.com/keplergl/kepler.gl/issues/929
Screen Shot 2020-01-30 at 12 14 55 PM

Issue #834, opened 2019-11-17
Create e2e tests


macrigiuseppe:
**Is your feature request related to a problem? 
Currently we have no way to test certain scenarios in the app (dropbox integration, download maps, etc..)

**Describe the solution you'd like**
Implements an e2e test strategy

**Describe alternatives you've considered**
jest + puppeteer

PR #940, opened 2020-02-06
[Test] e2e test

macrigiuseppe:
This PR implements #834


contributing/DEVELOPERS.md
```diff
@@ -239,35 +239,6 @@ To checkout the build
     cd website && yarn build
 ```
 
-Publish on github pages __Authorized User Only__.
```
heshan0131: Worth mention github action will auto public

heshan0131:
I have a PR out to run prettier on {examples,src,test} with the max-len 100, so yours shouldn't need to include all those changes

contributing/DEVELOPERS.md
macrigiuseppe: Yeah, I will add and netlify deployment to kepler.gl website

macrigiuseppe:
> I have a PR out to run prettier on {examples,src,test} with the max-len 100, so yours shouldn't need to include all those changes
> 
> #968

yeah i will rebase once your is merged

src/components/editor/feature-action-panel.js
```diff
@@ -76,7 +76,7 @@ export class FeatureActionPanel extends PureComponent {
         }}
       >
         <ActionPanel>
-          <ActionPanelItem label="layers" Icon={Layers}>
+          <ActionPanelItem className="editor-layers-list" label="layers" Icon={Layers}>
```
heshan0131: when passing classnames to styled componenet you can use the `attrs` method

```js
const StyledItem = styled.div.attrs(props => ({
  className: classnames('action-items', props.className)
}))`
  display: flex;
  flex-direction: row;
`
```

src/components/modals/export-data-modal.js
```diff
@@ -126,20 +126,22 @@ const ExportDataModalFactory = () => {
               </div>
               <div className="selection">
                 <StyledFilteredOption
+                  className="unfiltered-option"
                   selected={!filtered}
                   onClick={() => onChangeExportFiltered(false)}
                 >
-                  <div className="filtered-title">Unfiltered Data</div>
-                  <div className="filtered-subtitle">
+                  <div className="filter-option-title">Unfiltered Data</div>
```
heshan0131: is there a reason why we changed the className here? It might create issues for people who are using global css selector to override default style of thses dom element

src/components/editor/feature-action-panel.js
macrigiuseppe: this is a good approach when I want to have a base class name for a given styled component.

I think in this case, since i don't need a base CSS class, styled-components already take care of things automatically by simply passing className property.

src/components/modals/export-data-modal.js
macrigiuseppe: I changed to have a more meaningful name

Issue #768, opened 2019-10-04
Allow users to select multiple datasets per filter


macrigiuseppe:
The current UI only allows users to select one dataset per filter.

Modify the UI to enable multiple datasets per each filter.

Related
#146

luckysuman03:
Hello, any updates on this?

macrigiuseppe:
@luckysuman03 this feature is currently on hold. I will post updates as soon as we resume working on it

vonum:
Hello, any updates on this?

calvinbrewer:
@macrigiuseppe This issue references the UI's inability to select two data sources for a single filter but are you able to configure this via the API?

ramilrgk:
Hello, any updates on this?

DeuceDawg22:
Are there any updates on this? Would be a huge benefit to have this capability instead of needing to find a different tool to use.

ivashi:
Any updates on this? Would be very useful for my use case.

PR #954, opened 2020-02-18
[WIP] [Feat] Multidataset filters

macrigiuseppe:
This implements #768 

### DIFF
--- a/docs/api-reference/get-started.md
+++ b/docs/api-reference/get-started.md
@@ -79,9 +79,9 @@ this.props.dispatch(
       readOnly: false
     },
     // config
-    config: {
+    `config: {
       mapStyle: {styleType: 'light'}
-    }
+    }`
   })
 );
 ```
--- a/examples/demo-app/src/store.js
+++ b/examples/demo-app/src/store.js
@@ -45,7 +45,6 @@ let composeEnhancers = compose;
 /**
  * comment out code below to enable Redux Devtools
  */
-/*
 if (window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) {
   composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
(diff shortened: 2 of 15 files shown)


scivm:
How is this change going?  Was trying to use it to model different medical codes each in own csv in time series.   Was using the branch version but did not see any difference.   Selecting "+ Add data source" did not do anything.  This will be a good enhancement.

scivm:
kepler

macrigiuseppe:
@scivm this feature is currently a WIP (work in progress) and updates will be posted as soon as they are ready

monikasaran:
Hi @macrigiuseppe ,
This is a really cool feature and will be a great help for me. Is this PR ready to use?  Can you please share the update on this? 
Thank you so much in advance.

monikasaran:
@macrigiuseppe , If you can help in anyway it will be really appreciated.
I will try to explain my problem as simple as possible.

1. I am working to visualise supply chain on kepler gl.
2. The entities involved: Store address, customer address, Van trip routes
3. One store can deliver packages to multiple customers. Multiple Van can pick the packages from single store.
4. Use case is to be able to filter based on store numbers along with the ability to show details of each and every entity(customer address details and route details) on hover tooltip.

Approaches till now:
- I have tried to populate the data in a single dataset to make the filter work, but the render explodes because it's millions of records we are talking about.
- If I try to reduce the record by making collection of customer address against a single store then details related to the customer(hover tooltip) is impossible to show.
- If I create different datasets for all three, then it is impossible to filter across datasets where store acts as a primary key.

I am in a very rough position now. Let me know if any kind of clarification is required.
Thanks in advance.

macrigiuseppe:
@monikasaran this feature is currently on hold. Before moving to multi dataset filter we are going to restructure filters in order to be more extensible. Unfortunately we don't have an timeline for the new feature

PR #956, opened 2020-02-19
[Bug] Fixed json map export and added tests

### LINKED ISSUE #953: [Bug] Export map JSON provides empty output
**Describe the bug**
When trying to export a JSON output from the Kepler demo, I get an empty JSON output. 

**To Reproduce**
1. Add data to map (GeoJSON in my case)
2. Click on export map -- > JSON. Already now you see that the JSON preview doesn't follow your map. e.g. the color palette remains the same. 
3. Click export and JSON config is downloaded
4. See error from the file, that only has contents saying "[object Object]"

**Expected behavior**
Config should contain the contents of my styled map and it should work as described: "Export current map data and config into a single Json file. You can later open the same map by uploading this file to kepler.gl."

**Screenshots**
For example here you can see that the map config doesn't follow what I have on the map
image

When doing the export from here, I get an empty JSON. 

 - OS: Windows 
 - Browser Chrome
 - Version Chrome Version 79.0.3945.130 (Official Build) (64-bit)

**Additional context**
I have been previously able to export configs (just last week) without this issue.

macrigiuseppe:
This PR fixes #953 

### DIFF
--- a/src/utils/export-utils.js
+++ b/src/utils/export-utils.js
@@ -134,22 +134,26 @@ export function downloadFile(fileBlob, filename) {
   URL.revokeObjectURL(url);
 }
 
-export function exportImage(state, options) {
-  const {imageDataUri} = state.uiState.exportImage
+export function exportImage(state) {
+  const {imageDataUri} = state.uiState.exportImage;
   if (imageDataUri) {
     const file = dataURItoBlob(imageDataUri);
     downloadFile(file, DEFAULT_IMAGE_NAME);
   }
 }
 
-export function exportJson(state, options) {
+export function exportMapToJson(state, options = {}) {
   const {hasData} = options;
 
   const data = hasData
     ? KeplerGlSchema.save(state)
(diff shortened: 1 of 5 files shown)


src/utils/export-utils.js
```diff
@@ -134,22 +134,26 @@ export function downloadFile(fileBlob, filename) {
...
-export function exportJson(state, options) {
+export function exportMapToJson(state, options = {}) {
   const {hasData} = options;
 
   const data = hasData
     ? KeplerGlSchema.save(state)
     : KeplerGlSchema.getConfigToSave(state);
 
-  const fileBlob = new Blob([data], {type: 'application/json'});
+  return JSON.stringify(data, null, 2);
```
heshan0131: if the name of this function is `exportMapToJson` would it make more sense to return the `json` object before stringy it?

And rename the following function to `exportJsonFile` and run stringify in there?

PR #958, opened 2020-02-19
Decrease filter step size for small domains

isaacbrodsky:
Otherwise the detected filter step size may be larger than the domain,
which results in not being able to use the filter. This is applicable
when working with small percentages (<0.1%).

While changing this I also noticed this will treat negative numbers as
very small; I changed `getNumericStepSize` to compute the step size
based on absolute value. (Not sure the UI fully supports selecting negative
filter ranges, seemed like it wouldn't let me select them via mouse.)


src/utils/filter-utils.js
```diff
@@ -698,13 +698,23 @@ export function getNumericFieldDomain(data, valueAccessor) {
...
-  } else if (diff < 20 && diff > 3) {
+  } else if (diff > 3) {
     return 0.01;
-  } else if (diff <= 3) {
+  } else if (diff > 1) {
     return 0.001;
+  } else if (diff <= 1) {
+    // Try to get at least 100 steps - and keep the step size below that of
+    // the (diff > 1) case.
+    const x = diff / 100;
```
heshan0131: can you make this min step to be 1000?

isaacbrodsky: Sure, updated. I also updated tests based on the values I saw

isaacbrodsky:
From the Travis CI runs, it looks like the results are not consistent between Node 8 and Node 10. Is there an assertion utility for asserting "floating point equal within some epsilon"?

heshan0131:
you can try use `import {preciseRound} from 'utils/data-utils'` to compare the result?

isaacbrodsky:
Seems the tests are passing now although I didn't update them to use `preciseRound` (Would have needed to be change in `cmpFilters` in `comparison-utils`, potentially among others)

Issue #88, opened 2018-06-01
P0: Allow to import the exported map config


ziyili66:
That would also open up the ability to reuse styles across projects or even on multiple datasets in the same project. You could possibly even allow for multiple datasets to be in a single layer with a single style (assuming their data structure is consistent)

1. Save map config via save map choose config only, and downlaod the config json,  and store somewhere that can be accessed via url

2. Load kepler.gl/#/demo?mapConfig=http://xxxxx.xxxx.json. Load kepler.gl with a preset config (layers, filters, mapStyles etc). 

3.1 The config might contain 1 dataset, in which case kepler.gl try to map layer, filter to it
3.2 The config contain 2 dataset...........

Rules to match config to data: dataset must contain field that layer column and layer color / size by are depends on. must contain fields that filters are depends on....

Currently
if you call `receiveMapconfig` without data preloaded, kepler will try to match layers / filters based on `dataId`, it is not finding those fieds, then it will saved the layer/filter to `layersToBeMerged` `filtersToBeMerged`. When you then call `addDataToMap` without config. kepler will use the to be merged the ones.

For Example:

- If someone saved layer 1 -> dataset a; layer 2 -> dataset b
- upload dataset a -> kepler create layer 1. layer 2 will not be match, it will be stored as `layersToBeMerged` in visState reducer

PR #986, opened 2020-03-04
[Enhancement] Refactor load file tasks to better handle multiple file types

heshan0131:
- when load multiple files into file uploader. we do: 1.apply keplergl.json contain data  2. apply other files contain only data


heshan0131:
Implement this change allows us to easily handle apply just the config to map in the future

heshan0131:
#88

src/reducers/combined-updaters.js
```diff
@@ -78,10 +81,47 @@ const combinedUpdaters = null;
...
   isPlainObject(config) && isPlainObject(config.config) && config.version;
+
 export const defaultAddDataToMapOptions = {
   centerMap: true,
   keepExistingConfig: false
 };
+
+const identity = state => state;
+
+/* eslint-disable-next-line no-unused-vars */
```
macrigiuseppe: is this necessary?

src/reducers/combined-updaters.js
```diff
@@ -78,10 +81,47 @@ const combinedUpdaters = null;
...
+function log(text) {
+  return value => Console.log(text, value);
+}
+
+function payload_(p) {
```
macrigiuseppe: why the "_" at the end?

heshan0131: to make them consistent, because some of these functions have similar names as lodash, if we import pick from 'lodash.pick' we would have conflict with our own pick_ function

src/reducers/combined-updaters.js
heshan0131: log is a useful function for debugging

macrigiuseppe: no I meant if the  __/* eslint-disable-next-line no-unused-vars */__ line is necessary

heshan0131: yes, because log function is currently not used, it's for debugging

src/reducers/combined-updaters.js
```diff
@@ -78,10 +81,47 @@ const combinedUpdaters = null;
...
+  return {payload: p};
+}
+
+function apply_(updater, payload) {
```
macrigiuseppe: are planning to extend the usage of this technique to the rest of reducers?

heshan0131: yea sure. right now for a less disruptive change, I only used in combined updaters, it is helpful for reaching into subreducers and chain updater functions. But a huge gain for individual reducers at the moment

Issue #970, opened 2020-02-22
Kepler.gl Localization feature


macrigiuseppe:
We are currently looking for someone who would love to work with Kepler.gl code base and implement the new localization feature.

Going through all UI components, extract all embedded text out and create a single js file that can be used with [react-intl](https://github.com/formatjs/react-intl)

[Example on how to use react-intl](https://www.codeandweb.com/babeledit/tutorials/how-to-translate-your-react-app-with-react-intl)

Please let us know if you want to collaborate on this new and popular feature.

Joonalai:
I would like to help with this. I actually already posted a similar issue #967.

macrigiuseppe:
> #967

@Joonalai thank you for your interest. If you are interested I will assign this issue to you and start working on it.

Joonalai:
@macrigiuseppe that sounds great.

mayteio:
Give me a yell if you’d like to divide and conquer, I’d be happy to contribute too.

Joonalai:
Thanks @mayteio! I am almost done with the task but I'll give you a yell if I find something you could help with.

Joonalai:
@mayteio, would you perhaps have the time to check out the failing [browser tests](https://travis-ci.com/keplergl/kepler.gl/jobs/296469055)? I can check those later but maybe not during this week.

macrigiuseppe:
> @mayteio, would you perhaps have the time to check out the failing [browser tests](https://travis-ci.com/keplergl/kepler.gl/jobs/296469055)? I can check those later but maybe not during this week.

your error is the following: 

```
✖ Show not fail without props
    ------------------------------
      operator: throws
      expected: |-
        undefined
      actual: |-
        [Error: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.]
      at: Test.<anonymous> (/home/travis/build/keplergl/kepler.gl/test/browser/components/modals/save-map-modal-test.js:39:5)
      stack: |-
```

It means you need to inject IntlProvider in your tests or find a better way to isolate components from your localization implementation

Joonalai:
@macrigiuseppe, @mayteio I just fixed the tests related to localization.

mayteio:
Sorry guys, am on holiday without a laptop! Glad you got it sorted 👏

heshan0131:
@mayteio @Joonalai  I added some followup questions regarding future works, maybe this is something you can help on.

Joonalai:
[...]

PR #994, opened 2020-03-10
[Feat] Localization

Joonalai:
This PR implements #970.

In addition to extracting the strings to localization files, Finnish translation was provided.  
Localization could be changed by giving locale value to `uiState`:
```js
import {combineReducers} from 'redux';
import keplerGlReducer from 'kepler.gl/reducers';
import {LOCALES} from 'kepler.gl/constants';

const customizedKeplerGlReducer = keplerGlReducer
  .initialState({
    uiState: {
      // use Finnish locale
      locale: LOCALES.fi
    }
  });

const reducers = combineReducers({
 keplerGl: customizedKeplerGlReducer,
 app: appReducer
});
```
Also initial UI component for changing locale is provided, but hidden by default: `mapControls.mapLocale`.

### DIFF
--- a/src/components/map/map-control.js
+++ b/src/components/map/map-control.js
@@ -18,33 +18,34 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React, {Component} from 'react';
+import React, {Component, useCallback} from 'react';
 import PropTypes from 'prop-types';
 import {createSelector} from 'reselect';
 import styled from 'styled-components';
+import {FormattedMessage} from 'react-intl';
 import classnames from 'classnames';
 
-import {Tooltip, IconRoundSmall, MapControlButton} from 'components/common/styled-components';
+import {IconRoundSmall, MapControlButton, Tooltip} from 'components/common/styled-components';
 import MapLayerSelector from 'components/common/map-layer-selector';
 import KeplerGlLogo from 'components/common/logo';
 import MapLegend from './map-legend';
 import {
   Close,
-  Split,
-  Legend,
   Cube3d,
+  CursorClick,
   Delete,
-  Layers,
   DrawPolygon,
+  EyeSeen,
+  EyeUnseen,
+  Layers,
+  Legend,
   Polygon,
   Rectangle,
-  CursorClick,
-  EyeSeen,
-  EyeUnseen
+  Split
 } from 'components/common/icons';
 import VerticalToolbar from 'components/common/vertical-toolbar';
 import ToolbarItem from 'components/common/toolbar-item';
-import {EDITOR_MODES} from 'constants/default-settings';
+import {EDITOR_MODES, LOCALES} from 'constants/default-settings';
 
 const StyledMapControl = styled.div`
   right: 0;
@@ -101,7 +102,9 @@ ActionPanel.displayName = 'ActionPanel';
 
 const MapControlTooltip = React.memo(({id, message}) => (
   <Tooltip id={id} place="left" effect="solid">
-    <span>{message}</span>
+    <span>
+      <FormattedMessage id={message} />
+    </span>
   </Tooltip>
 ));
 
@@ -109,7 +112,9 @@ MapControlTooltip.displayName = 'MapControlTooltip';
 
 const MapLegendTooltip = ({id, message}) => (
   <Tooltip id={id} place="left" effect="solid">
-    <span>{message}</span>
+    <span>
+      <FormattedMessage id={message} />
+    </span>
   </Tooltip>
 );
 
@@ -128,11 +133,11 @@ const LayerSelectorPanel = React.memo(({items, onMapToggleLayer, isActive, toggl
       <Layers height="22px" />
       <MapControlTooltip
         id="toggle-layer"
-        message={isActive ? 'Hide layer panel' : 'Show layer panel'}
+        message={isActive ? 'tooltip.hideLayerPanel' : 'tooltip.showLayerPanel'}
       />
     </MapControlButton>
   ) : (
-    <MapControlPanel header="Visible layers" onClick={toggleMenuPanel}>
+    <MapControlPanel header="header.visibleLayers" onClick={toggleMenuPanel}>
       <MapLayerSelector layers={items} onMapToggleLayer={onMapToggleLayer} />
     </MapControlPanel>
   )
@@ -152,7 +157,9 @@ const MapControlPanel = React.memo(({children, header, onClick, scale = 1, isExp
       {isExport ? (
         <KeplerGlLogo version={false} appName="kepler.gl" />
       ) : (
-        <span style={{verticalAlign: 'middle'}}>{header}</span>
+        <span style={{verticalAlign: 'middle'}}>
+          <FormattedMessage id={header} />
+        </span>
       )}
       {isExport ? null : (
         <IconRoundSmall className="close-map-control-item" onClick={onClick}>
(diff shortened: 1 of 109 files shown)


heshan0131:
This is awesome!!! Thanks for putting in so much work for it! Looks like some tests are still failing.

A trick to show my stack trace of the error message is to delete `| tap-spec` at the end of `test-browser` or `test-node` script in package.json

Joonalai:
Thanks @heshan0131 :) I will try to fix tests now.

Joonalai:
@heshan0131, I think the failing test is not related to localization.

macrigiuseppe:
@Joonalai please rebase with master and your error will go away

Joonalai:
@macrigiuseppe I just added some minor additions and rebased.

Joonalai:
@macrigiuseppe finally the build went trough successfully.

src/components/side-panel/layer-panel/vis-config-by-field-selector.js
```diff
@@ -50,17 +51,30 @@ export default class VisConfigByFieldSelector extends Component {
...
+    const {property, showScale, selectedField, description, intl, scaleOptions = []} = this.props;
 
     return (
       <SidePanelSection>
         <SidePanelSection>
           <PanelLabelWrapper>
             <PanelLabel>
-              {this.props.label || `${capitalizeFirstLetter(property)} based on`}
+              {(this.props.label && (
+                <FormattedMessage id={`property.${camelize(this.props.label)}`} />
```
heshan0131: The `camelize` transform now needs to be language specific. Does it only make sense for languages that are Latin based?

src/components/map/map-control.js
```diff
@@ -291,6 +302,41 @@ const MapDrawPanel = React.memo(
...
+                key={locale}
+                onClick={() => onSetLocale(locale)}
+                label={locale}
+                active={activeLocale === locale}
+              />
+            ))}
+          </StyledToolbar>
+        ) : null}
+        <MapControlButton
+          onClick={e => {
```
macrigiuseppe: take advantage of useCallback to replace this one

src/components/map/map-legend.js
```diff
@@ -67,14 +69,23 @@ export const StyledMapControlLegend = styled.div`
 export const VisualChannelMetric = ({name}) => (
   <div className="legend--layer__title">
     <span className="legend--layer_by">by </span>
-    <span className="legend--layer_color_field">{name}</span>
+    <span className="legend--layer_color_field">
+      <FormattedHTMLMessage
+        id={name}
+        values={{
+          name: useIntl().formatMessage({id: `property.${camelize(name)}`, defaultMessage: name})
```
macrigiuseppe: whenever in function stateless components you should always try to take advantage of useMemo and useCallback.
in this case you can do the following:
```
const values = useMemo(() => ({
  name: useIntl().formatMessage({id: `property.${camelize(name)}`, defaultMessage: name})
}), [name]);
```

src/components/side-panel/layer-manager.js
```diff
@@ -25,33 +25,43 @@ import PropTypes from 'prop-types';
...
-      selectedItems={layerBlending}
-      options={Object.keys(LAYER_BLENDINGS)}
-      multiSelect={false}
-      searchable={false}
-      onChange={updateLayerBlending}
-    />
-  </SidePanelSection>
-);
+const LayerBlendingSelector = ({layerBlending, updateLayerBlending, intl}) => {
+  const labeledLayerBlendings = Object.keys(LAYER_BLENDINGS).reduce((ret, key) => {
```
macrigiuseppe:
```
const labeledLayerBlendings = Object.keys(LAYER_BLENDINGS).reduce((acc, current) => ({
  ...acc,
  [intl.formatMessage({id: LAYER_BLENDINGS[current].label})]: current
}), {});
```

src/components/side-panel/layer-manager.js
```diff
@@ -25,33 +25,43 @@ import PropTypes from 'prop-types';
...
+    <SidePanelSection>
+      <PanelLabel>
+        <FormattedMessage id="layerBlending.title" />
+      </PanelLabel>
+      <ItemSelector
+        selectedItems={intl.formatMessage({id: LAYER_BLENDINGS[layerBlending].label})}
+        options={Object.keys(labeledLayerBlendings)}
+        multiSelect={false}
+        searchable={false}
+        onChange={blending => updateLayerBlending(labeledLayerBlendings[blending])}
```
macrigiuseppe: take advantage of useCallback
```
const onChange = useCallback((blending) => updateLayerBlending(labeledLayerBlendings[blending]), [updateLayerBlending, labeledLayerBlendings]);
```

src/components/side-panel/layer-panel/text-label-panel.js
```diff
@@ -51,7 +52,7 @@ export default class TextLabelPanel extends Component {
     const {updateLayerTextLabel, textLabel, fields} = this.props;
     const currentFields = textLabel.map(tl => tl.field && tl.field.name).filter(d => d);
     return (
-      <LayerConfigGroup label={'label'} collapsible>
+      <LayerConfigGroup label={'panel.label.label'} collapsible>
```
macrigiuseppe: can we find a better naming for this:
```
panel.label.label
```
having label.label is not ideal

src/reducers/ui-state-updaters.js
```diff
@@ -103,9 +106,10 @@ export const DEFAULT_MAP_CONTROLS = [
     ...final,
     [current]: DEFAULT_MAP_CONTROLS_FEATURES
   }),
-  {}
+  {mapLocale: {show: false, active: false, activeMapIndex: 0}}
```
macrigiuseppe: where is this udes?

Joonalai: At the moment this is used only during development and I will use it in the application based on Kepler. I thought that it would be nice to include this also in core, but since it is not (probably) in line with design of other components, I decided to leave it hidden by default.

Joonalai: At the moment this is used only during development and I will use it in the application based on Kepler. I thought that it would be nice to include this also in core, but since it is not (probably) in line with design of other components, I decided to leave it hidden by default.

src/components/side-panel/layer-panel/text-label-panel.js
Joonalai: That is true. I changed `panel.label...` to `panel.text...`.

src/components/side-panel/layer-panel/vis-config-by-field-selector.js
Joonalai: I added `defaultMessage: label` and `defaultMessage: property`. That way the original value will be used if camelized string is not found from the localization files. Do you think that it will be sufficient?

heshan0131:
@Joonalai  Thank you  so much for the huge amount of work you put in. I only find 2 tiny glitchs:

#### 1.
Screen Shot 2020-03-22 at 2 22 47 PM

#### 2. load-storage-map component was later added, it needs react-intl too:
Screen Shot 2020-03-22 at 2 30 30 PM


Other than that everything looks great.  I will stamp your PR for now, trusting you will fix those

2 Follow up questions:

1. Will you be willing to add `react-intl` support for our demo-app too? There are some customized components there. Also it would be nice to have a tiny `language` button in the demo app for people to choose languages. something like:

Screen Shot 2020-03-22 at 2 52 23 PM

2. Would you like to add a doc in `docs/api-reference/localization` and some guidance for future developers about how to add another translation?

Joonalai:
Thank you @heshan0131, I will make the final improvements and actually the initial button is already implemented.

macrigiuseppe:
@Joonalai please rebase and make sure your test coverage does not decrease

Joonalai:
I am not able to run the e2e tests locally. I will try again later.

Joonalai:
@heshan0131 I don't get it, I managed to run the e2e tests locally without errors after pulling the newest changes from master, but still Travis build fails on them. Any ideas how to fix this?

heshan0131:
There is some issue with our e2e test on ci. I will land this PR for now, and fix our pipeien later.

Issue #998, opened 2020-03-10
[Bug] Cloud provider authentication errors doesn't clear after login


neokore:
**Describe the bug**
When an authentication error appears, it doesn't clear after authentication is fixed.

**To Reproduce**
1. Go to "Add Data".
2. Go to "Try sample Data".
3. Load any sample map.
4. When loaded, go to "Cloud Storage" > "Save".
5. Fill the name field.
6. Click on the "Save button".
7. An authentication error message should appear.
8. Select and login with any cloud provider using the "Cloud Storage" boxes above.
9. After login, the authentication error remains on screen.

**Expected behavior**
After loging in, the authentication error must disappear.

macrigiuseppe:
this is duplicate of #990

macrigiuseppe:
fixed

Issue #997, opened 2020-03-10
[Bug] Share map shows last saved URL


neokore:
**Describe the bug**
After saving a map, share map URL modal shows previously saved map URL as sharing URL.

**To Reproduce**
1. Load any map
2. Go to "Cloud Storage" > "Save"
3. Select a cloud provider
4. Fill the name field
5. Click on "Save"
6. After map is saved, go to "Share" > "Share Map URL"
7. On the modal, the "Share URL" field shows an URL

**Expected behavior**
"Share URL" field must not appear because the map is not shared yet.

macrigiuseppe:
fixed

PR #999, opened 2020-03-10
[Bug] Reset selected provider status after loading and before sharing

neokore:
Related to  #997 and #998.

Forces clearing selected provider errors and upload response after selecting a new provider and before sharing a map URL.

### DIFF
--- a/src/components/side-panel.js
+++ b/src/components/side-panel.js
@@ -103,7 +103,6 @@ export default function SidePanelFactory(
       datasets: PropTypes.object.isRequired,
       visStateActions: PropTypes.object.isRequired,
       mapStyleActions: PropTypes.object.isRequired,
-      providerActions: PropTypes.object.isRequired,
       availableProviders: PropTypes.object,
       mapSaved: PropTypes.string,
       panels: PropTypes.arrayOf(PropTypes.object)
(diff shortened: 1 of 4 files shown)


src/components/side-panel.js
```diff
@@ -161,7 +163,10 @@ export default function SidePanelFactory(
       this.props.uiStateActions.toggleModal(SAVE_MAP_ID);
     };
 
-    _onClickShareMap = () => this.props.uiStateActions.toggleModal(SHARE_MAP_ID);
+    _onClickShareMap = () => {
+      this.props.providerActions.resetProviderStatus();
```
macrigiuseppe: why are we resetting the status when we open the share modal dialog?

we should reset the provider status whenever we complete a sharing action,e .g. sharing successful

neokore: I applied it when opening to ensure that there is no error pending error in the state from any source.
In any case, I think that it must be reset after closing the save or share dialog in order to show the messages to user; but I'm afraid of showing other pending error message that could not be cleared before.

heshan0131: `resetProviderStatus` is called when Share map modal is closed. it sets 
```
  isProviderLoading: false,
  providerError: null,
  successInfo: {}
```

It's not very pretty but works. If you need to, we should probably call reset status when close `LoadDataModal` `SaveMapMadal` and `ShareMapModal`

neokore: Ok, I'll apply it this way.

neokore: I've applied as you advised and it looks way better, thanks! Is it OK for you?

Issue #990, opened 2020-03-09
[Bug] Log out in Load from Storage doesn't refresh the UI


VictorVelarde:
**Describe the bug**
Pressing the Logout link under cloud providers, doesn't refresh the UI

**To Reproduce**
1. Go to 'Add data'
2. Click on 'Load from Storage'
3. Login into a provider, then Logout

**Expected behavior**
UI refreshes accordingly

macrigiuseppe:
fixed

PR #1000, opened 2020-03-10
better handling provider tile update

heshan0131:
https://github.com/keplergl/kepler.gl/issues/990

### DIFF
--- a/src/components/modals/load-storage-map.js
+++ b/src/components/modals/load-storage-map.js
@@ -26,8 +26,11 @@ import LoadingDialog from './loading-dialog';
 import {Button} from 'components/common/styled-components';
 import CloudTile from './cloud-tile';
 import {Base, ArrowLeft} from 'components/common/icons';
+import ProviderModalContainer from './provider-modal-container';
 
-const StyledProviderSection = styled.div`
+const StyledProviderSection = styled.div.attrs({
+  className: 'provider-selection'
+})`
   display: flex;
 `;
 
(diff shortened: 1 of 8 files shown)


src/components/modals/load-storage-map.js
```diff
@@ -215,17 +218,23 @@ const VisualizationItem = ({vis, onClick}) => {
     </StyledVisualizationItem>
   );
 };
-const ProviderSelect = ({cloudProviders, onSetCloudProvider, currentProvider}) =>
+
+export const ProviderSelect = ({
+  cloudProviders = [],
+  onSelect,
+  onSetCloudProvider,
```
macrigiuseppe: is this still used?

src/components/modals/load-storage-map.js
```diff
@@ -262,21 +277,38 @@ function LoadStorageMapFactory() {
...
       });
     }
+
+    _clickBack = () => {
+      this.setState({showProviderSelect: true});
+    };
+
+    _selectProvider = providerName => {
+      this.props.onSetCloudProvider(providerName);
+      const provider = (this.props.cloudProviders || []).find(p => p.name === providerName);
```
macrigiuseppe: what if provider is null?

I know we have a check in getSavedMaps but should we even dispatch the action if provider is undefined

heshan0131: `_selectProvider` is callbacks from `ProviderSelect` which is rendered by cloudProviders. I don't  see in which case provider will be underfined

src/components/modals/load-storage-map.js
heshan0131: it is passed to `ProviderSelect`

Issue #1002, opened 2020-03-11
[Bug] Lack of feedback when loading a map from a cloud provider


neokore:
**Describe the bug**
When a user loads a map from a URL from a cloud provider, there is no loading animation or error messages when something goes wrong.

**To Reproduce**
1. Navigate to a shared map URL from a cloud provider
2. While it is loading, it just shows the "Load Data" modal.
3. In case of error, it just remains on the "Load Data" modal without feedback

**Expected behavior**
While loading, it should show the loading animation used when you load a map using the modal. In case of error, it should show an error notification.

macrigiuseppe:
fixed

PR #1003, opened 2020-03-11
Loading and error feedback for shared maps loaded from URL

neokore:
Related to #1002


src/reducers/provider-state-updaters.js
```diff
@@ -48,6 +49,7 @@ import KeplerGlSchema from 'schemas';
 
 export const INITIAL_PROVIDER_STATE = {
   isProviderLoading: false,
+  isCloudMapLoading: false,
```
heshan0131: why can't we just set `isProviderLoading`? looks like in all places these 2 values are the same

neokore: I tried, but it is used when getting maps list too and it enters in an endless loop and I preferred not to refactor it for tomorrow as it may affect the rest of map loading methods.

PR #1006, opened 2020-03-12
Fix typo in sample code for `addDataToMap`

plant99:
addDataToMap takes `data.options` as example, while in the sample code it's mistyped as `data.option`


heshan0131:
This file is autogenerated, you should change the jsDoc of `src/actions/actions.js` then run `npm run docs`

PR #1017, opened 2020-03-22
[Enhancement] Move layer panel visible toggle to end

heshan0131:
- Move eye to the end
- Make layer panel header a factory


heshan0131:
Screen Shot 2020-03-22 at 1 29 53 AM

jonsadka:
Niiiice!!!!

PR #1038, opened 2020-03-29
[Enhancement] Remove table cell size limit and increased height

macrigiuseppe:
Increased header size and remove cell header size limit

Screen Shot 2020-03-28 at 9 44 45 PM


heshan0131:
I feel like the header right now is too high. Can you add a constraint around column name and the tag, so they don't just expand and too far apart when header height increase? You can probably do it by adding `justify-content: flex-start`, which means you might now need to increase the header height after all

77840527-6898a780-713d-11ea-9315-4cf932f52ce1

macrigiuseppe:
> I feel like the header right now is too high. Can you add a constraint around column name and the tag, so they don't just expand and too far apart when header height increase? You can probably do it by adding `justify-content: flex-start`, which means you might now need to increase the header height after all
> 
> 77840527-6898a780-713d-11ea-9315-4cf932f52ce1

sounds good

PR #1046, opened 2020-03-31
Fixes for case sensitive fields in CARTO storage

VictorVelarde:
Hi, this is a small fix to deal in a proper way with case sensitive fields in CARTO provider.

### DIFF
--- a/examples/demo-app/src/cloud-providers/carto/carto-provider.js
+++ b/examples/demo-app/src/cloud-providers/carto/carto-provider.js
@@ -21,7 +21,7 @@
 import {OAuthApp} from '@carto/toolkit';
 import Console from 'global/console';
 import CartoIcon from './carto-icon';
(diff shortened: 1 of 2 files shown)


examples/demo-app/src/cloud-providers/carto/carto-provider.js
```diff
@@ -21,7 +21,7 @@
 import {OAuthApp} from '@carto/toolkit';
 import Console from 'global/console';
 import CartoIcon from './carto-icon';
-import {formatCsv} from 'kepler.gl/processors';
+import {formatCsv} from 'kepler.gl/processors/data-processor';
```
heshan0131, line 24: `formatCsv` is exported from processors. add the file path will cause demo-app failed to build on its own

VictorVelarde: Sorry, I didn't notice it. Thx for pointing it out @heshan0131

PR #1066, opened 2020-04-08
[Perf] make iconGeometry static to prevent redundant loading

mshuffett:
Before this change, iconGeometry was fetched (either from a CDN or disk
cache) every time IconLayer was constructed. This fetching would take
several hundred milliseconds even from disk. While the fetching was in
progress the icons would dissapear from the map which is especially
problematic when loading data using addDataToMap.

Signed-off-by: Michael Shuffett <michael@quantcollective.com>

### DIFF
--- a/src/layers/icon-layer/icon-layer.js
+++ b/src/layers/icon-layer/icon-layer.js
@@ -59,6 +59,7 @@ function flatterIconPositions(icon) {
 }
 
 export default class IconLayer extends Layer {
+  static iconGeometry = null;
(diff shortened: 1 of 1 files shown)


macrigiuseppe:
@mshuffett can you please make sure this will also work when you have multiple icon layers using different set of icons?
Make. sure all tests are passing correctly

igorDykhta:
The comment was never addressed, so I’m closing this stale PR.

Issue #1067, opened 2020-04-08
Geocoder Search


b2kdaman:
**Is your feature request related to a problem? Please describe.**
There is no location based search for kepler.gl

**Describe the solution you'd like**
Using `react-mapbox-geocoder` and additional changes in UI, ie Interaction panel.
Additional kepler-gl.js panel will be added to layout which will appear in the top center of UI.
It will be a simple type-ahead text box with 5 relevant items in dropdown to pick. When user will select the item new hidden datasource and layer will be added, red pin will appear in the provided location.

**Describe alternatives you've considered**
For our data visualization tool we have implemented this feature through the wrapper.

**Additional context**

thunter009:
A much needed feature IMO -- thank you!

heshan0131:
Love it! Will take a look today. Thanks!

kindjacket:
This is a great feature. The pin is wrong though. It is anchored by its center rather than the base. This makes it confusing as the user zooms out.
<img width="1205" alt="Screen Shot 2020-05-11 at 09 16 30" src="https://user-images.githubusercontent.com/7778402/81584975-59299280-9368-11ea-9daf-f012211236eb.png">
<img width="1207" alt="Screen Shot 2020-05-11 at 09 16 40" src="https://user-images.githubusercontent.com/7778402/81585017-647cbe00-9368-11ea-878c-fb9e40b33de0.png">
<img width="1305" alt="Screen Shot 2020-05-11 at 09 16 49" src="https://user-images.githubusercontent.com/7778402/81585049-6fcfe980-9368-11ea-9df8-697cc371ca0f.png">

b2kdaman:
This is how icon layer is rendered. @heshan0131 any suggestions?

PR #1068, opened 2020-04-08
[Feat] Geocoder Search

b2kdaman:
Created this PR for a feature: 
https://github.com/keplergl/kepler.gl/issues/1067

### DIFF
--- a/src/components/geocoder-panel.js
+++ b/src/components/geocoder-panel.js
@@ -55,7 +55,7 @@ const GeocoderPanelContent = styled.div`
   background-color: ${props => props.theme.panelBackground};
   padding: 0.5em 1em;
   position: absolute;
-  top: 1em;
+  top: 20px;
   left: 50%;
   margin-left: -20em;
   width: 40em;
@@ -172,13 +172,6 @@ export class GeocoderPanel extends Component {
           config: {
             visState: {
               layers: [ICON_LAYER]
-            },
-            interactionConfig: {
-              tooltip: {
-                fieldsToShow: {
-                  geocoder_dataset: ['text']
-                }
-              }
             }
           }
         }
--- a/src/components/geocoder-panel.js
+++ b/src/components/geocoder-panel.js
@@ -140,21 +140,15 @@ export class GeocoderPanel extends Component {
     mapboxApiAccessToken: PropTypes.string.isRequired
   };
 
-  constructor() {
-    super();
-    this.onSelected = this.onSelected.bind(this);
-    this.removeMarker = this.removeMarker.bind(this);
-
-    this.state = {
-      selectedGeoItem: null
-    };
-  }
+  state = {
+    selectedGeoItem: null
+  };
 
   removeGeocoderDataset() {
     this.props.dispatch(removeDataset(GEOCODER_DATASET_NAME));
   }
 
-  onSelected(viewport = null, geoItem) {
+  onSelected = (viewport = null, geoItem) => {
     const {
       center: [lon, lat],
       text,
@@ -183,14 +177,14 @@ export class GeocoderPanel extends Component {
     this.setState({
       selectedGeoItem: geoItem
     });
-  }
+  };
 
-  removeMarker() {
+  removeMarker = () => {
     this.setState({
       selectedGeoItem: null
     });
     this.removeGeocoderDataset();
-  }
+  };
 
   render() {
     const {isGeocoderEnabled, mapboxApiAccessToken} = this.props;
--- a/src/components/geocoder-panel.js
+++ b/src/components/geocoder-panel.js
@@ -1,0 +1,243 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import {SidePanelSection, PanelLabel} from './common/styled-components';
+import {FormattedMessage} from 'react-intl';
+import Geocoder from 'react-mapbox-gl-geocoder';
+import Processors from 'processors';
+import {fitBounds, addDataToMap, removeDataset} from 'actions';
+
+const GEOCODER_DATASET_NAME = 'geocoder_dataset';
+const QUERY_PARAMS = {};
+const GEO_OFFSET = 0.1;
+const ICON_LAYER = {
+  id: 'geocoder_layer',
+  type: 'icon',
+  config: {
+    label: 'Geocoder Layer',
+    color: [255, 0, 0],
+    dataId: GEOCODER_DATASET_NAME,
+    columns: {
+      lat: 'lt',
+      lng: 'ln',
+      icon: 'icon',
+      label: 'text'
+    },
+    isVisible: true,
+    hidden: true,
+    visConfig: {
+      radius: 80
+    }
+  }
+};
+
+const GeocoderPanelContent = styled.div`
+  background-color: ${props => props.theme.panelBackground};
+  padding: 0.5em 1em;
+  position: absolute;
+  top: 1em;
+  left: 50%;
+  margin-left: -20em;
+  width: 40em;
+  z-index: 10000;
+  box-sizing: border-box;
+`;
+
+const GeocoderWrapper = styled.div`
+  color: ${props => props.theme.textColor};
+  font-size: ${props => props.theme.fontSize};
+
+  .react-geocoder {
+    position: relative;
+  }
+
+  .react-geocoder input {
+    ${props => props.theme.secondaryInput};
+  }
+
+  .react-geocoder-results {
+    background-color: ${props => props.theme.panelBackground};
+    position: absolute;
+    width: 43.5em;
+  }
+
+  .react-geocoder-item {
+    ${props => props.theme.dropdownListItem};
+    ${props => props.theme.textTruncate};
+  }
+
+  .remove-layer {
+    background: transparent;
+    border: none;
+    bottom: 28px;
+    color: ${props => props.theme.textColor};
+    cursor: pointer;
+    display: inline;
+    font-size: 16px;
+    padding: 2px 8px;
+    position: absolute;
+    right: 16px;
+
+    :hover,
+    :focus,
+    :active {
+      background: transparent !important;
+      border: none;
+      box-shadow: 0;
+      color: ${props => props.theme.textColor};
+      opacity: 0.6;
+      outline: none;
+    }
+  }
+`;
+
+function generateGeocoderDataset(lat, lon, text) {
+  return {
+    data: Processors.processRowObject([
+      {
+        lt: lat,
+        ln: lon,
+        icon: 'place',
+        text
+      }
+    ]),
+    id: GEOCODER_DATASET_NAME,
+    info: {
+      hidden: true,
+      id: GEOCODER_DATASET_NAME,
+      label: GEOCODER_DATASET_NAME
+    }
+  };
+}
+
+function isValid(key) {
+  return /pk\..*\..*/.test(key);
+}
+
+export class GeocoderPanel extends Component {
+  static propTypes = {
+    isGeocoderEnabled: PropTypes.bool.isRequired,
+    mapboxApiAccessToken: PropTypes.string.isRequired
+  };
+
+  constructor() {
+    super();
+    this.onSelected = this.onSelected.bind(this);
+    this.removeMarker = this.removeMarker.bind(this);
+
+    this.state = {
+      selectedGeoItem: null
+    };
+  }
+
+  removeGeocoderDataset() {
+    this.props.dispatch(removeDataset(GEOCODER_DATASET_NAME));
+  }
+
+  onSelected(viewport = null, geoItem) {
+    const {
+      center: [lon, lat],
+      text,
+      bbox
+    } = geoItem;
+    this.removeGeocoderDataset();
+    this.props.dispatch(
+      addDataToMap({
+        datasets: [generateGeocoderDataset(lat, lon, text)],
+        options: {
+          keepExistingConfig: true
+        },
+        config: {
+          version: 'v1',
+          config: {
+            visState: {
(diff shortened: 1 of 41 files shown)


src/components/geocoder-panel.js
```diff
@@ -0,0 +1,218 @@
...
+    onSelected(viewport = null, geoItem) {
+      const {
+        center: [lon, lat],
+        address,
+        bbox
+      } = geoItem;
+      this.removeGeocoderDataset();
+      this.props.dispatch(
+        addDataToMap({
+          datasets: [generateGeocoderDataset(lat, lon, address)]
```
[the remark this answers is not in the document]
b2kdaman: But it centers on the pin or **bbox** passed. Thanks for the advice! I'll use it.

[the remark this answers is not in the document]
heshan0131: You can also manually pass in the desired `mapState` in the config, and set `centerMap:false`

```
    options: {
          keepExistingConfig: true,
          centerMap: false
        },
       config: {
         version: 'v1',
         config: {
           mapState: {
               latitude: ...
               longitude: ...
               zoom
           },
           visState: {
             layers: [ICON_LAYER]
          }
        }
```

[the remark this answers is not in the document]
b2kdaman: But it doesn't support bbox...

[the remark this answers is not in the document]
b2kdaman: `keepExistingConfig: true` didn't do the trick
Is my config structure OK?

b2kdaman:
@heshan0131 updated the config part, seems to work BUT: only when user moves the mouse...

src/components/geocoder-panel.js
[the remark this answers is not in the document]
b2kdaman: @heshan0131 updated the config part, seems to work BUT: only when user moves the mouse...

b2kdaman:
@heshan0131 any updates?

src/components/geocoder-panel.js
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  onSelected(viewport = null, geoItem) {
+    const {
+      center: [lon, lat],
+      text,
+      bbox
+    } = geoItem;
+    this.removeGeocoderDataset();
+    this.props.dispatch(
```
heshan0131, line 164: this can be problematic, because you are adding dataset with the same id over and over again. I would check if the dataset is already exist (there is already some geocode marker), if it doesn't, call`addDataToMap`, with the dataset and layer.  if it does, create a new `visStateAction` call it `appendRowsToDataset`

make the updater as this:
```js
export function appendRowsToDataset(state, action) {
  const {rows, dataId} = action;
  if (!state.datasets[dataId]) {
    // dataset doesnot exist
    console.warn(`dataset.${dataId} doesn't exist`);
    return state;
  }
  const oldDataset = {
    ...state.datasets[dataId]
  };

  // create a temp dataset and run filter on it
  const tempDataset = {
    id: dataId,
    filterRecord: {},
    allData: rows,
    allIndexes: rows.map((_, i) => i)
  };
  const layersToFilter = (state.layers || []).filter(l => l.config.dataId === dataId);
  const appliedFilters = (state.filters || []).filter(d => shouldApplyFilter(d, dataId));
  
  const tempfiltered = filterDataset(tempDataset, appliedFilters, layersToFilter);
  const allData = oldDataset.allData.concat(rows);
  const offsetIdx = i => i + oldDataset.allData.length;

  const newDataset = {
    ...oldDataset,
    allData,
    allIndexes: allData.map((_, i) => i),
    filteredIndex: oldDataset.filteredIndex.concat(tempfiltered.filteredIndex.map(offsetIdx)),
    filteredIndexForDomain: oldDataset.filteredIndexForDomain.concat(tempfiltered.filteredIndex.map(offsetIdx))
  }
  const newState = {
    ...state,
    datasets:
    {
      ...state.datasets,
      [dataId]: newDataset
    }
  }

  // update layers associatred with this data
  return updateAllLayerDomainData(newState, [dataId]);
}
```

I just worte this one, haven't test it yet, but i think this function will also help enable data streaming in the future

b2kdaman: @heshan0131 is this related to mouse move bug? I'm still trying to figure it out. I can always ad _.uniqueId to dataset. Also I'm removing that dataset before adding. I haven't seen any problems yet.

b2kdaman: @heshan0131 every time new dataset is generated since we only need one value and one row. Please check https://deploy-preview-1068--keplergl.netlify.com/ to see how Geocoder search works now.

src/components/geocoder-panel.js
```diff
@@ -0,0 +1,243 @@
...
+    this.state = {
+      selectedGeoItem: null
+    };
+  }
+
+  removeGeocoderDataset() {
+    this.props.dispatch(removeDataset(GEOCODER_DATASET_NAME));
+  }
+
+  onSelected(viewport = null, geoItem) {
```
heshan0131, line 157: make it ` nSelected = (viewport = null, geoItem) {...}` will auto bind `this`

src/components/geocoder-panel.js
```diff
@@ -0,0 +1,243 @@
...
+    );
+    this.props.dispatch(
+      fitBounds(bbox || [lon - GEO_OFFSET, lat - GEO_OFFSET, lon + GEO_OFFSET, lat + GEO_OFFSET])
+    );
+    this.setState({
+      selectedGeoItem: geoItem
+    });
+  }
+
+  removeMarker() {
```
heshan0131, line 195: make it `removeMarker = () {...}` will auto bind `this`

src/components/geocoder-panel.js
```diff
@@ -0,0 +1,243 @@
...
+}
+
+export class GeocoderPanel extends Component {
+  static propTypes = {
+    isGeocoderEnabled: PropTypes.bool.isRequired,
+    mapboxApiAccessToken: PropTypes.string.isRequired
+  };
+
+  constructor() {
+    super();
```
heshan0131, line 144: make `onSelected` and `removeMarker` auto bind. define [state as directly to the class](https://github.com/keplergl/kepler.gl/blob/master/src/components/common/checkbox.js#L74-L76
), then you don't need `constructor` anymore

heshan0131:
Can you make sure geocoder is saved to` interactionConfig` when export map, and saved map with geocode can be loaded?

b2kdaman:
@heshan0131 updated with layer hiding

b2kdaman:
@heshan0131 I think that adding geocoder point data to `interactionConfig` is useful in next iteration when I'll move it to separate mapbox layer as we previously talked about

macrigiuseppe:
@b2kdaman this is great, I created a map using geocoder and found the following two issues 

0. The geocoder bar z-index is too high and covers the dialog box:

Screen Shot 2020-04-21 at 3 36 11 PM

Make sure the geocoder z-index is lower than the modal box on.

1. The geocoder icon doesn't disappear from the map when a user disables geocoder from the panel and the same thing goes for the new dataset. See below

geocoder-disable-bug

In this case, once the user deactivates the geocoder you can simply remove the newly created dataset and that should remove the icon layer: You can find remove dataset here: https://docs.kepler.gl/docs/api-reference/actions/actions#removedataset.

2. I exported a map where geocoder was enabled with a location. When I try to import the map back to Kepler, I see the icon on the map but the geocode switch is disabled, the geocoder bar is not visible and the geocode layer appears in the list of layers

import-geocoder-map

I think you need to update the vi-state-schema in order to make sure you export correctly geocoder information 
https://github.com/keplergl/kepler.gl/blob/master/src/schemas/vis-state-schema.js#L553

and to make sure you import both geocoder and layer info correctly you can look at this: ihttps://github.com/keplergl/kepler.gl/blob/dec712e39899c1aa9027cf40a419ffcc8c54b344/src/reducers/vis-state-merger.js#L191

3. Read-only mode
Maps have a read-only mode and geocoder bar should not appear in read mode. You can look at the way we handle readMode in here: https://github.com/keplergl/kepler.gl/blob/dec712e39899c1aa9027cf40a419ffcc8c54b344/src/components/map-container.js#L87

@heshan0131 I looked into exporting data, should a user be able to export the geocoder as seen here?

Screen Shot 2020-04-21 at 3 37 43 PM

b2kdaman:
@macrigiuseppe 
0 - got it.
1 - can be implemented in this iteration, but I'd prefer it in next (look at `interactionConfig` comments}
2 - same it will be easier to manage with separate layer in mapbox(next iteration)
3 - I feel like Geocoder is a valid feature in read only mode still. @heshan0131 what do you think?

b2kdaman:
@macrigiuseppe #0 and #1 are fixed

macrigiuseppe:
> @macrigiuseppe #0 and #1 are fixed

@b2kdaman I checked and I have the following notes:
0. Fixed
1. This is not fixed. Disabling geocoder in the control panel does still not remove the geocoder icon nor the dataset. This is part of the main functionality and it should be implemented.
2. I am not too clear on the reasoning why export and import functionalities is depending on the next iteration using mapbox layer. Import and export use the underline structure from redux which is used to generate layers, whether we generate deck.gl or mapbox is completely independent from the data structure. Besides, if people try to export a map with geocoder and import again they will see an extra layer which was not part of their original map; this experience is not just inconsistent but people may consider it a bug.
3. Read-Only mode only allows people to visualize the map and not change anything. Any feature that allows to change the visualization should not be available in read-only mode, therefore the geocoder input panel should be always hidden in read-only mode.

I am not really clear on the different steps that should be implemented in this iteration vs the next one but I think the core functionality should be fully working when we merge it.

src/reducers/vis-state-updaters.js
```diff
@@ -1069,6 +1028,54 @@ export const layerHoverUpdater = (state, action) => ({
...
+      }
+    });
+  }
+
+  const newState = {
+    ...state,
+    interactionConfig
+  };
+
+  if (config.id === 'geocoder' && !config.enabled) {
```
macrigiuseppe, line 1072: why ```!config.enabled``` ?

src/reducers/vis-state-updaters.js
```diff
@@ -1069,6 +1028,54 @@ export const layerHoverUpdater = (state, action) => ({
   hoverInfo: action.info
 });
 
+/* eslint-enable max-statements */
```
macrigiuseppe, line 1031: is this necessary?

b2kdaman:
I checked #1 couple of times before I pushed. I'll check again. 
Okay, I'll do 2 & 3.
But since you can't save in read only mode Geocoder pin will only appear on the user side/session. @macrigiuseppe

macrigiuseppe:
> I checked #1 couple of times before I pushed. I'll check again.
> Okay, I'll do 2 & 3.
> But since you can't save in read only mode Geocoder pin will only appear on the user side/session. @macrigiuseppe

@b2kdaman the read-only mode is an export mode. If a user creates a map and place a geocoder icon somewhere, said icon should be visible in read-only mode.
The geocoder address input should not be visible because it would allow users to change the icon location. If this is not clear, please reach out directly on slack and we can go over this.

src/reducers/vis-state-updaters.js
heshan0131: because it needs to remove the dataset when config is disabled

test/fixtures/config_v0_arc_cluster_point.js
```diff
@@ -481,6 +481,9 @@ export const parsedConfigV0 = {
       },
       brush: {
         enabled: false
+      },
+      geocoder: {
```
heshan0131, line 485: this is for backward compatibility test, you should not add `geocoder` in here. The idea is when loading an older version of config which doesn't contain geocoder, loading should not faill, and geocoder should be diabled by default

test/fixtures/config_v0_geojson_point.js
```diff
@@ -168,6 +168,9 @@ export const parsedConfigV0 = {
       },
       brush: {
         enabled: false
+      },
+      geocoder: {
```
heshan0131, line 172: same above

test/fixtures/config_v0_geojson_poly_fill_ele.js
```diff
@@ -147,6 +147,9 @@ export const parsedConfigV0 = {
       },
       brush: {
         enabled: false
+      },
+      geocoder: {
```
heshan0131, line 151: same above

test/fixtures/config_v0_geojson_polygon.js
```diff
@@ -282,6 +282,9 @@ export const parsedConfigV0 = {
       },
       brush: {
         enabled: false
+      },
+      geocoder: {
```
heshan0131, line 286: same above

test/fixtures/polygon-filter-map.js
```diff
@@ -355,6 +355,9 @@ export const polygonFilterMap = {
           brush: {
             size: 0.5,
             enabled: false
+          },
+          geocoder: {
```
heshan0131, line 359: remove this

test/fixtures/polygon.js
```diff
@@ -192,6 +192,9 @@ export const mockPolygonData = {
     brush: {
       size: 0.5,
       enabled: false
+    },
+    geocoder: {
```
heshan0131, line 196: remove this

test/fixtures/state-saved-v0.js
```diff
@@ -214,6 +214,7 @@ export const savedStateV0 = {
                 fieldIdx: 7
               }
             },
+            hidden: false,
```
heshan0131, line 217: this file should not be changed

test/fixtures/state-saved-v0.js
```diff
@@ -259,6 +260,7 @@ export const savedStateV0 = {
...
```
heshan0131, line 263: remove this

test/fixtures/state-saved-v1-1.js
```diff
@@ -648,6 +648,7 @@ export const savedStateV1 = {
               columns: {
                 geojson: 'simplified_shape_v2'
               },
+              hidden: false,
```
heshan0131, line 651: this file should not be changed

test/fixtures/state-saved-v1-2.js
```diff
@@ -86,7 +86,10 @@ export const stateSavedV1_2 = {
             },
             enabled: true
           },
-          brush: {size: 0.5, enabled: false}
+          brush: {size: 0.5, enabled: false},
+          geocoder: {
```
heshan0131, line 90: this file should not be changed

test/fixtures/state-saved-v1-4.js
```diff
@@ -149,6 +150,9 @@ export const stateSavedV1 = {
           brush: {
             size: 0.5,
             enabled: false
+          },
+          geocoder: {
```
heshan0131, line 154: the saved state should not be changed, the merged config should have geocoder set to `{enablced: false}`

test/fixtures/state-saved-v1-3.js
```diff
@@ -283,7 +285,10 @@ export const savedStateV1 = {
             },
             enabled: true
           },
-          brush: {size: 0.5, enabled: false}
+          brush: {size: 0.5, enabled: false},
+          geocoder: {
```
heshan0131, line 289: the saved state should not be changed, the merged config should have geocoder set to `{enablced: false}`

test/fixtures/state-saved-v1-5.js
```diff
@@ -214,6 +215,9 @@ export const savedStateV1TripGeoJson = {
           brush: {
             size: 0.5,
             enabled: false
+          },
+          geocoder: {
```
heshan0131, line 219: the saved state should not be changed, the merged config should have geocoder set to `{enablced: false}`

test/fixtures/state-saved-v1-6.js
```diff
@@ -750,6 +750,9 @@ export const savedStateWIthNonValidFilters = {
           brush: {
             size: 1,
             enabled: false
+          },
+          geocoder: {
```
heshan0131, line 754: the saved state should not be changed, the merged config should have geocoder set to `{enablced: false}`

test/fixtures/state-saved-v1-7.js
```diff
@@ -751,6 +751,9 @@ export const savedStateV1InteractionCoordinate = {
           },
           coordinate: {
             enabled: true
+          },
+          geocoder: {
```
heshan0131, line 755: the saved state should not be changed, the merged config should have geocoder set to `{enablced: false}`

b2kdaman:
@heshan0131 got it

Issue #992, opened 2020-03-09
[Bug] When aggregation takes too long only a few color bins show up. 


MorganGellert:
**Describe the bug**
When doing a hex bin aggregation on a lot of data where you want to color by a field (aggregating by mode) and the aggregation takes too long or the granularity is reduced to small bins bins only show up for a few of the classes. 

**To Reproduce**
If you use the earthquake demo and switch to the hexbin and color by MagType you may see that very little data shows up for a radius of 2km but much more shows up for a radius of 5 km. 

The following warning appears as well every time the config changes in the sidebar. Unclear how related it is but it may point in the correct direction.
`Warning: Cannot update a component from inside the function body of a different component.
    in DeckGL (created by MapContainer)
    in div (created by Context.Consumer)
    in div (created by AutoSizer)
    in AutoSizer (created by StaticMap)
    in div (created by StaticMap)
    in StaticMap (created by InteractiveMap)
    in div (created by InteractiveMap)
    in InteractiveMap (created by MapContainer)
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by MapContainer)
    in MapContainer (created by KeplerGL)
    in div (created by KeplerGL)
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by KeplerGL)
    in ThemeProvider (created by KeplerGL)
    in KeplerGL (created by Context.Consumer)
    in WithTheme(KeplerGL) (created by ConnectFunction)
    in ConnectFunction (created by WithConnectSelector)
    in WithConnectSelector (created by Container)
    in Container (created by ConnectFunction)
    in ConnectFunction (created by AutoSizer)
    in div (created by AutoSizer)
    in AutoSizer (created by App)
    in div (created by App)
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by App)
    in ThemeProvider (created by App)
    in App (created by ConnectFunction)
    in ConnectFunction (created by RouterContext)
    in RouterContext (created by Router)
    in Router (created by Root)
    in Provider (created by Root)
    in Root`

**Desktop (please complete the following information):**
 - OS: Mac Os Mojave
 - Browser chrome and firefox

heshan0131:
Hi @MorganGellert the error is actually unrelated to your aggregation case. How many points are we talking about? Do you have other filters applied at the same time?

MorganGellert:
[...]

PR #1070, opened 2020-04-09
[Bug] Fix out-of-domain coloring for valid strings

MorganGellert:
deck.gl BinSorter only properly sorts numbers. It uses '-' as it's
comparison operator. This operator returns NaN for strings which
causes the domain to be sorted randomly. We then decide if a
value is valid by comparing to the first and last elements of the
"sorted" domain which do not provide accurate bounds for string
elements. Short of making BinSorter work for multiple types, the
easiest fix is to check if a string value is in the valueDomain
of the dimensonState.

### DIFF
--- a/src/deckgl-layers/layer-utils/cpu-aggregator.js
+++ b/src/deckgl-layers/layer-utils/cpu-aggregator.js
@@ -19,9 +19,10 @@
 // THE SOFTWARE.
 
 /* eslint-disable guard-for-in */
-import {_BinSorter as BinSorter, AGGREGATION_OPERATION} from '@deck.gl/aggregation-layers';
+import {AGGREGATION_OPERATION} from '@deck.gl/aggregation-layers';
 import {console as Console} from 'global/window';
 
+import EnhancedBinSorter from './enhanced-bin-sorter';
 import {aggregate} from 'utils/aggregate-utils';
 import {AGGREGATION_TYPES, SCALE_FUNC} from 'constants/default-settings';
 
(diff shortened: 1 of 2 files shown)


MorganGellert:
Addresses  #992

macrigiuseppe:
@MorganGellert can you please fix the DCO check and also, may you provide some screenshots (before and after) to highlight your changes?
Thank you

MorganGellert:
Before: Screen Shot 2020-04-14 at 10 16 45 AM
After: Screen Shot 2020-04-14 at 10 14 16 AM

As you can see many bins which have data are missing. The legend gets it right, bins exist with all of those colors. In the before picture only Mx and Unk are getting flagged as "in domain" but most of the bins are labeled Md which do not show up at all.

src/deckgl-layers/layer-utils/cpu-aggregator.js
```diff
@@ -182,7 +182,8 @@ function getSubLayerAccessor(dimensionState, dimension, layerProps) {
     const cv = bin && bin.value;
     const domain = scaleFunc.domain();
 
-    const isValueInDomain = cv >= domain[0] && cv <= domain[domain.length - 1];
+    const isValueInDomain =
+      (cv >= domain[0] && cv <= domain[domain.length - 1]) || valueDomain.includes(cv);
```
heshan0131, line 186: I am concerned about performance. this function is called in each render pass on all bins. if this is only to cover categorical value, a more sensible solution is to call deck.gl bin sorter to properly sort categorical values `string` etc so we can still use the `>=` comparison

MorganGellert: That's fair. I'll go ahead and extend BinSorter to use a better sort function.

PR #1129, opened 2020-06-01
[Bug] Tooltip formatting

b2kdaman:
* fix percentage formatting
* fix formatting for +12,345.132
* update tests

Signed-off-by: Valentine <vpanch@softserveinc.com>

### DIFF
--- a/src/constants/tooltip.js
+++ b/src/constants/tooltip.js
@@ -25,7 +25,7 @@ export const TOOLTIP_FORMAT_TYPES = {
   PERCENTAGE: 'percentage'
 };
 
-export const TOOLTIP_KEY = 'format';
+export const TOOLTIP_KEY = 'id';
 
 export const TOOLTIP_FORMATS = {
   NONE: {
@@ -89,7 +89,7 @@ export const TOOLTIP_FORMATS = {
   DECIMAL_DELTA: {
     id: 'DECIMAL_DELTA',
     label: '+12,345.432',
-    format: '+,',
+    format: '+,.3f',
     type: TOOLTIP_FORMAT_TYPES.DECIMAL
   },
   DECIMAL_CURRENCY: {
--- a/src/constants/tooltip.js
+++ b/src/constants/tooltip.js
@@ -25,7 +25,7 @@ export const TOOLTIP_FORMAT_TYPES = {
   PERCENTAGE: 'percentage'
 };
 
-export const TOOLTIP_KEY = 'id';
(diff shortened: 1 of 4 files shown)


src/constants/tooltip.js
```diff
@@ -25,7 +25,7 @@ export const TOOLTIP_FORMAT_TYPES = {
   PERCENTAGE: 'percentage'
 };
 
-export const TOOLTIP_KEY = 'format';
+export const TOOLTIP_KEY = 'id';
```
heshan0131, line 28: I would like to keep the format as key, so people can pass in custom fomat without an associated id with the default format

b2kdaman: In some cases we have same formats...

b2kdaman: I was thinking about adding custom formats in next iteration

heshan0131: is it the percent full? I dont like that format anyways. a format should not change the value, it should only change the display of it. I would say we shouldn't have duplicate formats

heshan0131: or we allow format to be either a string or a function so you can have `.3f` or you can have
d => d *100

b2kdaman:
@heshan0131 Idk how it got closed :(

src/constants/tooltip.js
b2kdaman: Good thoughts

b2kdaman: So maybe we can use format as ID and if it's not found we will use it's value as D3 format? Still can't use moment.js format this way. Have to add additional field.
Using format value won't let us have it as a function. Since how then we will store it...

test/node/schemas/vis-state-schema-test.js
```diff
@@ -208,7 +208,7 @@ test('#visStateSchema -> v1 -> save load interaction -> tooltip format', t => {
     tooltip: {
       enabled: true,
       fieldsToShow: {
-        [testCsvDataId]: [{name: 'gps_data.utc_timestamp', format: 'LL'}],
+        [testCsvDataId]: [{name: 'gps_data.utc_timestamp', format: 'DATE_LL'}],
```
heshan0131, line 211: why is the id being saved afterall? Can you confirm it's the `format` key being saved to the reducer?

b2kdaman: Oh sorry!

PR #1166, opened 2020-06-23
Bump deck.gl and luma.gl to v8.2

manassra:
Tested that all sample datasets load properly like in the deployed demo site.


manassra:
Bumped the version to the stable release of v8.2.0.

PR #1187, opened 2020-07-13
[Feat] pass initialUiState to prop

heshan0131:


src/actions/identity-actions.d.ts
```diff
@@ -19,14 +19,16 @@
 // THE SOFTWARE.
 
 import ActionTypes from 'constants/action-types';
+import {UiState} from 'reducers/ui-state-updaters';
 
 export type RegisterEntryUpdaterAction = {
   payload: {
```
macrigiuseppe, line 25: shouldn't we create a type for this?

PR #1199, opened 2020-07-19
New image export approach

### LINKED ISSUE #974: [Help] custom screenshot UI - how do I generate image exports without having the export image modal open?
I'm creating my own custom form popup for saving maps with a custom UI. As part of it, I'm trying to save an image of the visualisation. I have tried a few approaches and have been unable to get desirable results, though I think I'm close. Any guidance would be greatly appreciated.

My question is **How can I get <PlotContainer /> to export and image *without opening the export image modal*?**

Screen Shot 2020-02-26 at 9 04 48 pm

```jsx
const { exportImage } = useSelector(s => s.keplerGl.editor.uiState);
<img
  src={exportImage.imageDataUri}
  width="100%"
  height="auto"
/>
```

**Best approach - path of least resistance, using `injectComponents` and `withState`**
I figured I'm over-thinking it with my other two approaches below, and thought to look at dependency injection again. I remember saying it was possible to pass in custom state via `withState`.

Looking at the source of `<KeplerGl />`, it seems it begins exporting the image when:

https://github.com/keplergl/kepler.gl/blob/4da09e06b4cf84aa6c6a9925b4b4888f31fb3500/src/components/kepler-gl.js#L319-L320

I figure via the `withState` helper, I can inject parts of my own app state in place of this to trick it into exporting when my own `state.app.saveDialogOpen` is set to true. Well, I can't get my own state within [these 'lenses'](https://github.com/keplergl/kepler.gl/blob/master/docs/api-reference/advanced-usages/replace-ui-component.md#pass-custom-component-props), so can I just trick it into thinking the Modal is open? 

```javascript
import { withState, PlotContainerFactory } from "kepler.gl/components";
import ActionTypes from "kepler.gl/dist/constants/action-types";
import { EXPORT_IMAGE_ID } from "kepler.gl/dist/constants/default-settings";

export const CustomPlotContainerFactory = () =>
  withState([
    (reduxState) => ({
      uiState: {
        ...reduxState.uiState,
        currentModal: EXPORT_IMAGE_ID
      }
    })
  ])(PlotContainerFactory());
```

Of course, no love. Also tried using the 'wildcard' mentioned to override the `uiState` passed in:

```js
import { withState, PlotContainerFactory } from "kepler.gl/components";
import { EXPORT_IMAGE_ID } from "kepler.gl/dist/constants/default-settings";

export const CustomPlotContainerFactory = () =>
  withState(
    [], // skip any lenses
    (state: any) => ({ // override uiState prop directly
      uiState: { 
        ...state.keplerGl.editor.uiState, 
        // let this component think the export image modal is open
        currentModal: state.app.saveDialogOpen && EXPORT_IMAGE_ID }})
  )(PlotContainerFactory());
```

Still no love.

Here are the other approaches I tried;

**Approach #1 - use `convertToPng` directly**
I initially tried to wrap my entire component in a custom `<Screenshot />` component that used convertToPng from utils directly, however the layer panel came out in the screenshot and no data showed up, despite there being data on the map.

**Approach #2 - try to re-use `<PlotContainer />`/`<MapContainer>`**
I tried using `PlotContainerFactory` directly, passing in all the props it requires. I continually got the error "Nothing was returned from render". I assume this has something to do with using it outside the `injectComponents`, losing some props it needs from redux.

I can go back through my git history to pull out code examples of these two approaches if you need though I think, based on the docs, they aren't correct.

**How can I get a get <PlotContainer /> to export and image *without opening the export image modal*?**

macrigiuseppe:
This PR changes the export image by the following:
- Rename startExportingImage to setExportingImage
- Clicking on on export image will set uiState.exportImage.exporting to true and open the modal dialog through the new action startExportImage
- Once image export modal dialog loads up the plot-container should be already loaded and the preview should appear quicker
- StartExportImage now accepts params to customize the export image programmatically. 
- StartExportImage also accepts a "center" property that will allow to generate an exported image that capture the center of all data layers

This PR also fixes a double loading bug when showing a map image preview in either export image or save to cloud functionalities. The bug appears the first time you try to export an image or to save the map into the cloud. See gif below

export-image-flow-bugs

This PR fixes #974 

In order to programmatically generate a map image you can call the setExportImageSetting action like the following

```
setExportImageSetting({exporting: true, resolution: ..., ratio ...})
```

and listen for changes to redux store uiState.exportImage.imageDataUri.

Once the map image will be ready, kepler will push the dataUri to uiState.exportImage.imageDataUri


heshan0131:
Toggling settings doesnt trigger regeneration of image
ezgif com-video-to-gif (7)

macrigiuseppe:
> Toggling settings doesnt trigger regeneration of image
> ezgif com-video-to-gif (7)

@heshan0131 fixed the issue

src/components/kepler-gl.js
```diff
@@ -300,6 +296,7 @@ function KeplerGlFactory(
         mapboxApiUrl,
         mapState,
         uiState,
+        visState,
```
heshan0131, line 299: why is the entire visState need to be passed here? The other mapFields items come from visState. Can we pick out exactly what it is needed in visState to pass along?

src/actions/ui-state-actions.d.ts
```diff
@@ -91,16 +91,31 @@ export function removeNotification(
...
+  payload: {
+    ratio?: string;
+    resolution?: string;
+    legend?: string;
+    center?: boolean;
+    exporting?: boolean;
+    processing?: boolean;
+  };
 };
 export function setExportImageSetting(newSetting: {
```
heshan0131, line 103: you can use `setExportImageSetting(newSetting: SetExportImageSettingUpdaterAction['payload'])` as a way to avoid duplicating code

src/reducers/ui-state-updaters.js
```diff
@@ -441,10 +431,35 @@ export const cleanupExportImageUpdater = state => ({
...
+ * @type {typeof import('./ui-state-updaters').startExportingImage}
+ * @public
+ */
+export const startExportingImageUpdater = (state, {payload: options = {}}) => {
+  const imageSettings = {
+    ...options,
+    exporting: true
+  };
+
+  const modifiers = [
```
heshan0131, line 454: the `composet-helpers` functions are designed to compose updater functions with a simplier style
you can just do:
`import {payload_, apply_, compose_} from './composer-helpers';`
then
```js
  return compose_([
    cleanupExportImageUpdater,
    apply_(setExportImageSettingUpdater, payload_(imageSettings)),
    apply_(toggleModalUpdater, payload_(EXPORT_IMAGE_ID)),
  ])(state);
```

src/reducers/ui-state-updaters.js
```diff
@@ -698,3 +713,26 @@ export const setLocaleUpdater = (state, {payload: {locale}}) => ({
...
+/**
+ * Start saving storage  flow
+ * @memberof uiStateUpdaters
+ * @param state
+ * @param mapSaved
+ * @returns {UiState}
+ * @type {typeof import('./ui-state-updaters').startSaveStorage}
+ */
+export const startSaveStorage = (state, {payload: mapSaved}) => {
+  const modifiers = [
```
heshan0131, line 726: same as above

src/components/kepler-gl.js
```diff
@@ -358,14 +353,14 @@ function KeplerGlFactory(
                 <div className="maps" style={{display: 'flex'}}>
                   {mapContainers}
                 </div>
-                {isExporting && (
+                {isExportingImage && (
                   <PlotContainer
                     width={width}
                     height={height}
                     exportImageSetting={uiState.exportImage}
                     mapFields={mapFields}
```
heshan0131, line 361: i realized the `splitMaps` are not exported as 2 maps. digging into it, you just need to pass in `splitMaps` here into the `PlotContainer` then you will be able to export a splitted image:

Screen Shot 2020-07-23 at 11 39 29 PM

src/components/modals/export-image-modal.js
```diff
@@ -64,31 +60,34 @@ const ExportImageModalFactory = () => {
       mapH: PropTypes.number.isRequired,
       exportImage: PropTypes.object.isRequired,
       // callbacks
-      onUpdateSetting: PropTypes.func.isRequired
+      onUpdateImageSetting: PropTypes.func.isRequired
     };
 
     componentDidMount() {
-      this._updateMapDim();
+      // this._updateMapDim();
```
heshan0131, line 67: why is this comment out?

src/components/modals/save-map-modal.js
```diff
@@ -145,6 +145,10 @@ function SaveMapModalFactory() {
       onUpdateImageSetting: nop
     };
 
+    componentWillUnmount() {
```
heshan0131, line 148: shouldn't this be added to `ImageModalContainer` ?

heshan0131:
Added some comments, nothing block except adding the `splitMaps` prop to `PlotContainer` which will fix a bug exporting splitMap images. Please address before merge

src/components/kepler-gl.js
```diff
@@ -323,22 +320,20 @@ function KeplerGlFactory(
...
-            />
-          ));
-
-      const isExporting =
-        uiState.currentModal === EXPORT_IMAGE_ID ||
-        uiState.currentModal === SAVE_MAP_ID ||
-        uiState.currentModal === SHARE_MAP_ID ||
-        uiState.currentModal === OVERWRITE_MAP_ID;
+      const mapContainers = !isSplit ? (
+        <MapContainer key={0} index={0} {...mapFields} mapLayers={null} />
```
manassra, line 324: for consistency, have this be an array of one item?

src/components/plot-container.js
```diff
@@ -133,24 +148,41 @@ export default function PlotContainerFactory(MapContainer) {
...
+
+      const bounds = findMapBounds(mapFields.visState.layers);
+      const width = size.width / (isSplit ? 2 : 1);
+      const height = size.height;
       const scale = this.mapScaleSelector(this.props);
+      const newMapState = {
+        ...mapState,
+        width,
+        height,
+        zoom: mapState.zoom + (Math.log2(scale) || 0)
```
manassra, line 167: nit: instantiate zoom value above and reuse here and in line 177.

src/components/kepler-gl.js
macrigiuseppe: we can get rid of it completely. layers is already part of mapFields

src/components/modals/export-image-modal.js
macrigiuseppe: that's correct, we no longer need it since _updateMapDim is also called on componentDidUpdate

src/components/plot-container.js
macrigiuseppe: this and line 177 shouldn't be reading the same zoom value because on line 167 we read the current map zoom value and on line 177 we are recomputing zoom and center based on data layers middle point

manassra: Ah you are right. Might be worth wrapping it in a function then (feel free to ignore).

abdelsalam-almzayyen:
Hello, I've been having trouble using this and actually getting the imgDataUri to work in my component
```
import store, { RootState } from '../store'; // Import the RootState type from your Redux store

  const imageDataUri = useSelector((state: RootState) => state.keplerGl?.map?.uiState?.exportImage?.imageDataUri);
```

how can I accurately get the state of the map in my component? That would be appreciated.

PR #1288, opened 2020-09-29
[Enhancement] Add preserveLayerOrder to layer merger, enable load data sequantially

heshan0131:


ibgreen (approved):

src/reducers/vis-state-merger.js
```diff
@@ -149,37 +151,85 @@ export function mergeLayers(state, layersToMerge) {
...
+  // perservedOrder ['a', 'b', 'c'];
+  // layerOrder [1, 0, 3]
+  // layerOrderMap ['a', 'c']
+  let layerOrderQueue = currentOrder.map(i => currentLayers[i].id);
+  let newLayers = currentLayers;
+
+  for (const newLayer of newLayers) {
+    // find where to insert it
+    const expectedIdx = preservedOrder.indexOf(newLayer.id);
+    // if cant find place to insert, insert at the font
```
manassra, line 203: type: front*

test/node/reducers/vis-state-merger-test.js
```diff
@@ -285,6 +287,72 @@ test('VisStateMerger.current -> mergeLayers -> toEmptyState', t => {
...
+    stateWithData1.visState.preserveLayerOrder,
+    ['hexagon-2', 'point-0', 'geojson-1'],
+    'shoud preserve layer order'
+  );
+  t.deepEqual(
+    stateWithData1.visState.layers.map(l => l.id),
+    ['geojson-1', 'hexagon-2', 'point-0'],
+    'shoud load 2 layers'
+  );
+  t.deepEqual(stateWithData1.visState.layerOrder, [1, 2, 0], 'layerOrder should be correct');
```
manassra, line 346: Trying to understand how the `preserveLayerOrder` works. Is the `layerOrder` the index of where the layer should go? If so, shouldn't the layerOrder be `[2, 0, 1]`? That is, `geojson-1` should be in index 2, `hexagon-2` should be in index 0, and point-0 should be at index 1?

manassra (approved):
Much needed feature! Thank you for adding it!

PR #1294, opened 2020-10-13
[Chores] Fixed security vulnerabilities and added new factories

macrigiuseppe:


package.json
```diff
@@ -239,16 +239,18 @@
...
     "@loaders.gl/json": "^2.3.0-alpha.9",
     "@loaders.gl/loader-utils": "^2.3.0-alpha.9",
-    "@loaders.gl/polyfills": "^2.3.0-alpha.9"
+    "@loaders.gl/polyfills": "^2.3.0-alpha.9",
+    "dot-prop": "5.1.1",
+    "jpeg-js": "0.4.0",
+    "lodash": "4.17.19",
+    "minimist": "1.2.3",
+    "node-fetch": "^2.6.1",
+    "tough-cookie": "^2.3.0"
```
heshan0131, line 253: The suggested resolution is to fix version to `2.3.0`

heshan0131 (approved):

src/components/side-panel/layer-panel/layer-type-selector.js
```diff
@@ -103,7 +103,7 @@ const DropdownListWrapper = styled.div`
 `;
 
 function LayerTypeListItemFactory() {
-  const LayerTypeListItem = ({value, isTile}) => (
+  return ({value, isTile}) => (
```
heshan0131, line 106: you will probably hahve the `missing displayName` error without the `const LayerTypeListItem=` it's also better for React devTools if we keep the declaration

macrigiuseppe:
> just to double-check, did you run yarn after updated the resolution?
> I can still see multiple version `kind-of` in yarn.lock
> Screen Shot 2020-10-14 at 8 16 57 AM

i think i did. Let me try to re-run it

macrigiuseppe:
@heshan0131 pushed again, all kind-of versions are forced to use 6.0.3, see heer: 
https://github.com/keplergl/kepler.gl/pull/1294/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR8719

PR #1295, opened 2020-10-15
Make datatable header cell overrideable

nicolasmartinrojo:
This need to be used for the statistics header on unfolded-studio

### DIFF
--- a/src/components/common/data-table/index.js
+++ b/src/components/common/data-table/index.js
@@ -26,16 +26,12 @@ import {createSelector} from 'reselect';
 import get from 'lodash.get';
 import debounce from 'lodash.debounce';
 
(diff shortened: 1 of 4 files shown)


src/components/common/data-table/index.js
```diff
@@ -379,97 +376,6 @@ function DataTableFactory(FieldToken) {
...
-    renderHeaderCell = (
-      columns,
-      isPinned,
-      props,
-      toggleMoreOptions,
-      moreOptionsColumn,
-      TokenComponent
-    ) => {
-      // eslint-disable-next-line react/display-name
-      return cellInfo => {
```
heshan0131, line 391: return cellInfo => <HeaderCell {...props} />

PR #1296, opened 2020-10-18
[Bug] Fixed bug when reversing color schema

macrigiuseppe:
Signed-off-by: Giuseppe Macri <macri.giuseppe@gmail.com>


src/reducers/vis-state-updaters.js
```diff
@@ -652,6 +652,9 @@ export const addFilterUpdater = (state, action) =>
  */
 export const layerColorUIChangeUpdater = (state, {oldLayer, prop, newConfig}) => {
   const newLayer = oldLayer.updateLayerColorUI(prop, newConfig);
+  if (prop === 'colorRange') {
```
macrigiuseppe, line 655: this check makes sure we update layer data only when we change the color range property.

heshan0131 (approved):

PR #1305, opened 2020-10-22
extend template for light theme

nicolasmartinrojo:


src/styles/base.js
```diff
@@ -1114,6 +1115,7 @@ export const theme = {
   selectFontWeight,
   selectColorLT,
   selectFontWeightBold,
+  panelTabColor,
```
heshan0131, line 1118: you need to also add `panelTabColor` to  `themeLT` and `themeBS` and have it equals to the value of `subtextColor` in that theme

src/styles/base.js
```diff
@@ -1310,6 +1312,7 @@ export const theme = {
   lineHeight,
   textColor,
   textColorLT,
+  dataTabletextColor,
```
heshan0131, line 1315: same as above

src/styles/base.js
```diff
@@ -1350,6 +1353,7 @@ export const theme = {
   sliderHandleWidth,
   sliderHandleColor,
   sliderHandleTextColor,
+  sliderInactiveBorderColor,
```
heshan0131, line 1356: same as above

src/styles/base.js
nicolasmartinrojo: Nice catch!. as both themes extends the theme variable I've added to themeLT and themeBS just the cases where that value was overwritten

src/styles/base.js
nicolasmartinrojo: sliderHandleTextColor

src/styles/base.js
nicolasmartinrojo: same as above

src/styles/base.js
```diff
@@ -42,6 +42,7 @@ export const labelColorLT = '#6A7485';
 
 export const textColor = '#A0A7B4';
 export const textColorLT = '#3A414C';
+export const dataTabletextColor = textColorLT;
```
heshan0131, line 45: `dataTableTextColor`

src/styles/base.js
```diff
@@ -1494,6 +1498,7 @@ export const themeLT = {
   sliderBarColor: '#A0A7B4',
   sliderBarBgd: '#D3D8E0',
   sliderHandleColor: '#F7F7F7',
```
heshan0131, line 1500: You created 4 new theme names
```
dataTabletextColor
sliderInactiveBorderColor
panelTabColor
dataTabletextColor
```

Stll missing some in `temeLT` and `themeBS`
Be diligent please

nicolasmartinrojo: Both themeLT and ThemeBS extends from `theme`. the variables has been added there(in theme). should I add them again in both themes?

heshan0131: for the ones share same value between dark and light, you don't need to, but for the ones have different values you need to add them to light specifically

nicolasmartinrojo: This change was attended

heshan0131 (approved):

PR #1307, opened 2020-10-26
[Bug] avoid duplicated h3 layer detection (#93)

heshan0131:
* Use only hex field to create h3 layer

### DIFF
--- a/src/components/common/item-selector/item-selector.js
+++ b/src/components/common/item-selector/item-selector.js
@@ -55,8 +55,9 @@ export const StyledDropdownSelect = styled.div.attrs({
 
 const DropdownSelectValue = styled.span`
   color: ${props =>
-    props.hasPlaceholder && props.inputTheme === 'light' ?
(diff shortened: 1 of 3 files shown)


src/components/common/item-selector/item-selector.js
```diff
@@ -55,8 +55,9 @@ export const StyledDropdownSelect = styled.div.attrs({
 
 const DropdownSelectValue = styled.span`
   color: ${props =>
-    props.hasPlaceholder && props.inputTheme === 'light' ?
-    props.theme.selectColorPlaceHolderLT :  props.hasPlaceholder
+    props.hasPlaceholder && props.inputTheme === 'light'
```
macrigiuseppe, line 58: this is also in the previous PR

macrigiuseppe (approved):

src/components/common/item-selector/item-selector.js
heshan0131: yep, will land one and rebase the other. they are automatically corrected by `lint`

PR #1327, opened 2020-11-24
[Bug] Fix 12350 format in tooltip

b2kdaman:
Somehow formatting for 12350 was faulty and misleading from the start

### DIFF
--- a/src/constants/tooltip.js
+++ b/src/constants/tooltip.js
@@ -80,8 +80,8 @@ export const TOOLTIP_FORMATS = {
   },
   DECIMAL_INT: {
     id: 'DECIMAL_INT',
-    label: '12350',
-    format: '.5r',
+    label: '12345 → 12350',
+    format: '.4~r',
     type: TOOLTIP_FORMAT_TYPES.DECIMAL
   },
   DECIMAL_THREE: {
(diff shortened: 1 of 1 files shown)


b2kdaman:
@macrigiuseppe @heshan0131

heshan0131:
As a best practice, can you please create an issue first before creating a PR next time? An issue can be searched in the future and more people can contribute to the discussion

I think the misleading part is the label not the formatted itself. I would not change the ` .4r`, rather update the label. A format once created should behave the same for backward compatibility. So that people's saved map will show the same tooltip even when the software is updated.

the `.4r` does exactly what it should do: rounds to 4 significant digits so that `12345 -> 12350`,  let's use `12345 -> 12350` as label instead of `12350`

One improvement you can make is add `~` and make it `.4~r`  so it trims the empty 0s 

```
0.1 -> 0.1 (instead of 0.100)

```

b2kdaman:
I would not change the .4r
OR
One improvement you can make is add ~ and make it .4~r so it trims the empty 0s
So which one is it?
.4~r would break compatibility. Seems like our "fixes" doesn't really common use cases @thunter

b2kdaman:
> As a best practice, can you please create an issue first before creating a PR next time? An issue can be searched in the future and more people can contribute to the discussion
Yes you're right. I thought that it would be small fixes that would not fire such a discussion 🤣 My bad.

b2kdaman:
@heshan0131 please review

b2kdaman:
@heshan0131 ?

heshan0131 (approved):

PR #1329, opened 2020-11-24
[Bug] fix 1.23k formatting

b2kdaman:
Added custom formatter support

Signed-off-by: Valentine <vpanchin@gmail.com>

### DIFF
--- a/src/constants/tooltip.js
+++ b/src/constants/tooltip.js
@@ -46,7 +46,13 @@ export const TOOLTIP_FORMATS = {
     id: 'DECIMAL_SHORT_COMMA',
     label: '12.3k',
     format: '.3s',
-    type: TOOLTIP_FORMAT_TYPES.DECIMAL
+    type: TOOLTIP_FORMAT_TYPES.DECIMAL,
+    custom: formatter => {
+      return value => {
+        const formatted = formatter(value);
+        return Math.abs(value) < 100 ? formatted.split('.')[0] : formatted;
+      };
+    }
   },
   DECIMAL_PERCENT_FULL_1: {
     id: 'DECIMAL_PERCENT_FULL_1',
--- a/src/constants/tooltip.js
+++ b/src/constants/tooltip.js
@@ -18,6 +18,13 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+export const decimalShortCommaFormatter = formatter => {
+  return value => {
+    const formatted = formatter(value);
+    return Math.abs(value) < 100 ? formatted.split('.')[0] : formatted;
+  };
+};
+
 export const TOOLTIP_FORMAT_TYPES = {
   NONE: 'none',
   DATE: 'date',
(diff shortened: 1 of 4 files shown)


b2kdaman:
@heshan0131 @macrigiuseppe

heshan0131:
Can you explain what this bug is about? why do you need a custom formatted on a default format?

b2kdaman:
> Can you explain what this bug is about? why do you need a custom formater on a default format?

It's for easier values representation. Basially limits value to 3 digits, so we can see values <1000

macrigiuseppe:
> > Can you explain what this bug is about? why do you need a custom formater on a default format?
> 
> It's for easier values representation. Basially limits value to 3 digits, so we can see values <1000

This is too specific for your use case and it may not work with all the other scenarios. I personally don't think we should implement such a specific formatter

thunter009:
> > > Can you explain what this bug is about? why do you need a custom formater on a default format?
> > 
> > 
> > It's for easier values representation. Basially limits value to 3 digits, so we can see values <1000
> 
> This is too specific for your use case and it may not work with all the other scenarios. I personally don't think we should implement such a specific formatter

This format string is useful for scenarios where you have data that is power law distributed and lots of values fall <1000 but a few values are > 1000.

heshan0131:
with the current formatter `.3s`, it formats numbers as decimal notation with an SI prefix, rounded to 3 significant digits.

So expected output should be
```
123 -> 123
1223 -> 1.22k
```

You can perhaps add a `~` to `.3~s` so it trims the 0s

```
9000 -> 9k (instead of 9.00k)
```

Is this something work for you?

test/browser/index.js
```diff
@@ -28,6 +28,9 @@ require('./components');
 // test layers
 require('./layer-tests');
 
+// test formatters
+require('./formatters');
```
heshan0131, line 32: if you only have 1 formatter test, just directly import here

`require('./formatters-tests')`

thunter009:
> with the current formatter `.3s`, it formats numbers as decimal notation with an SI prefix, rounded to 3 significant digits.
> 
> So expected output should be
> 
> ```
> 123 -> 123
> 1223 -> 1.22k
> ```
> 
> You can perhaps add a `~` to `.3~s` so it trims the 0s
> 
> ```
> 9000 -> 9k (instead of 9.00k)
> ```
> 
> Is this something work for you?

This would work perfectly.

Ideally, I'd love to have support for the user entering/editing arbitrary d3 format strings as that feature would obviate the need for building this specific format option. This was a feature in Superset that I absolutely **loved**

test/browser/index.js
b2kdaman: But what if we would have additional formatters in the future?

b2kdaman:
> with the current formatter `.3s`, it formats numbers as decimal notation with an SI prefix, rounded to 3 significant digits.
> 
> So expected output should be
> 
> ```
> 123 -> 123
> 1223 -> 1.22k
> ```
> 
> You can perhaps add a `~` to `.3~s` so it trims the 0s
> 
> ```
> 9000 -> 9k (instead of 9.00k)
> ```
> 
> Is this something work for you?

It actually doesn't
image
 
I was playing with d3 formatter and it has some issues. This is why I had to create this additional "formatter"

b2kdaman:
@heshan0131 sorry for not being clear from the start. This is an old fix tho. 
So the whole problem was that when there are only two digits supplied we would see: `99 -> 99.0`
So this relsolves issue with numbers `< 100`

heshan0131:
depends on which version of `d3-format` you are using. the result of `~s` will different

b2kdaman:
> depends on which version of `d3-format` you are using. the result of `~s` will different

Have you implemented this in kepler/unfolded already?

PR #1330, opened 2020-11-24
[Feature] limit geocoder search area to map viewport

b2kdaman:


heshan0131:
hmm i don't think this is something we want to limit on. Like I always use the geocoder as a way to jump to different location. e.g. map view is San Francisco, I want to go to New York. I would use the geocoder to go there. 

Why do you want to limit the search area to map viewport?

thunter009:
> hmm i don't think this is something we want to limit on. Like I always use the geocoder as a way to jump to different location. e.g. map view is San Francisco, I want to go to New York. I would use the geocoder to go there.
> 
> Why do you want to limit the search area to map viewport?

I guess it depends on how zoomed in your viewport is :) We made this change so that we wouldn't see geocoder results in other countries when trying to find properties in the USA. Perhaps some sort of geocoder config UI is needed?

b2kdaman:
@thunter009 @heshan0131 yes I can add additional control to enable/disable Limit Search to Viewport option.

heshan0131:
> @thunter009 @heshan0131 yes I can add additional control to enable/disable Limit Search to Viewport option.

sure

b2kdaman:
Added control

b2kdaman:
@heshan0131 pls review

src/components/geocoder/geocoder.js
```diff
@@ -49,6 +50,17 @@ export const testForCoordinates = query => {
   return [isValid, Number(tokens[0]), Number(tokens[1])];
 };
 
+export const fixMeridian = bbox => {
```
heshan0131, line 53: please write a function that returns a new `bbox`, instead of directly editing it

src/components/geocoder-panel.js
```diff
@@ -25,14 +25,15 @@ import Processors from 'processors';
 import {FlyToInterpolator} from '@deck.gl/core';
 import KeplerGlSchema from 'schemas';
 import {getCenterAndZoomFromBounds} from 'utils/projection-utils';
+import {isString} from 'lodash';
```
heshan0131, line 28: don't import `lodash`, it's a huge library

src/components/geocoder-panel.js
```diff
@@ -93,8 +94,8 @@ function generateGeocoderDataset(lat, lon, text) {
   };
 }
 
-function isValid(key) {
-  return /pk\..*\..*/.test(key);
+export function isValidMapboxKey(key) {
+  return isString(key) ? /pk\..*\..*/.test(key) : false;
```
heshan0131, line 98: `return typeof key === 'string' && /pk\..*\..*/.test(key)` should be sufficient

src/components/kepler-gl.js
```diff
@@ -209,7 +209,8 @@ export const geoCoderPanelSelector = props => ({
   mapState: props.mapState,
   updateVisData: props.visStateActions.updateVisData,
   removeDataset: props.visStateActions.removeDataset,
-  updateMap: props.mapStateActions.updateMap
+  updateMap: props.mapStateActions.updateMap,
+  limitSearch: props.visState.interactionConfig.geocoder.config.limitSearch
```
heshan0131, line 213: should pass in the entire config
`props.visState.interactionConfig.geocoder.config`

b2kdaman: Gotcha

PR #1350, opened 2020-12-04
[Enhancement] automatically re-project GeoDataFrame to EPSG:4326

### LINKED ISSUE #1209: Automatically reproject GeoDataFrame to WGS84
**Is your feature request related to a problem? Please describe.**
Kepler.gl support only WGS84 geometry and passing GeoDataFrame with any other results in a failure.

**Describe the solution you'd like**
Optimally, kepler.gl should check CRS and automatically reproject to 4326 if necessary.

**Describe alternatives you've considered**
Alternatively, it could raise an informative warning, if you don't want to force reprojection, but I don't see a reason why not.

**Additional context**
Add the following after this line should do the trick.
https://github.com/keplergl/kepler.gl/blob/7eace482ca5512e8a759f45d177f1d2c40435c2e/bindings/kepler.gl-jupyter/keplergl/keplergl.py#L37

```
if copy.crs and not copy.crs == 4326:
    copy = copy.to_crs(4326)
```

It checks if any CRS is assigned and if it is the required one. If not, it reprojects. (`not crs == 4326` instead of `crs != 4326` is intentional to support older pyproj with a bug). The superb thing would be to support and reproject multiple geometry columns, but that is another story.

Happy to do a PR.

martinfleis:
If you pass a GeoDataFrame to Jupyter widget it gets automatically re-projected to EPSG:4326 if `crs` is set and is different. That way user does not have to worry about projections when adding data to a map.

I haven't found a test suite for Jupyter plugin, can you point me to the right place so I can write tests?

Closes #1209

### DIFF
--- a/bindings/kepler.gl-jupyter/keplergl/keplergl.py
+++ b/bindings/kepler.gl-jupyter/keplergl/keplergl.py
@@ -29,6 +29,9 @@ def _gdf_to_dict(gdf):
     Returns:
     - dictionary: a dictionary variable that can be used in Kepler.gl
     '''
+    # reproject to 4326 if needed
+    if gdf.crs and not gdf.crs == 4326:
+        gdf = gdf.to_crs(4326)
 
     # get name of the geometry column
     # will cause error if data frame has no geometry column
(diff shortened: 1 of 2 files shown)


heshan0131 (approved):

heshan0131:
Would be nice to add to documentation data format section explain what type of projection is supported:

https://github.com/keplergl/kepler.gl/tree/master/docs/keplergl-jupyter#3-data-format

PR #1365, opened 2020-12-07
[Bug] Prevent crash in react-map-gl when zoom cannot be calculated

heshan0131:
Signed-off-by: Ilija Puaca <ilija@unfolded.ai>


macrigiuseppe (approved):

ilijapuaca:
The reasons tests were failing is because changes made to center calculation in https://github.com/keplergl/kepler.gl/pull/1314 were undone. We should merge both center and zoom calculation updates together and revert the expected results in the tests.

I can help and fix this up if need be

PR #1370, opened 2020-12-08
[Feat] layer duplication

b2kdaman:
Signed-off-by: Valentine <vpanchin@gmail.com>


src/reducers/vis-state-updaters.js
```diff
@@ -933,6 +933,43 @@ export const removeLayerUpdater = (state, {idx}) => {
...
+/**
+ * duplicate layer
+ * @memberof visStateUpdaters
+ * @type {typeof import('./vis-state-updaters').duplicateLayerUpdater}
+ * @public
+ */
+export const duplicateLayerUpdater = (state, {idx}) => {
+  const {layers, layerData} = state;
+  const layerToDuplicate = cloneDeep(state.layers[idx]);
+  layerToDuplicate.id = `${layerToDuplicate.config.id}_copy`;
```
heshan0131, line 945: this won't work if user copies the same layer twice. use `import {generateHashId} from '../utils/strings';` to get a new hash id for each duplicated layer

b2kdaman: Got it

b2kdaman:
@heshan0131 it's ready

heshan0131:
rewrite in https://github.com/keplergl/kepler.gl/pull/1401

Issue #1374, opened 2020-12-10
[Bug] Clicking on Share Map URL crashes app


davidywang:
**Describe the bug**
Clicking on Share Map URL on any valid map (including sample maps) crashes the app.

**To Reproduce**
1. Upload any valid map (including the sample maps).
2. Click on Share icon on the top right of the side bar.
3. Click on Share Map URL.
4. App crashes. See error in Screenshots:

**Expected behavior**
It should go to the original Share Map URL pop up with the Dropbox and Cartio options.

**Screenshots**
<img width="1268" alt="Screen Shot 2020-12-09 at 9 24 01 PM" src="https://user-images.githubusercontent.com/31499820/101734764-de8f3580-3a64-11eb-8d08-6f020f4506d0.png">

**Desktop:**
 - OS: macOS 10.15.7
 - Browser: Chrome
 - Version: 87.0.4280.88 (Official Build) (x86_64)

**Additional context**
I've also tried this on a Windows 10 laptop using Chrome Version 87 and it has the same error.

Ignatych:
I checked it in different browsers on my Mac and take same problem

toasfand:
Do we have any resolution for this issue?

davidywang:
Just checked it today and it's still not working.

toasfand:
And do we have any idea who will be looking after these issues?

tt87:
Still not working. Same exceptions.

![image](https://user-images.githubusercontent.com/9537310/106125495-4ca2cf80-616d-11eb-984c-bc7c2759529b.png)

heshan0131:
this is fixed now

PR #1399, opened 2021-01-20
fix: save to map provider

VictorVelarde:
Issue #1374

This should fix the error that appears when trying to Save map to a provider (Dropbox / Carto).

The error currently happens at **MapContainer**, lin 506. The presence of `mapControls.mapDraw` is assumed, but that's not the case from this Providers > Save option flow

This small change looks enough to avoid the error and present the popup, but I'm not sure if it needs any other extra check.

### DIFF
--- a/src/components/map-container.js
+++ b/src/components/map-container.js
@@ -503,7 +503,7 @@ export default function MapContainerFactory(MapPopover, MapControl, Editor) {
         transformRequest
       };
 
-      const isEdit = mapControls.mapDraw.active;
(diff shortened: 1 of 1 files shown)


VictorVelarde:
Would you please @heshan0131 have a quick look into this one? Thx!

VictorVelarde:
Hi, any chance to review this small PR soon? Right now the whole panel is failing due to this issue.

heshan0131 (approved):

VictorVelarde:
Thx Shan!

PR #1401, opened 2021-01-25
[Feat] Duplicate layer and add layer from config

heshan0131:


src/reducers/vis-state-updaters.js
```diff
@@ -896,6 +898,7 @@ export const removeFilterUpdater = (state, action) => {
  */
 export const addLayerUpdater = (state, action) => {
   const defaultDataset = Object.keys(state.datasets)[0];
+  if (props.type && state.layerClasses[props.type])
```
ilyabo, line 901: This doesn't look right. Was this line added by mistake?

src/reducers/vis-state-updaters.js
```diff
@@ -937,6 +940,69 @@ export const removeLayerUpdater = (state, {idx}) => {
...
+  const {layers} = state;
+  const original = state.layers[idx];
+  if (!original) {
+    Console.warn(`layer.${idx} is undefined`);
+    return state;
+  }
+  let newLabel = `Copy of ${original.config.label}`;
+  let postfix = 0;
+  while (layers.find(l => l.config.label === newLabel)) {
+    newLabel = `${newLabel} ${postfix++}`;
```
ilyabo, line 959: This will probably generate something like `Copy of My layer 0 1 2 3 4 5 6` and so on: starting with 0 and appending the next number to it. It'd be probably better to define some kind of template in a separate constant like `Copy of ${layerName}${postfix}` and then replace the variables in the loop.

heshan0131: good catch

ilyabo (approved):
lgtm 👍

PR #1446, opened 2021-03-29
[jupyter] Add a "Save Map" button to PanelHeader

chrisgervang:
Screen Shot 2021-03-29 at 10 31 36 AM

### DIFF
--- a/bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
+++ b/bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
@@ -22,11 +22,14 @@ import {PanelHeaderFactory, Icons, withState} from 'kepler.gl/components';
 import {toggleModal} from 'kepler.gl/actions';
 import React from 'react';
 import {IntlProvider} from 'react-intl';
+import {visStateLens} from 'kepler.gl/reducers';
+import {exportJson} from 'kepler.gl/dist/utils/export-utils'
 
 const KEPLER_DOC = 'https://docs.kepler.gl/docs/keplergl-jupyter';
 
 export function CustomPanelHeaderFactory() {
   const PanelHeader = PanelHeaderFactory();
+  const defaultActionItems = PanelHeader.defaultProps.actionItems;
 
   const actionItems = props => [
     {
@@ -36,6 +39,12 @@ export function CustomPanelHeaderFactory() {
       blank: true,
       tooltip: 'tooltip.documentation',
       onClick: () => {}
+    },
+    {
+      ...defaultActionItems.find(item => item.id === 'save'),
+      label: null,
+      tooltip: 'Save Map',
+      onClick: () => exportJson({visState: props.visState})
     }
   ];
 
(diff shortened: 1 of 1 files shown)


bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
```diff
@@ -22,11 +22,14 @@ import {PanelHeaderFactory, Icons, withState} from 'kepler.gl/components';
 import {toggleModal} from 'kepler.gl/actions';
 import React from 'react';
 import {IntlProvider} from 'react-intl';
+import {visStateLens} from 'kepler.gl/reducers';
+import {exportJson} from 'kepler.gl/dist/utils/export-utils'
```
heshan0131, line 26: you should add export `exportJson` from `utils/index.js` them import it here with
`import {exportJson} from 'kepler.gl/utils'`

chrisgervang: None of `kepler.gl/utils` is exported directly. It's missing from the npm [files](https://github.com/keplergl/kepler.gl/blob/master/package.json#L76). Should I add that too?

bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
```diff
@@ -36,6 +39,12 @@ export function CustomPanelHeaderFactory() {
       blank: true,
       tooltip: 'tooltip.documentation',
       onClick: () => {}
+    },
+    {
+      ...defaultActionItems.find(item => item.id === 'save'),
+      label: null,
+      tooltip: 'Save Map',
+      onClick: () => exportJson({visState: props.visState})
```
heshan0131, line 47: why only export `visState`? why not the entire kepler.gl state?

bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
heshan0131: urr right, it needs to be add to `package.json` `"files"` field

heshan0131: I can do that and cut a new release

chrisgervang: Thanks. Can you update the version the jupyter binding uses at the same time? It's still on `2.4.0`.

bindings/kepler.gl-jupyter/js/lib/keplergl/components/panel-header.js
chrisgervang: The export function only uses `visState`. However, the only reason I did it like this is because I copied a `visStateLens` line from elsewhere. Is this [keplerGlReducerCore](https://github.com/keplergl/kepler.gl/blob/master/src/reducers/index.js#L27) (default export) the entire store as a lens?

igorDykhta:

PR #1469, opened 2021-05-02
[Enhancement] add Japanese translation

sw1227:
I've added a new language "Japanese" (ja), following the documentation below:
https://github.com/keplergl/kepler.gl/tree/master/docs/api-reference/localization

### DIFF
--- a/src/localization/locales.js
+++ b/src/localization/locales.js
@@ -23,7 +23,8 @@ export const LOCALES = {
   fi: 'Suomi',
   pt: 'Português',
   es: 'Español',
-  ca: 'Català'
+  ca: 'Català',
+  ja: 'Japanese'
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/localization/locales.js
```diff
@@ -23,7 +23,8 @@ export const LOCALES = {
   fi: 'Suomi',
   pt: 'Português',
   es: 'Español',
-  ca: 'Català'
+  ca: 'Català',
+  ja: 'Japanese'
```
heshan0131, line 27: `ja: 日本語`

heshan0131 (approved):

src/localization/locales.js
sw1227: Thanks!
I fixed it in the second commit.

PR #1478, opened 2021-05-12
[Enhancement] Elevation zoom factor toggle

igorDykhta:
- adds a switch to toggle elevation zoom factor to layers that support height.

image


ibgreen (approved):
Wow - 22 files...

heshan0131 (approved):

PR #1483, opened 2021-05-18
[Enhancement] Support elevation in Icon layer

igorDykhta:
- optional elevation attribute for Icon layer

image


heshan0131 (approved):

heshan0131:
Same, could you include a screenshot of the column selector UI? and a rendered example?

Issue #1515, opened 2021-06-24
[Bug] Kepler removes layers passed directly to DeckGL instance via DeckGLProps 


punov:
**Describe the bug**
There's a chance to pass props directly to DeckGL instance, via DeckGLProps property of <KeplerGL /> component.
But layers that we pass to DeckGLProps get erased by Kepler.

**To Reproduce**
1. Create array with one custom layer (MVTLayer, for example), pass it to KeplerGL via props, using DeckGLProps.
```
import React from 'react';
import KeplerGl from 'kepler.gl';
import { MVTLayer } from '@deck.gl/geo-layers';

const Map = () => {
  const layer = new MVTLayer({ ... });
  return (
    <KeplerGl
      id="map"
      mapboxApiAccessToken={token}
      deckGlProps={{ [ layer ] }}
      width={width}
      height={height}
    />
  );
};
export default Map;
```
2. **Layer is visible on initial render, but removed by the second render, after any user action (like mouse over, for example).**
export default Map;

**Expected behavior**
Layers passed to DeckGL instance rendered once and keep being exist for the entire component session.

**Additional context**
We see that layers passed to DeckGL component via DeckGLProps fully ignored, 
map-container.js file, lines 420-425:
```
      return (
        <DeckGL
          {...this.props.deckGlProps}
          viewState={mapState}
          id="default-deckgl-overlay"
          layers={deckGlLayers} 
```

FdeFabricio:
Does passing deck's MVTLayer as props allow you to use Kepler's interface to add other mvt layers?
I implemented a custom modal to load MVT data from URL and wanted to add MVTLayers to the map.

heshan0131:
it should work as long as tooltip is disabled. kepler.gl will try to pass layerHoverInfo from the layer hover call back deck.gl passes. let me know if you encounters any error, I can help fix it on the  kepler.gl side

PR #1516, opened 2021-06-24
Initialise DeckGL layers from props

punov:
PR to address the DeckGLProps layers issue: https://github.com/keplergl/kepler.gl/issues/1515


heshan0131:
This is a interesting way to pass additional layers to kepler. Though you do need to make sure your layer.data is constructed in a way that layer onHover event will trigger the correct tooltip to be rendered. 

In other words, if you pass in your own deck.gl layers, you should disable picking...

heshan0131 (approved):

PR #1533, opened 2021-07-06
[Bug] Geojson layer is not updated when dataset updated

delfrrr:
In current implementation when `addDataToMap` is called second time for the same `dataset.info.id`, `geojson-layer` will not  update correctly. Because `calculateDataAttribute` in GeoJSON layer require precalculated `this.dataToFeature` property which calculated in `updateLayerMeta`. So If metadata did not change (same columns but different rows), `updateLayerMeta` will not be called and layer will not update correctly.

Given that `updateLayerMeta` takes `allData` as param, it will be logical to call it when data updates.

### DIFF
--- a/src/layers/base-layer.js
+++ b/src/layers/base-layer.js
@@ -846,12 +846,12 @@ class Layer {
     this.meta = {...this.meta, ...meta};
   }
 
-  getDataUpdateTriggers({filteredIndex, id}) {
+  getDataUpdateTriggers({filteredIndex, id, allData}) {
     const {columns} = this.config;
 
     return {
-      getData: {datasetId: id, columns, filteredIndex},
-      getMeta: {datasetId: id, columns},
+      getData: {datasetId: id, allData, columns, filteredIndex},
+      getMeta: {datasetId: id, allData, columns},
       ...(this.config.textLabel || []).reduce(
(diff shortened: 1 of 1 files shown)


src/layers/base-layer.js
```diff
@@ -873,7 +873,7 @@ class Layer {
     const dataUpdateTriggers = this.getDataUpdateTriggers(layerDataset);
     const triggerChanged = this.getChangedTriggers(dataUpdateTriggers);
 
-    if (triggerChanged.getMeta) {
+    if (triggerChanged.getMeta || triggerChanged.getData) {
```
heshan0131, line 876: for this case, you should change `getDataUpdateTriggers`  in base-layer, and add `allData` to the return,

```js
 return {
    getData: {datasetId: id, allData, columns, filteredIndex},
    getMeta: {datasetId: id, allData, columns},
...
}
```

delfrrr: sounds good, will do

delfrrr: @heshan0131 updated as requested

heshan0131 (approved):

delfrrr:
@heshan0131 so what are the next steps? is there anything I can do t get this merged? I guess you need to approve some workflows to run.

Issue #1488, opened 2021-05-20
[Bug][Jupyter Widget] localStorage causing SecurityError


chrisgervang:
**Describe the bug**
When using the jupyter kepler plugin in Databricks (which is a hosted jupyter notebook), there is an intermittent localStorage issue causing the kepler widget to break on load.

**To Reproduce**
There isn't a clear way to reproduce this issue. It seems to happen every couple of weeks. We've noticed clearing cookies and cache in Chrome sometimes resolves the issue, it has also been reported that it has unintentionally been resolved after a couple days (users don't know why the issue goes away when it does).

**Expected behavior**
When this error is experienced, I'd expect it to be caught and gracefully handled. While performing a code review for localStorage usage, it seems to only be used in the widget html template for google analytics. Perhaps we can provide a config flag to disable google analytics if it's the issue.

**Screenshots**
<img width="1017" alt="Screen Shot 2021-05-19 at 3 25 07 PM" src="https://user-images.githubusercontent.com/2461547/118894267-c267ae00-b8b8-11eb-98b9-9ba29477cc40.png">

**Environment (please complete the following information):**
 - Python version: python3
 - keplergl Widget version: latest (0.2.2)

**Additional context**
The widget only appears to use localStorage for google analytics.
https://github.com/keplergl/kepler.gl/blob/14c35fc048a745faab0c6770cab7a4625ccedda3/bindings/kepler.gl-jupyter/js/template/keplergl-html.ejs#L102

macrigiuseppe:
@chrisgervang we can look into it but in the meanwhile you can edit the export html file as workaround

chrisgervang:
We haven't been able to work around this since the kepler widget is embedded in a iframe within serialized jupyter notebook exported to an html file, rather than simply the kepler html file on its own. I attempted to work around all these layers, but our existing files can't easily be fixed.

The issue seems to be that in our ipython notebook environment (provided by databricks), every `display()` call sandboxes the widget in an iframe. Since kepler is in an iframe, cross-domain issues [like this](https://developers.google.com/analytics/devguides/collection/analyticsjs/cross-domain) can occur.

nanshanli:
For us, the same issue consistently arises within Databricks when the "block third party cookies" setting in the browser is enabled. As that setting is controlled by the organisation administrator on our corporate network, our only workaround thus far has been to remove the google analytics script in the binding (similar to as in your PR).

PR #1543, opened 2021-07-13
[enhancement] Add iframe_safe option to jupyter binding

chrisgervang:
For #1488


igorDykhta:

PR #1555, opened 2021-07-21
[feature] Abstract Data Container

igorDykhta:
Changes: 
- DataContainerInterface defines how a data container should look like.
- RowDataContainer stores data internally as any[][] (the same way the data was stored with allData).
- IndexedDataContainer can be used to filter the parent data container.

- KeplerTable.allData is substituted with KeplerTable.dataContainer. Most of the places that expected any[][] now expect a dataContainer.
- Data literal objects like {data: allData[index], index} now look like {index}
- Data container is provided to all accessors as a separate argument: 
  in layers: 
`(config.columns) => (dataContainer) => (dataElement) => []`
  in filterValueAccessor:
` (dataContainer) => (getData, getIndex) => (dataElement) => []`

- [ ] upgrade guide as a follow-up

### DIFF
--- a/src/layers/geojson-layer/geojson-layer.js
+++ b/src/layers/geojson-layer/geojson-layer.js
@@ -61,7 +61,8 @@ export const geojsonVisConfigs = {
 };
 
 export const geoJsonRequiredColumns = ['geojson'];
-export const featureAccessor = ({geojson}) => d => d[geojson.fieldIdx];
+export const featureAccessor = ({geojson}) => dc => d => dc.valueAt(d.index, geojson.fieldIdx);
+
 // access feature properties from geojson sub layer
 export const defaultElevation = 500;
 export const defaultLineWidth = 1;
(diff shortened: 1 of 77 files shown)


src/layers/geojson-layer/geojson-layer.js
```diff
@@ -73,7 +74,7 @@ export default class GeoJsonLayer extends Layer {
 
     this.dataToFeature = [];
     this.registerVisConfig(geojsonVisConfigs);
-    this.getPositionAccessor = () => featureAccessor(this.config.columns);
+    this.getPositionAccessor = dataContainer => featureAccessor(this.config.columns)(dataContainer);
```
heshan0131, line 77: This should just be
```js
this.getPositionAccessor = featureAccessor(this.config.columns);

```

heshan0131, line 77: delete the other `getPositionAccessor` in this layer class? that was an old implementation

src/schemas/dataset-schema.js
```diff
@@ -74,6 +74,12 @@ class DatasetSchema extends Schema {
   key = 'dataset';
 
   save(dataset) {
+    if (dataset.dataContainer) {
+      dataset = {
+        ...dataset,
```
heshan0131, line 79: `dataset` is a class, this will make it a plain object. you can use the `copyTableAndUpdate` from `kepler-table.js` to avoid directly modifying `dataset

heshan0131 (approved):

PR #1570, opened 2021-08-11
Added Russian localization

sham-sr:


heshan0131:
Thanks for this work. Can you add one more entry to
https://github.com/keplergl/kepler.gl/blob/master/docs/api-reference/localization/README.md

in the Current supported languages are: ... table?

heshan0131 (approved):

sham-sr:
Ok I added this. Sorry for the carelessness.

heshan0131 (approved):

PR #1588, opened 2021-08-29
[Feat] Add MapsLayoutFactory for custom split map layouts

heshan0131:
- Add MapsLayoutFactory for custom split map layouts
- Avoid manual calculation of the sizes of the map-containers and positioning them with CSS instead
- Add observeDimensions and useDimensions to allow components to listen to their size changes
- Fixed the issue with the HistogramPlot not being updated by adding a resize observer to RangeSlider
- Fix Crash on Double-click in dual map 
- Fix: Pinning a tooltip moves it to the top left of the screen
- Changed width > 0 || height > 0 to width > 0 && height > 0
- Fix formatting
- Fix: Map layers can't be moved on top


ilyabo (approved):
There was also a fix which came later for blocking the input while swiping. We should probably merge it too

heshan0131:
> There was also a fix which came later for blocking the input while swiping. We should probably merge it too

is that in kepler or studio?

ilyabo:
@heshan0131 Ah indeed, it [was a Studio fix](https://github.com/UnfoldedInc/platform/pull/1664)

Issue #1585, opened 2021-08-23
Display legend by default on a kepler gl point map


elolelo:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I export my point map made  using kepler gl to an interactive html file. This interactive html file should have a legend (colour key) visible by default. What I mean is I shouldn't click on the show legend button to see the meaning of colours of the points on the map.

**Describe the solution you'd like**
I'd like that when I export an html file of a map to instantly see the legend  without clicking the legend button. 

**Describe alternatives you've considered**
Instead of exporting the map, I have tried hosting it in one of the cloud services but the outcome is still the same. Then I tried using react instead but the problem is I will have tom create 10 react  map apps because I want 10 maps.
 
**Additional context**
I would like to have (as the below image shows ) that legend fixed and displaying by default on the exported html kepler.gl map
<img width="2048" alt="Screenshot 2021-08-23 at 15 19 52" src="https://user-images.githubusercontent.com/32453496/130454297-75e1ebec-5de0-4619-a116-0bded3f026bb.png">

elolelo:
@lixun910 Thank you very much - this has worked.

PR #1598, opened 2021-09-01
[Enhancemen] Export Map (maplegend and locale) as HTML 

lixun910:
This PR is for the issue https://github.com/keplergl/kepler.gl/issues/1585 (display map legend on an exported map). There is a workaround for this issue that requires user to update the exported HTML file. However,  this seems like a reasonable feature enhancement request for the “Export Map as HTML” feature:
1. when exporting a map to a HTML file, uiState of mapLegend and map locale should be saved in config.
2. when loading config via `addDataToMapUpdater()`, the uiState of mapLegend and map locale should be loaded and used as default.

What's added and changed in this PR:

1. Add a new `uiStateSchema` class in _src/schemas/ui-state-schema.js_
2. Add above `uiStateSchema` in `reducerSchema` in _src/schemas/schema-manager.js_, so the meta data of `locale` and `mapControls:mapLegend` in `uiState` can be saved and loaded via Schema Manager
3. Since the mapLegend and locale in uiState can be added in JSON configuration, the `addDataToMap` action should be updated (_src/reducers/combined-updaters.js_) so customized `mapLegend` (on/off) and `locale` can be loaded as default.

4. Test case `#uiStateSchema -> v1 -> save load uiState` in _test/node/schemas/ui-state-schema-test.js_
5. Test case `#composerStateReducer - addDataToMapUpdater: mapLegend` in _test/node/reducers/composer-state-test.js_
6. Test case `#composerStateReducer - addDataToMapUpdater: locale` in _test/node/reducers/composer-state-test.js_

Example: an exported map with `Legend` switched ON and `locale` with Spanish(ES)
Screen Shot 2021-08-31 at 4 08 40 PM

### DIFF
--- a/src/reducers/combined-updaters.js
+++ b/src/reducers/combined-updaters.js
@@ -20,9 +20,7 @@
 
 import {
   toggleModalUpdater,
-  toggleMapControlUpdater,
-  loadFilesSuccessUpdater as uiStateLoadFilesSuccessUpdater,
-  setLocaleUpdater
+  loadFilesSuccessUpdater as uiStateLoadFilesSuccessUpdater
 } from './ui-state-updaters';
 import {
   updateVisDataUpdater as visStateUpdateVisDataUpdater,
@@ -30,6 +28,7 @@ import {
 } from './vis-state-updaters';
 import {receiveMapConfigUpdater as stateMapConfigUpdater} from './map-state-updaters';
 import {receiveMapConfigUpdater as styleMapConfigUpdater} from './map-style-updaters';
+import {receiveMapConfigUpdater as uiStateReceiveMapConfigUpdater} from './ui-state-updaters';
 import {findMapBounds} from 'utils/data-utils';
 import {isPlainObject} from 'utils/utils';
 import {filesToDataPayload} from 'processors/file-handler';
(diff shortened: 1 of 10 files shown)


src/reducers/combined-updaters.js
```diff
@@ -170,6 +178,18 @@ export const addDataToMapUpdater = (state, {payload}) => {
 
     pick_('uiState')(apply_(toggleModalUpdater, payload_(null))),
 
+    if_(
```
heshan0131, line 181: stead of directly call `pick_('uiState')` in combined updaters, you should add a `receiveMapConfigUpdater` to `ui-state-updaters`,  and import it here, call it with 
```
import {receiveMapConfigUpdater as uiStateReceiveMapConfigUpdater} from './map-style-updaters';
...
pick_('mapStyle')(apply_(styleMapConfigUpdater, payload_({config: parsedConfig, options}))),
    
pick_('uiState')(apply_(uiStateReceiveMapConfigUpdater, payload_({config: parsedConfig, options}))),

pick_('uiState')(apply_(uiStateLoadFilesSuccessUpdater, payload_(null))),
```

we have `receiveMapConfigUpdater` in each of the sub-reducers, to handle merging config into state. We should follow the pattern with uiState too

lixun910: Thanks, @heshan0131! I added `receiveMapConfigUpdater` in `ui-state-updaters`, specifically for 'mapLegend' and 'locale'. Are there any other things I should consider in this function when merging received uistate? Thanks!

PR #1612, opened 2021-09-13
[Enhancement] Render map control tooltip with TippyTooltip

heshan0131:
Close map drawing when clicking on map legend


ilyabo (approved):
👍

PR #1627, opened 2021-09-21
[Enhancement] Allow changing MAX_DEFAULT_TOOLTIPS

### LINKED ISSUE #1477: Kepler tooltip by default shows 5 fields. How can that be changed to include more fields in the tooltip.
2D8AD8B3-DB8A-4BCA-9177-30D07A671717_4_5005_c

Here it is set as 5, If I am following https://github.com/keplergl/kepler.gl/blob/14c35fc048a745faab0c6770cab7a4625ccedda3/src/templates/export-map-html.js to insert my data into Kepler, how can I change this default number of fields in the tooltip?
https://github.com/keplergl/kepler.gl/blob/14c35fc048a745faab0c6770cab7a4625ccedda3/src/constants/default-settings.js#L649

Rikuoja:
Resolve #1477


test/node/utils/interaction-utils-test.js
```diff
@@ -131,7 +131,7 @@ test('interactionUtil -> autoFindTooltipFields', t => {
   };
 
   t.deepEqual(
-    findFieldsToShow({fields, id: 'test'}),
+    findFieldsToShow({fields, id: 'test', maxDefaultTooltips: 5}),
```
heshan0131, line 134: I would change the value here to be `maxDefaultTooltips: 3` to test wether the custom value is passed in

heshan0131 (approved):

test/node/utils/interaction-utils-test.js
Rikuoja: Good point! Will fix this.

PR #1630, opened 2021-09-24
[Feat] enable time range filter also for date fields

### LINKED ISSUE #151: Enable playback for date only series
Thanks for the awesome lib. I believe the following use case would be a valuable addition for the playback feature: 

Given a dataset including a date only series (=without times), choosing the date column as a filter should enable the playback feature so that the useless timestamp `00:00:00`, which has to be added to enable playback, can be omitted.

This is particularly relevant if the dataset is fairly big (+2.5 million records) as the file size can be reduced dramatically.

Joonalai:
Resolve #151.

### DIFF
--- a/src/utils/filter-utils.js
+++ b/src/utils/filter-utils.js
@@ -330,7 +330,6 @@ export function getFilterProps(field, fieldDomain) {
       };
 
     case ALL_FIELD_TYPES.string:
-    case ALL_FIELD_TYPES.date:
       return {
         ...filterProps,
         type: FILTER_TYPES.multiSelect,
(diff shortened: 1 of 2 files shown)


Joonalai:
@heshan0131 I didn't fix the tests yet so it seems that the functionality of selecting few separate dates is lost after this PR. Do you have any ideas of how to go forward with this? Should we perhaps allow use to choose which filter type to use?

heshan0131:
> @heshan0131 I didn't fix the tests yet so it seems that the functionality of selecting few separate dates is lost after this PR. Do you have any ideas of how to go forward with this? Should we perhaps allow use to choose which filter type to use?

We are working on another feature here to allow different filter options, so user can choose a different filter widget based on field type. 

For date field we would add an option to filter by `Select Dates` or `Date range`. So that we can display different UI upon selection. This will still support the select dates option, and it also won't break the old saved maps which had the date filtered by multiple dates

Screen Shot 2021-09-24 at 9 19 45 AM

Joonalai:
Oh, I understand. I will close this for now.

PR #1631, opened 2021-09-24
[Feat] add custom value to layer slider

morphoix:
Signed-off-by: morphoix <morphoix@gmail.com>

One small problem is that UI changes render not after input but after toggling the map or slider, it must be smth with life cycles of the components but I'm not sure what exactly :) 

Screenshot from 2021-09-24 18-52-52
Screenshot from 2021-09-24 18-52-35
Screenshot from 2021-09-24 18-51-49
Screenshot from 2021-09-24 18-51-04
Screenshot from 2021-09-24 18-50-00

### DIFF
--- a/src/components/side-panel/layer-panel/vis-config-slider.js
+++ b/src/components/side-panel/layer-panel/vis-config-slider.js
@@ -18,12 +18,47 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
+import React, {useCallback, useState} from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import {PanelLabel, SidePanelSection} from 'components/common/styled-components';
 
 import RangeSliderFactory from 'components/common/range-slider';
 import {FormattedMessage} from 'localization';
+import {Checkbox} from 'components';
+
+const InputWrapper = styled.div`
+  display: flex;
+  line-height: 12px;
+  margin-bottom: 12px;
+`;
+
+const CustomInputWrapper = styled.div`
+  display: flex;
+`;
+
+const CustomInputLabel = styled.label`
+  color: ${props => props.theme.textColor};
+  font-weight: 500;
+  letter-spacing: 0.2px;
+  font-size: ${props => props.theme.layerConfigGroupLabelLabelFontSize};
+  padding-right: 15px;
+
+  &:last-child {
+    position: absolute;
+    right: 0;
+    padding: 0;
+  }
+`;
+
+const RangeInput = styled.input`
+  ${props => props.theme.input};
+  font-size: ${props => props.theme.sliderInputFontSize};
+  width: 44px;
+  overflow: auto;
+  height: 20px;
+  margin-top: 5px;
+`;
 
 const propTypes = {
   layer: PropTypes.object.isRequired,
@@ -50,31 +85,124 @@ export default function VisConfigSliderFactory(RangeSlider) {
     disabled,
     onChange,
     inputTheme
-  }) => (
-    <SidePanelSection disabled={Boolean(disabled)}>
-      {label ? (
-        <PanelLabel>
-          {typeof label === 'string' ? (
-            <FormattedMessage id={label} />
-          ) : typeof label === 'function' ? (
-            <FormattedMessage id={label(config)} />
+  }) => {
+    const [custom, setCustom] = useState(false);
+    const [values, setValues] = useState({
+      value: '',
+      value0: '',
+      value1: ''
+    });
+
+    const updateField = e => {
+      setValues({
+        ...values,
+        [e.target.name]: e.target.value
+      });
+    };
+
+    const onKeyPress = useCallback(
+      e => {
+        if (e.key === 'Enter') {
+          const val = e.target.value;
+          switch (e.target.name) {
+            case 'value':
+              config.visConfig[property] = range[1] = Number(val);
+              break;
+            case 'value0':
+              config.visConfig.radiusRange[0] = range[0] = Number(val);
+              break;
+            case 'value1':
+              config.visConfig.radiusRange[1] = range[1] = Number(val);
+              break;
+            default:
+              break;
+          }
+        }
+      },
+      [config.visConfig, property, range]
+    );
+
+    const renderCustomInput = () => {
+      return (
+        <CustomInputWrapper>
+          {isRanged ? (
+            <InputWrapper>
+              <CustomInputLabel>
+                min
+                <RangeInput
+                  type="number"
+                  name="value0"
+                  value={values.value0}
+                  onChange={updateField}
+                  onKeyPress={onKeyPress}
+                />
+              </CustomInputLabel>
+              <CustomInputLabel>
+                max
+                <RangeInput
+                  type="number"
+                  name="value1"
+                  value={values.value1}
+                  onChange={updateField}
+                  onKeyPress={onKeyPress}
+                />
+              </CustomInputLabel>
+            </InputWrapper>
           ) : (
-            <FormattedMessage id={`property.${property}`} />
+            <InputWrapper>
+              <RangeInput
+                type="number"
+                name="value"
+                value={values.value}
+                onChange={updateField}
+                onKeyPress={onKeyPress}
+              />
+            </InputWrapper>
           )}
-        </PanelLabel>
-      ) : null}
-      <RangeSlider
-        range={range}
-        value0={isRanged ? config.visConfig[property][0] : range[0]}
-        value1={isRanged ? config.visConfig[property][1] : config.visConfig[property]}
-        step={step}
-        isRanged={Boolean(isRanged)}
-        onChange={value => onChange({[property]: isRanged ? value : value[1]})}
-        inputTheme={inputTheme}
-        showInput
-      />
-    </SidePanelSection>
-  );
+        </CustomInputWrapper>
+      );
+    };
+
+    return (
+      <SidePanelSection disabled={Boolean(disabled)}>
+        {label ? (
+          <PanelLabel>
+            {typeof label === 'string' ? (
+              <FormattedMessage id={label} />
+            ) : typeof label === 'function' ? (
+              <FormattedMessage id={label(config)} />
+            ) : (
+              <FormattedMessage id={`property.${property}`} />
+            )}
+          </PanelLabel>
+        ) : null}
+
+        <InputWrapper>
+          <CustomInputLabel>custom input</CustomInputLabel>
+          <Checkbox
+            id={`property.${property}`}
+            checked={custom}
+            onChange={() => setCustom(!custom)}
+          />
+        </InputWrapper>
+
+        {!custom ? (
+          <RangeSlider
+            range={range}
+            value0={isRanged ? config.visConfig[property][0] : range[0]}
+            value1={isRanged ? config.visConfig[property][1] : config.visConfig[property]}
+            step={step}
+            isRanged={Boolean(isRanged)}
+            onChange={value => onChange({[property]: isRanged ? value : value[1]})}
+            inputTheme={inputTheme}
+            showInput
+          />
+        ) : (
+          renderCustomInput()
+        )}
+      </SidePanelSection>
+    );
+  };
 
   VisConfigSlider.propTypes = propTypes;
 


src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
-      {label ? (
-        <PanelLabel>
-          {typeof label === 'string' ? (
-            <FormattedMessage id={label} />
-          ) : typeof label === 'function' ? (
-            <FormattedMessage id={label(config)} />
+  }) => {
+    const [custom, setCustom] = useState(false);
+    const [values, setValues] = useState({
+      value: config.visConfig[property],
```
heshan0131, line 91: you don't need 3 values here, `customValue` should be either a number or a [number, number] if slider is a range

```
    const [customValue, setCustomValues] = useState({
      value: config.visConfig[property]
    });
```

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
+            range = value;
+            break;
+          default:
+            break;
+        }
+        onChange({[property]: value});
+      }
+    };
+
+    const renderCustomInput = () => {
```
heshan0131, line 124: don't create react compoent inside a component move it outside 
```js
const CustomInput = ({isRanged, value, onChange}) => (
      <CustomInputWrapper>
        {isRanged ? (
          <InputWrapper>
            <CustomInputLabel>
              min
              <RangeInput type="number" name="value0" value={value[0]} onChange={onChange} />
            </CustomInputLabel>
            <CustomInputLabel>
              max
              <RangeInput type="number" name="value1" value={value[1]} onChange={onChange} />
            </CustomInputLabel>
          </InputWrapper>
        ) : (
          <InputWrapper>
            <RangeInput type="number" name="value" value={value} onChange={onChange} />
          </InputWrapper>
        )}
      </CustomInputWrapper>
    );
)

```

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
+
+    const updateField = e => {
+      setValues({
+        ...values,
+        [e.target.name]: e.target.value
+      });
+    };
+
+    const onKeyPress = e => {
+      if (e.key === 'Enter') {
```
heshan0131, line 104: use `e.keyCode` here

```
   switch (e.keyCode) {
        case KeyEvent.DOM_VK_ENTER:
        case KeyEvent.DOM_VK_RETURN:
``

you can import KeyEvent from 
```
import KeyEvent from 'constants/keyevent';
```

File: src/components/side-panel/layer-panel/vis-config-slider.js (hunk 1 above)
Comment (heshan0131) [reply src/components/side-panel/layer-panel/vis-config-slider.js:91]: Also if you use `customValue` to temporarily store component state value, and `onChange` will update the actual value in the reducer state, you should update `customValue` when value in the reducer state changes. An `useEffect` will do

```js
const value = config.visConfig[property];
const [customValue, setCustomValues] = useState(value);
useEffect(() => {
   setCustomValues(value);
}, [value]);
```

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
-    <SidePanelSection disabled={Boolean(disabled)}>
-      {label ? (
-        <PanelLabel>
-          {typeof label === 'string' ? (
-            <FormattedMessage id={label} />
-          ) : typeof label === 'function' ? (
-            <FormattedMessage id={label(config)} />
+  }) => {
+    const [custom, setCustom] = useState(false);
+    const [values, setValues] = useState({
```
heshan0131, line 90: instead of create a state here for value, value, value1, I would defer to have the actual input component to be stateful, 
for example, you can create a lazy input component, that will only call `onChange` when key is pressed and onBlue
```js
  const LazyInput = ({value, onChange, name}) => {
    const [stateValue, setValue] = useState(value);
    const inputRef = useRef(null);
    useEffect(() => {
      setValue(value);
    }, [value]);

    const onKeyDown = useCallback(
      e => {
        switch (e.keyCode) {
          case KeyEvent.DOM_VK_ENTER:
          case KeyEvent.DOM_VK_RETURN:
            onChange(stateValue);
            if (inputRef !== null) {
              // @ts-ignore
              inputRef?.current.blur();
            }
            break;
          default:
            break;
        }
      },
      [onChange, stateValue]
    );

    const _onChange = useCallback(e => setValue(e.target.value), [setValue]);
    const onBlur = useCallback(() => onChange(name, stateValue), [onChange, name, stateValue]);

    return (
      <RangeInput
        type="number"
        ref={inputRef}
        value={stateValue}
        onChange={_onChange}
        onBlur={onBlur}
        onKeyDown={onKeyDown}
        id={name}
      />
    );
  };
```

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
+    });
+
+    const updateField = e => {
+      setValues({
+        ...values,
+        [e.target.name]: e.target.value
+      });
+    };
+
+    const onKeyPress = e => {
```
heshan0131, line 103: remove this function, it is moved into the `LazyInput` component

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
+            value0={isRanged ? config.visConfig[property][0] : range[0]}
+            value1={isRanged ? config.visConfig[property][1] : config.visConfig[property]}
+            step={step}
+            isRanged={Boolean(isRanged)}
+            onChange={value => onChange({[property]: isRanged ? value : value[1]})}
+            inputTheme={inputTheme}
+            showInput
+          />
+        ) : (
+          renderCustomInput()
```
heshan0131, line 200: now we can just render our `CustomInput` component passing in the value from prop
```
        <CustomInput
            isRanged={isRanged}
            value={config.visConfig[property]}
            onChangeCustomInput={v => onChange({[property]: v})}
          />
```

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -50,31 +85,123 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
-            <FormattedMessage id={label(config)} />
+  }) => {
+    const [custom, setCustom] = useState(false);
+    const [values, setValues] = useState({
+      value: config.visConfig[property],
+      value0: config.visConfig.radiusRange[0],
+      value1: config.visConfig.radiusRange[1]
+    });
+
+    const updateField = e => {
```
heshan0131, line 96: remove this function, it is moved into the `CustomInput` component

heshan0131:
One more scenario to address here. When user set the slide from `custom = true` to `custom = false` we need to re adjust the value so it fits the slider. otherwise you will see slider rendered as below, because the custom input value is larger then the max value of the slider.

Screen Shot 2021-09-27 at 11 52 15 PM

The way to do that is to add additional handling in `setCustom`

```js
import {clamp} from 'utils/data-utils';
const value = config.visConfig[property];

// inside VisConfigSlider
const onChangeCheckbox = useCallback(() => {
  if (custom) {
    // we are swithcing from custom to not custom
    // adjust value to range
    const adjustedValue = isRanged
     ? [clamp(range, value[0]), clamp(range, value[1])]
     : clamp(range, value);
    onChange({[property]: adjustedValue});
  }

   setCustom(!custom);
  }, [onChange, property, isRanged, value, range, custom, setCustom]);

// pass onChangeCheckbox to checkbox
<Checkbox
  id={`property.${property}`}
  checked={custom}
  onChange={onChangeCheckbox}
/>
```

morphoix:
> One more scenario to address here. When user set the slide from `custom = true` to `custom = false` we need to re adjust the value so it fits the slider. otherwise you will see slider rendered as below, because the custom input value is larger then the max value of the slider.
> 
> Screen Shot 2021-09-27 at 11 52 15 PM
> 
> The way to do that is to add additional handling in `setCustom`
> 
> ```js
> import {clamp} from 'utils/data-utils';
> 
> // inside VisConfigSlider
> const value = config.visConfig[property];
> const onChangeCheckbox = useCallback(() => {
>   if (custom) {
>     // we are swithcing from custom to not custom
>     // adjust value to range
>     const adjustedValue = isRanged
>      ? [clamp(range, value[0]), clamp(range, value[1])]
>      : clamp(range, value);
>     onChange({[property]: adjustedValue});
>   }
> 
>    setCustom(!custom);
>   }, [onChange, property, isRanged, value, range, custom, setCustom]);
> 
> // pass onChangeCheckbox to checkbox
> <Checkbox
>   id={`property.${property}`}
>   checked={custom}
>   onChange={onChangeCheckbox}
> />
> ```

ok, got it, fixed in last commit :)

heshan0131:
As a followup, because we don't want to allow custom input on all layer visconfig sliders. Some of the properties doesn't make sense to allow user to input value out of range. For example, we don't want to allow user to set 1.5 on opacity, which should always to between 0 to 1.

To achieve this, let's  add a new property `allowCustomValue: true` to `LAYER_VIS_CONFIGS` in `layer-factory.js` 
```js
export const LAYER_VIS_CONFIGS = {
  thickness: {
    type: 'number',
    defaultValue: 2,
    label: 'layerVisConfigs.strokeWidth',
    isRanged: false,
    range: [0, 100],
    step: 0.1,
    group: PROPERTY_GROUPS.stroke,
    property: 'thickness',
    allowCustomValue: true
  },...
}
```

add let's add `allowCustomValue: true` to
```js
thickness
strokeWidthRange
trailLength
radius
radiusRange
clusterRadius
clusterRadiusRange
sizeScale
worldUnitSize
elevationScale
elevationRange
heightRange
weight
heatmapRadius
```

and `allowCustomValue: false` to all the other properties where ` type: 'number'`.

Finally in `vis-config-slider` let's only render the custom value toggle when `allowCustomValue` is true

morphoix:
> As a followup, because we don't want to allow custom input on all layer visconfig sliders. Some of the properties doesn't make sense to allow user to input value out of range. For example, we don't want to allow user to set 1.5 on opacity, which should always to between 0 to 1.
> 
> To achieve this, let's add a new property `allowCustomValue: true` to `LAYER_VIS_CONFIGS` in `layer-factory.js`
> 
> ```js
> export const LAYER_VIS_CONFIGS = {
>   thickness: {
>     type: 'number',
>     defaultValue: 2,
>     label: 'layerVisConfigs.strokeWidth',
>     isRanged: false,
>     range: [0, 100],
>     step: 0.1,
>     group: PROPERTY_GROUPS.stroke,
>     property: 'thickness',
>     allowCustomValue: true
>   },...
> }
> ```
> 
> add let's add `allowCustomValue: true` to
> 
> ```js
> thickness
> strokeWidthRange
> trailLength
> radius
> radiusRange
> clusterRadius
> clusterRadiusRange
> sizeScale
> worldUnitSize
> elevationScale
> elevationRange
> heightRange
> weight
> heatmapRadius
> ```
> 
> and `allowCustomValue: false` to all the other properties where ` type: 'number'`.
> 
> Finally in `vis-config-slider` let's only render the custom value toggle when `allowCustomValue` is true

ok, I also thought about it :)

src/layers/layer-factory.js
```diff
@@ -86,7 +86,8 @@ export const LAYER_VIS_CONFIGS = {
     range: [0, 100],
     step: 0.1,
     group: PROPERTY_GROUPS.stroke,
-    property: 'thickness'
+    property: 'thickness',
+    allowCustomValue: true
```
heshan0131, line 90: can you add `allowCustomValue: false` to properties that is `type: 'number'`? for example: `opacity`, `coverage`..

morphoix: > can you add `allowCustomValue: false` to properties that is `type: 'number'`? for example: `opacity`, `coverage`..

ok, I also added it to percentage, is it correct?

heshan0131: Yes

src/components/side-panel/layer-panel/vis-config-slider.js
```diff
@@ -47,34 +156,68 @@ export default function VisConfigSliderFactory(RangeSlider) {
...
+              <FormattedMessage id={label} />
+            ) : typeof label === 'function' ? (
+              <FormattedMessage id={label(config)} />
+            ) : (
+              <FormattedMessage id={`property.${property}`} />
+            )}
+          </PanelLabel>
+        ) : null}
+
+        {allowCustomValue ? (
```
heshan0131, line 193: It is also possible to have a saved maps with custom input value, so we need to add additional condition here to show CustomInputLabel if `props.value` is outside `props.range`. You can use `isInRange` from `filter-utils` to do the check

morphoix: > It is also possible to have a saved maps with custom input value, so we need to add additional condition here to show CustomInputLabel if `props.value` is outside `props.range`. You can use `isInRange` from `filter-utils` to do the check

{allowCustomValue && !isInRange(value, range) ? (
 did I understand correct?

Issue #1659, opened 2021-11-30
[Bug] Geocoder data is visible in the Layer and Filters tab


mletic:
**Describe the bug**
If you add a Geocoder marker, the Geocoder dataset is visible in the Layers and Filters section.

Also the tooltip section in the interactions tab has fields from the Geocoder but no tooltip is shown. This might be reported in another issue if we actually want to show that tooltip. If not, it should be removed from there too.

**To Reproduce**
1. Go to `Interactions` tab
2. Enable Geocoder
3. Enter any location to show the Geocoder tooltip
4. `geocoder_dataset` appears in the `Layers` and `Filters` tab section

**Expected behavior**
Geocoder dataset shouldn't be visible in the `Layers` and `Filters` tab section.

**Screenshots**
<img width="1792" alt="Screen Shot 2021-11-24 at 1 46 08 PM" src="https://user-images.githubusercontent.com/7241522/144052601-50ef2384-e46c-4e15-8ca3-be4277d71a6b.png">
<img width="1792" alt="Screen Shot 2021-11-24 at 1 46 16 PM" src="https://user-images.githubusercontent.com/7241522/144052614-ab1c5608-49ee-429c-b507-efa7c2f7067d.png">
<img width="1792" alt="Screen Shot 2021-11-24 at 1 46 04 PM" src="https://user-images.githubusercontent.com/7241522/144052621-cdab832e-9898-439c-be9c-019a4c7f4ba0.png">
<img width="1792" alt="Screen Shot 2021-11-24 at 1 50 35 PM" src="https://user-images.githubusercontent.com/7241522/144052625-53f37cf0-3a10-4b63-b782-3f1afa4d00f7.png">

PR #1661, opened 2021-12-04
[Bug] do not display geocoder dataset in side panel

hodoje:
This PR fixes the issue described here: https://github.com/keplergl/kepler.gl/issues/1659
Signed-off-by: Nikola Karaklic <nikolakaraklic@gmail.com>

### DIFF
--- a/src/components/geocoder-panel.js
+++ b/src/components/geocoder-panel.js
@@ -97,6 +97,16 @@ function isValid(key) {
   return /pk\..*\..*/.test(key);
 }
 
+export function getUpdateVisDataPayload(lat, lon, text) {
+  return [
+    [generateGeocoderDataset(lat, lon, text)],
+    {
+      keepExistingConfig: true
+    },
+    PARSED_CONFIG
+  ];
+}
+
 export default function GeocoderPanelFactory() {
   class GeocoderPanel extends Component {
     static propTypes = {
@@ -122,13 +132,8 @@ export default function GeocoderPanelFactory() {
         bbox
       } = geoItem;
       this.removeGeocoderDataset();
-      this.props.updateVisData(
-        [generateGeocoderDataset(lat, lon, text)],
-        {
-          keepExistingConfig: true
-        },
-        PARSED_CONFIG
-      );
+      const updateVisDataPayload = getUpdateVisDataPayload(lat, lon, text);
+      this.props.updateVisData(...updateVisDataPayload);
       const bounds = bbox || [
         lon - GEOCODER_GEO_OFFSET,
         lat - GEOCODER_GEO_OFFSET,
--- a/src/components/geocoder-panel.js
+++ b/src/components/geocoder-panel.js
@@ -132,8 +132,7 @@ export default function GeocoderPanelFactory() {
         bbox
       } = geoItem;
       this.removeGeocoderDataset();
-      const updateVisDataPayload = getUpdateVisDataPayload(lat, lon, text);
-      this.props.updateVisData(...updateVisDataPayload);
+      this.props.updateVisData(...getUpdateVisDataPayload(lat, lon, text));
       const bounds = bbox || [
         lon - GEOCODER_GEO_OFFSET,
         lat - GEOCODER_GEO_OFFSET,
--- a/src/components/kepler-gl.js
+++ b/src/components/kepler-gl.js
@@ -27,6 +27,7 @@ import {connect as keplerGlConnect} from 'connect/keplergl-connect';
 import {IntlProvider} from 'react-intl';
 import {messages} from '../localization';
 import {RootContext} from 'components/context';
+import cloneDeep from 'lodash.clonedeep';
 
 import * as VisStateActions from 'actions/vis-state-actions';
 import * as MapStateActions from 'actions/map-state-actions';
@@ -39,7 +40,8 @@ import {
   KEPLER_GL_NAME,
   KEPLER_GL_VERSION,
   THEME,
-  DEFAULT_MAPBOX_API_URL
+  DEFAULT_MAPBOX_API_URL,
+  GEOCODER_DATASET_NAME
 } from 'constants/default-settings';
 import {MISSING_MAPBOX_TOKEN} from 'constants/user-feedbacks';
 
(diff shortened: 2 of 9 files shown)


src/components/kepler-gl.js
```diff
@@ -138,30 +140,47 @@ export const mapFieldsSelector = props => ({
...
+export const sidePanelSelector = (props, availableProviders) => {
+  // visibleDatasets
+  const filteredDatasets = getVisibleDatasets(props.visState.datasets);
 
-  width: props.sidePanelWidth,
-  availableProviders,
-  mapSaved: props.providerState.mapSaved
-});
+  // And we don't want Geocoder dataset to be present in the tooltip field picker
+  const interactionConfigClone = cloneDeep(props.visState.interactionConfig);
```
heshan0131, line 153: don't use `cloneDeep` in compoennt.

Do this filter in the interaction config panel, instead of here

heshan0131:
```js
        {Object.keys(config.fieldsToShow).map(dataId => 
           dataId === GEOCODER_DATASET_NAME ? null : 
          (<DatasetTooltipConfig
            key={dataId}
            config={config}
            onChange={onChange}
            dataset={datasets[dataId]}
          />)
        ))}
```

hodoje: Thanks, will do!

hodoje: cloneDeep shouldn't be used inside a component because it is slow or is there some other reason?

macrigiuseppe: it is really slow since he copied every single property recursively

hodoje: Yeah, that's what I thought. New commit in a minute.

src/components/geocoder-panel.js
```diff
@@ -122,13 +132,8 @@ export default function GeocoderPanelFactory() {
...
       } = geoItem;
       this.removeGeocoderDataset();
-      this.props.updateVisData(
-        [generateGeocoderDataset(lat, lon, text)],
-        {
-          keepExistingConfig: true
-        },
-        PARSED_CONFIG
-      );
+      const updateVisDataPayload = getUpdateVisDataPayload(lat, lon, text);
```
macrigiuseppe, line 135: you don't need to create a redundant variable. You can do this
```
this.props.updateVisData(...getUpdateVisDataPayload(lat, lon, text));
```

src/components/kepler-gl.js
heshan0131: Also, react does shallow compare on props to determine whether a component should be rerendered. deep clone will cause all the components w/ `interactionConfig` as props to rerender all the time

src/components/geocoder-panel.js
```diff
@@ -97,6 +97,16 @@ function isValid(key) {
   return /pk\..*\..*/.test(key);
 }
 
+export function getUpdateVisDataPayload(lat, lon, text) {
+  return [
+    [generateGeocoderDataset(lat, lon, text)],
+    {
+      keepExistingConfig: true
+    },
+    PARSED_CONFIG
```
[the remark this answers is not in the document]
hodoje: This was a copy-paste from an already existing code.

src/components/geocoder-panel.js
hodoje: At first, I did that but then I wanted to be more readable from my perspective so I added it to a variable. Will fix that.

src/components/kepler-gl.js
hodoje: Fixed.

macrigiuseppe:
The geocoder dataset tooltip customization panel is no longer available. Checkout the picture below, on the left side your branch, on the right side master branch

Screen Shot 2021-12-06 at 11 07 11 AM

On the right picture you ca see the geocoder tooltip panel

heshan0131:
> The geocoder dataset tooltip customization panel is no longer available. Checkout the picture below, on the left side your branch, on the right side master branch
> 
> Screen Shot 2021-12-06 at 11 07 11 AM
> 
> On the right picture you ca see the geocoder tooltip panel

Correct. And that's what we want in the PR? We don't want people to customize the tooltip. It should be displayed by default

hodoje:
I don't understand. I was told that we don't want to have the geocoder dataset tooltip customization panel. Also there is a problem of not even displaying a tooltip when you hover over the marker.

macrigiuseppe:
> I don't understand. I was told that we don't want to have the geocoder dataset tooltip customization panel since we don't want to have the geocoder dataset displayed anyway. Also, there is a problem of not even displaying a tooltip when you hover over the marker (this problem was present before any of my changes).

just following up
- geocoder dataset tooltip panel should be hidden (no work required here)
- geocoder layer should display the tooltip with coordinate information

hodoje:
> > I don't understand. I was told that we don't want to have the geocoder dataset tooltip customization panel since we don't want to have the geocoder dataset displayed anyway. Also, there is a problem of not even displaying a tooltip when you hover over the marker (this problem was present before any of my changes).
> 
> just following up
> 
> * geocoder dataset tooltip panel should be hidden (no work required here)
> * geocoder layer should display the tooltip with coordinate information

Yeah, in the original Issue the focus was on the Layers and Filters and the Tooltip was mentioned in terms of maybe making a separate issue for that, but yeah, will fix the issue of not displaying a tooltip for the geocoder marker.

macrigiuseppe:
after some considerations we realized adding tooltip to geocoder will take too long

macrigiuseppe (approved):

src/components/kepler-gl.js
```diff
@@ -138,30 +139,40 @@ export const mapFieldsSelector = props => ({
...
-  filters: props.visState.filters,
-  layers: props.visState.layers,
-  layerOrder: props.visState.layerOrder,
-  layerClasses: props.visState.layerClasses,
-  interactionConfig: props.visState.interactionConfig,
-  mapInfo: props.visState.mapInfo,
-  layerBlending: props.visState.layerBlending,
+export const sidePanelSelector = (props, availableProviders) => {
+  // visibleDatasets
+  const filteredDatasets = getVisibleDatasets(props.visState.datasets);
```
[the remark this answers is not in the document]
hodoje: The problem here is that the thing that should be memoized is the reference to props.visState.datasets, not its content.
KeplerGL's state and visState will change onHover, this means the render function will be called invoking the sidePanelSelector function.
This function will use props.visState.datasets that are passed to SidePanels.
The objective is to filter out the GEOCODER_DATASET and we can do that either in a mutable or immutable way.
If we do it in an immutable way then we lose the reference since we are creating a new object (current solution).
The only solution is to do it in a mutable way where we change the props.visState.datasets object but keep the reference.
The dirty way is to just do something like: delete props.visState.datasets[GEOCODER_DATASET_NAME] (which works) but something better should be implemented if possible (will try).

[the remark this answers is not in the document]
heshan0131: you can create selector for `filteredDatasets`

```js
  // component method:
  filteredDatasetsSelector = createSelector(props => props.visState.datasets, getVisibleDatasets);

  // in render function
  const filteredDatasets = this.filteredDatasetsSelector(this.props);
  // add filteredDatasets as additional argument to sidePanelSelector
  const sideFields = sidePanelSelector(this.props, availableProviders, filteredDatasets);
```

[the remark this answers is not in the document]
hodoje: This doesn't solve the issue of calling render for LayerManager (or its children). Memoization needs to be done in children too.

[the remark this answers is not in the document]
hodoje: Shan and I agreed that we add the filteredDatasetsSelector in order to not change the datasets reference, but the other issues will be handled some other time.

PR #1665, opened 2021-12-10
[Enhancement] extract layers list to a separate component

hodoje:
Extracts the layers list into a separate component which will help the implementation of https://github.com/keplergl/kepler.gl/issues/1658

Signed-off-by: hodoje <nikolakaraklic@gmail.com>

### DIFF
--- a/src/components/side-panel/layer-panel/layer-list.d.ts
+++ b/src/components/side-panel/layer-panel/layer-list.d.ts
@@ -1,0 +1,19 @@
+import React from 'react';
+import {Layer, LayerClassesType} from 'layers';
+import {Datasets} from 'reducers/vis-state-updaters';
+import * as VisStateActions from 'actions/vis-state-actions';
+import * as UIStateActions from 'actions/ui-state-actions';
+
+export type LayerListProps = {
+  datasets: Datasets;
+  layerClasses: LayerClassesType;
+  layers: Layer[];
+  layerOrder: number[];
+  uiStateActions: typeof UIStateActions;
+  visStateActions: typeof VisStateActions;
+  isSortable: Boolean;
+};
+
+export default function LayerListFactory(
+  LayerPanel: React.Component
+): React.Component<LayerListProps>;
--- a/src/components/side-panel/layer-panel/layer-list.d.ts
+++ b/src/components/side-panel/layer-panel/layer-list.d.ts
@@ -11,7 +11,7 @@ export type LayerListProps = {
   layerOrder: number[];
   uiStateActions: typeof UIStateActions;
   visStateActions: typeof VisStateActions;
-  isSortable: Boolean;
+  isSortable?: Boolean;
 };
 
 export default function LayerListFactory(
--- a/src/components/side-panel/layer-panel/layer-list.js
+++ b/src/components/side-panel/layer-panel/layer-list.js
@@ -1,21 +1,35 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
 import React, {useCallback, useMemo, useState} from 'react';
-
 import {arrayMove} from 'utils/data-utils';
 import styled from 'styled-components';
 import classnames from 'classnames';
 import {SortableContainer, SortableElement} from 'react-sortable-hoc';
-
 import LayerPanelFactory from './layer-panel';
-
 // make sure the element is always visible while is being dragged
 // item being dragged is appended in body, here to reset its global style
 const SortableStyledItem = styled.div`
   z-index: ${props => props.theme.dropdownWrapperZ + 1};
-
   &.sorting {
     pointer-events: none;
   }
-
   &.sorting-layers .layer-panel__header {
     background-color: ${props => props.theme.panelBackgroundHover};
     font-family: ${props => props.theme.fontFamily};
@@ -33,9 +47,7 @@ const SortableStyledItem = styled.div`
     }
   }
 `;
-
 LayerListFactory.deps = [LayerPanelFactory];
-
 function LayerListFactory(LayerPanel) {
   // By wrapping layer panel using a sortable element we don't have to implement the drag and drop logic into the panel itself;
   // Developers can provide any layer panel implementation and it will still be sortable
@@ -46,11 +58,9 @@ function LayerListFactory(LayerPanel) {
       </SortableStyledItem>
     );
   });
-
   const WrappedSortableContainer = SortableContainer(({children}) => {
     return <div>{children}</div>;
   });
-
   const LayerList = props => {
     const {
       layers,
@@ -63,7 +73,6 @@ function LayerListFactory(LayerPanel) {
     } = props;
     const {toggleModal: openModal} = uiStateActions;
     const [isSorting, setIsSorting] = useState(false);
-
     const layerTypeOptions = useMemo(
       () =>
         Object.keys(layerClasses).map(key => {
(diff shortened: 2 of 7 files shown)


src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
...
+      uiStateActions,
+      visStateActions,
+      layerClasses,
+      isSortable
+    } = props;
+    const {toggleModal: openModal} = uiStateActions;
+    const [isSorting, setIsSorting] = useState(false);
+
+    const layerClassSelector = () => layerClasses;
+    const layerTypeOptionsSelector = createSelector(layerClassSelector, layerCls =>
```
heshan0131, line 70: We are moving from reselect to react hooks. Here we should use `useMemo` instead of `createSelector`

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
...
+        return {
+          id: key,
+          label: layer.name,
+          icon: layer.layerIcon,
+          requireData: layer.requireData
+        };
+      })
+    );
+
+    const layerTypeOptions = layerTypeOptionsSelector(props);
```
heshan0131, line 82: same here

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
...
+      duplicateLayer: visStateActions.duplicateLayer
+    };
+
+    const panelProps = {
+      datasets,
+      openModal,
+      layerTypeOptions
+    };
+
+    const _handleSort = ({oldIndex, newIndex}) => {
```
heshan0131, line 101: use `useCallback`

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
...
+      visStateActions.reorderLayer(arrayMove(props.layerOrder, oldIndex, newIndex));
+      setIsSorting(false);
+    };
+
+    const _onSortStart = () => {
```
heshan0131, line 106: use `useCallback`

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
...
+      setIsSorting(true);
+    };
+
+    const _updateBeforeSortStart = ({index}) => {
```
heshan0131, line 110: use `useCallback`

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,174 @@
+import React, {useState} from 'react';
```
heshan0131, line 1: add a `layer-list.d.ts` for prop types

src/components/side-panel/layer-panel/layer-list.js
hodoje: This can't be added to useCallback since it doesn't have sideeffects. At least that's what VSC is telling me.

src/components/side-panel/layer-panel/layer-list.js
hodoje: Since this is callable we need to use useCallback, not useMemo.

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,168 @@
...
+            label: layer.name,
+            icon: layer.layerIcon,
+            requireData: layer.requireData
+          };
+        }),
+      [layerClasses]
+    );
+
+    const layerTypeOptions = useMemo(() => layerTypeOptionsSelector(), [layerTypeOptionsSelector]);
+
```
macrigiuseppe, line 82: why creating `layerTypeOptionsSelector` when you only use it once?
you can simply do:
```
const layerTypeOptions = useMemo(() => Object.keys(layerClasses).map(key => {
          const layer = new layerClasses[key]();
          return {
            id: key,
            label: layer.name,
            icon: layer.layerIcon,
            requireData: layer.requireData
          };
        }), [layerClasses]);
```

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,168 @@
...
+    const layerActions = {
+      layerColorUIChange: visStateActions.layerColorUIChange,
+      layerConfigChange: visStateActions.layerConfigChange,
+      layerVisualChannelConfigChange: visStateActions.layerVisualChannelConfigChange,
+      layerTypeChange: visStateActions.layerTypeChange,
+      layerVisConfigChange: visStateActions.layerVisConfigChange,
+      layerTextLabelChange: visStateActions.layerTextLabelChange,
+      removeLayer: visStateActions.removeLayer,
+      duplicateLayer: visStateActions.duplicateLayer
+    };
```
macrigiuseppe, line 92: take advantage of useMemo here otherwise you will trigger too many re-renderes

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,168 @@
...
+
+    const panelProps = {
```
macrigiuseppe, line 94: same here

test/browser/components/side-panel/layer-list-test.js
```diff
@@ -0,0 +1,78 @@
...
+          datasets={mock.visState.datasets}
+          layers={mock.visState.layers}
+          layerOrder={mock.visState.layerOrder}
+          layerClasses={mock.visState.layerClasses}
+        />
+      </IntlWrapper>
+    );
+  }, 'LayerList should render');
+
+  t.notEqual(wrapper.find('.sortable-layer-items').length, 0, 'should render sortable items');
```
macrigiuseppe, line 45: don't check if the number of sortable items is `0` but check the exact number you expect
```
t.equal(wrapper.find('.sortable-layer-items').length, N, 'should render N sortable items');
```

src/components/side-panel/layer-panel/layer-list.d.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as UIStateActions from 'actions/ui-state-actions';
+
+export type LayerListProps = {
+  datasets: Datasets;
+  layerClasses: LayerClassesType;
+  layers: Layer[];
+  layerOrder: number[];
+  uiStateActions: typeof UIStateActions;
+  visStateActions: typeof VisStateActions;
+  isSortable: Boolean;
```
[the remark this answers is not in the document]
hodoje: What should be the default value then?
I've put it to required since we will have only two displays (list or by-dataset) and in first we want it to be sortable and in other not to be sortable.

src/components/side-panel/layer-panel/layer-list.js
hodoje: Idk, the whole code was just a copy-paste from layer-manager so I didn't think about it much, honestly.

test/browser/components/side-panel/layer-list-test.js
hodoje: I can do that but the idea was that 'sortable-layer-items' selector will be only present in case it is rendering the sortable list, so any number greater than zero (i was looking for a check for greaterThan, but it doesn't exist in 'tape') will mean it is a sortable list.

src/components/side-panel/layer-panel/layer-list.d.ts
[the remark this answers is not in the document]
hodoje: I was thinking about if we should be explicit about what mode it should be in?

test/browser/components/side-panel/layer-list-test.js
macrigiuseppe: yeah but we also want to make sure that the list will display exactly the items you expect, e.g. if one the layers is hidden it should not be displayed

hodoje: Shouldn't that be in a separate test? I mean, it's not the same use-case. But yeah, I'll put the exact number.

test/browser/components/side-panel/layer-list-test.js
```diff
@@ -0,0 +1,78 @@
...
+
+// default props from initial state
+const defaultProps = {
+  datasets: InitialState.visState.datasets,
+  layerClasses: InitialState.visState.layerClasses,
+  layerOrder: InitialState.visState.layerOrder,
+  layers: InitialState.visState.layers,
+  uiStateActions: UIStateActions,
+  visStateActions: VisStateActions
+};
```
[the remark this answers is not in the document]
hodoje: I was looking at other tests and saw the pattern where the defaultProps with the initial state were created and then overridden with custom props in a specific test. I'll change to what you proposed if you think that is better.

macrigiuseppe (approved):

test/browser/components/side-panel/layer-list-test.js
```diff
@@ -0,0 +1,59 @@
...
+      <IntlWrapper>
+        <LayerList {...defaultProps} isSortable />
+      </IntlWrapper>
+    );
+  }, 'LayerList should render');
+
+  t.equal(wrapper.find('.sortable-layer-items').length, 3, 'should render 3 sortable items');
+
+  const layers = wrapper.find('.layer-panel');
+  t.equal(layers.length, 3, 'should render 3 layer panels');
```
heshan0131, line 38: please add test for action layer order, select layer panel and its displayed layer label to see if the order rendered correctly

test/browser/components/side-panel/layer-list-test.js
```diff
@@ -0,0 +1,59 @@
...
+      <IntlWrapper>
+        <LayerList {...defaultProps} isSortable={false} />
+      </IntlWrapper>
+    );
+  }, 'LayerList should render');
+
+  t.equal(wrapper.find('.sortable-layer-items').length, 0, 'should not render sortable items');
+
+  const layers = wrapper.find('.layer-panel');
+  t.equal(layers.length, 3, 'should render 3 layer panels');
```
heshan0131, line 56: same here

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,179 @@
...
+      visStateActions.layerConfigChange,
+      visStateActions.layerVisualChannelConfigChange,
+      visStateActions.layerTypeChange,
+      visStateActions.layerVisConfigChange,
+      visStateActions.layerTextLabelChange,
+      visStateActions.removeLayer,
+      visStateActions.duplicateLayer
+    ]);
+
+    const panelProps = useMemo(() => {
```
heshan0131, line 103: You don't need to instantiate `panelProps` with `useMemo` `panelProps` will be spread into `<LayerPanel>`,  useMemo will do anything to improve perf

hodoje: I listened to Giuseppe so I added it. It wasn't there at first and it didn't make sense to me but ok. I will remove it.

src/components/side-panel/layer-panel/layer-list.js
```diff
@@ -0,0 +1,179 @@
...
+            id: key,
+            label: layer.name,
+            icon: layer.layerIcon,
+            requireData: layer.requireData
+          };
+        }),
+      [layerClasses]
+    );
+
+    const layerActions = useMemo(() => {
```
heshan0131, line 81: same here, `layerActions` will be spread into component

hodoje:
This PR is ready for review.

heshan0131 (approved):

PR #1666, opened 2021-12-13
[Feat] allow for multiple value column type

crabatel:
* add array type matching
* add array parsing
* add array ordinal domain generation

### DIFF
--- a/src/constants/default-settings.js
+++ b/src/constants/default-settings.js
@@ -333,6 +333,7 @@ export const ALL_FIELD_TYPES = keyMirror({
   real: null,
   string: null,
   timestamp: null,
+  array: null,
   point: null
 });
 
@@ -388,6 +389,7 @@ const BLUE = '140, 210, 205';
 const BLUE2 = '106, 160, 206';
 const BLUE3 = '0, 172, 237';
 const GREEN = '106, 160, 56';
+const GREEN2 = '23, 207, 114';
 const RED = '237, 88, 106';
 
 export const FILED_TYPE_DISPLAY = {
@@ -418,6 +420,10 @@ export const FILED_TYPE_DISPLAY = {
   [ALL_FIELD_TYPES.timestamp]: {
     label: 'time',
     color: GREEN
+  },
+  [ALL_FIELD_TYPES.array]: {
+    label: 'array',
+    color: GREEN2
   },
   // field pairs
   [ALL_FIELD_TYPES.point]: {
--- a/src/utils/filter-utils.js
+++ b/src/utils/filter-utils.js
@@ -329,6 +329,7 @@ export function getFilterProps(field, fieldDomain) {
         gpu: false
       };
 
+    case ALL_FIELD_TYPES.array:
     case ALL_FIELD_TYPES.string:
     case ALL_FIELD_TYPES.date:
       return {
@@ -948,6 +949,7 @@ export function mergeFilterDomainStep(filter, filterProps) {
   };
 
   switch (filterProps.fieldType) {
+    case ALL_FIELD_TYPES.array:
     case ALL_FIELD_TYPES.string:
     case ALL_FIELD_TYPES.date:
       return {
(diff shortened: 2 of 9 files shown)


heshan0131:
Please `git rm dist` from your commit.

package.json
```diff
@@ -214,8 +214,6 @@
     "prettier": "1.19.1",
     "progress-bar-webpack-plugin": "^2.1.0",
     "puppeteer": "^2.1.1",
-    "react": "^16.8.4",
```
heshan0131: any reason why this is removed?

src/utils/filter-utils.js
```diff
@@ -413,7 +414,13 @@ export function getFilterFunction(field, dataId, filter, layers) {
     case FILTER_TYPES.range:
       return data => isInRange(valueAccessor(data), filter.value);
     case FILTER_TYPES.multiSelect:
-      return data => filter.value.includes(valueAccessor(data));
+      return data => {
+        const value = valueAccessor(data);
+        if (Array.isArray(value)) {
```
heshan0131: To make this function more efficient, we should do the checking out side the `data => boolean` function
```js
case FILTER_TYPES.multiSelect:
return feld.type === ALL_FILED_TYPES.array ? data => {
  const value = valueAccessor(data);
  return Array.isArray(value) && value.some(v => filter.value.includes(v));
} : data => filter.value.includes(valueAccessor(data));
```

package.json
crabatel: Hello,
Actually I removed this because I thought having react as a dependency there was the cause of an error when trying to `npm link` the Kepler package (@see https://stackoverflow.com/questions/66975613/error-invalid-hook-call-hooks-can-only-be-called-inside-of-the-body-of-a-funct), but it appeared it wasn't the reason as we could not get rid of this error.

As we could not `npm link` we resolved to pushing the `dist/` folder on our repo and updating our dependency to point to our repo, hence the presence of the `dist/` folder.

Concerning this PR, we were not sure it could be accepted as we made the assumption that an array typed column in the CSV would be in a JSON format. This is the format we decided to use for our project, but it may not be the case for everyone, and maybe it should be discussed.

If the changes can be accepted, I'll remove the `dist/` from this branch as well as the dependency change.

src/utils/filter-utils.js
crabatel: Indeed it's better, I'll update this one.

.github/workflows/npmpublish.yml
```diff
@@ -1,8 +1,10 @@
 name: Node.js Package
 
 on:
-  release:
-    types: [created]
+  push:
```
heshan0131, line 4: please remove this

package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "kepler.gl",
```
heshan0131, line 2: please remove changes to package.json

src/constants/default-settings.js
```diff
@@ -46,7 +46,7 @@ import {TOOLTIP_FORMAT_TYPES} from './tooltip';
 import {LAYER_TYPES} from 'layers/types';
 
 export const ACTION_PREFIX = '@@kepler.gl/';
-export const CLOUDFRONT = 'https://d1a3f4spazzrp4.cloudfront.net/kepler.gl';
+export const CLOUDFRONT = 'https://raw.githubusercontent.com/datatlas-erasme/front/master/src/src/static/';
```
heshan0131, line 49: remove this

src/layers/icon-layer/icon-layer.js
```diff
@@ -30,7 +30,7 @@ import {getTextOffsetByRadius, formatTextLabelData} from '../layer-text-label';
 
 const brushingExtension = new BrushingExtension();
 
-export const SVG_ICON_URL = `${CLOUDFRONT}/icons/svg-icons.json`;
+export const SVG_ICON_URL = `${CLOUDFRONT}/icons.json`;
```
heshan0131, line 33: remove here

.github/workflows/npmpublish.yml
Pipazoul: done, sorry for that

PR #1675, opened 2022-01-05
[Enhancement] order layers by dataset

hodoje:
This PR addresses the issue: https://github.com/keplergl/kepler.gl/issues/1658
And also it relies on https://github.com/keplergl/kepler.gl/pull/1665

Here are some component diagrams showing the current and future component layout
Screenshot 2022-01-05 at 21 21 30
Screenshot 2022-01-05 at 21 21 55
Screenshot 2022-01-05 at 21 23 33

How things look now:

List view with two datasets and two layers (one per dataset):
Screenshot 2022-01-05 at 22 52 51

Sort by dataset view for the same datasets and layers:
Screenshot 2022-01-05 at 22 53 08

Clicking on the AddLayer button will open a dropdown list that will list all the loaded datasets and clicking on one of the items will create an empty layer for the selected dataset:
Screenshot 2022-01-05 at 22 53 24

Screenshot 2022-01-05 at 22 53 47

Did the same thing for the second dataset:
Screenshot 2022-01-05 at 22 54 08

How the order looks when in list view:
Screenshot 2022-01-05 at 22 54 23

When in list view the layers are sortable so after some sorting, we now have this in list view:
Screenshot 2022-01-05 at 22 54 43

Which in sort by dataset view looks like this:
Screenshot 2022-01-05 at 22 54 55


src/actions/vis-state-actions.d.ts
```diff
@@ -157,9 +157,11 @@ export function addFilter(
 
 export type AddLayerUpdaterAction = {
   config: object;
+  datasetId?: string;
 };
 export function addLayer(
-  config: object
+  config: object,
+  datasetId: string,
```
heshan0131, line 164: `datasetId?`

src/components/common/icons/list.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import Base from './base';
+import PropTypes from 'prop-types';
+
+const List = props => (
```
heshan0131, line 5: `List` is a very general word. `ViewByList` is more specific

src/components/index.js
```diff
@@ -59,12 +59,21 @@ export {CollapseButtonFactory, default as SidebarFactory} from './side-panel/sid
...
-export {AddDataButtonFactory, default as LayerManagerFactory} from './side-panel/layer-manager';
+export {default as LayerManagerFactory} from './side-panel/layer-manager';
 export {default as LayerPanelFactory} from './side-panel/layer-panel/layer-panel';
 export {default as LayerPanelHeaderFactory} from './side-panel/layer-panel/layer-panel-header';
 export {default as LayerConfiguratorFactory} from './side-panel/layer-panel/layer-configurator';
 export {default as TextLabelPanelFactory} from './side-panel/layer-panel/text-label-panel';
 export {LayerConfigGroupLabelFactory} from './side-panel/layer-panel/layer-config-group';
+export {
+  AddDataButtonFactory,
+  default as DatasetSectionFactor
```
heshan0131, line 70: `DatasetSectionFactory`

src/components/index.js
```diff
@@ -59,12 +59,21 @@ export {CollapseButtonFactory, default as SidebarFactory} from './side-panel/sid
...
 
-export {AddDataButtonFactory, default as LayerManagerFactory} from './side-panel/layer-manager';
+export {default as LayerManagerFactory} from './side-panel/layer-manager';
 export {default as LayerPanelFactory} from './side-panel/layer-panel/layer-panel';
 export {default as LayerPanelHeaderFactory} from './side-panel/layer-panel/layer-panel-header';
 export {default as LayerConfiguratorFactory} from './side-panel/layer-panel/layer-configurator';
 export {default as TextLabelPanelFactory} from './side-panel/layer-panel/text-label-panel';
 export {LayerConfigGroupLabelFactory} from './side-panel/layer-panel/layer-config-group';
+export {
+  AddDataButtonFactory,
```
heshan0131, line 69: remove this export, it is exported below

src/components/side-panel/layer-panel/dataset-layer-section.js
```diff
@@ -0,0 +1,55 @@
...
+      updateTableColor,
+      showDeleteDataset,
+      removeDataset,
+      layerOrder,
+      layerClasses,
+      uiStateActions,
+      visStateActions
+    } = props;
+
+    const datasets = {[dataset.id]: dataset};
```
heshan0131, line 28: `useMemo`

heshan0131:
Please add UI screenshots for this PR, essentially showing what UI has changed, and how it looks like now

src/components/side-panel/layer-panel/dataset-layer-section.js
hodoje: Are there some extreme cases of loaded datasets?
I mean, won't the useMemo overhead just slow things down over this simple operation?
Just my guess, list of datasets won't be longer than let's say 20 (I'm not sure what is the extreme case).
Using useMemo would probably have to do a lot of checks on which we lose time on each render, won't we?

hodoje:
> Please add UI screenshots for this PR, essentially showing what UI has changed, and how it looks like now

I thought the screenshots in the issue itself were enough.
Will add more screenshots.

heshan0131:
> > Please add UI screenshots for this PR, essentially showing what UI has changed, and how it looks like now
> 
> I thought the screenshots in the issue itself were enough. Will add more screenshots.

UI mockup might look different than the actual implementation

hodoje:
> > > Please add UI screenshots for this PR, essentially showing what UI has changed, and how it looks like now
> > 
> > 
> > I thought the screenshots in the issue itself were enough. Will add more screenshots.
> 
> UI mockup might look different than the actual implementation

True.

src/components/side-panel/layer-panel/dataset-layer-section.js
hodoje: Also, this here is just a single dataset being used to construct a new object.

src/components/side-panel/layer-panel/panel-view-list-toggle.js
```diff
@@ -0,0 +1,110 @@
...
+    label: 'Sort by dataset'
+  }
+];
+
+PanelViewListToggleFactory.deps = [ToggleOptionFactory];
+
+function PanelViewListToggleFactory(ToggleOption) {
+  const PanelViewListToggle = () => {
+    const layerPanelListViewMode = useSelector(
+      state => state.demo.keplerGl.map.uiState.layerPanelListView
```
heshan0131, line 76: don't use `useSelector` inside kepler.gl, kepler reducer can be mounted at any app state path. the `state.demo.keplerGl` is the demo app path. and for other application kepler can be mounted at other state path. Here it's better to just pass `uiState.layerPanelListView` through KeplerGl -> SidePanel

src/components/side-panel/layer-panel/dataset-layer-section.js
heshan0131: The part that it constructs a new object is the problem here. because `datasets` is always a new object, `SourceDataCatalog` and `LayerList` will be re-rendered all the time. If the component is small, or a leaf component, it might not be too bad, But here. it's a quite high lever component. It renders all the layer panel via the `LayerList` component. which means, all the layer panel component will be rerendered. because of the construct of a new object inside render function

hodoje: Yeah, true, totally forgot about the children.

src/components/side-panel.js
```diff
@@ -212,29 +212,56 @@ export default function SidePanelFactory(
...
+                  showDatasetTable={onShowDatasetTable}
+                  updateTableColor={onUpdateTableColor}
+                  showAddDataModal={onShowAddDataModal}
+                  showAddMapStyleModal={onShowAddMapStyleModal}
+                  uiStateActions={uiStateActions}
+                  visStateActions={visStateActions}
+                  panelMetadata={currentPanel}
+                  layerPanelListView={uiState.layerPanelListView}
+                />
+              )
```
macrigiuseppe, line 264: all of this can be simplified to the following:
```
{currentPanel.id !== 'layer' ? (
    <PanelTitle className="side-panel__content__title">
      <FormattedMessage id={currentPanel.label} />
    </PanelTitle>
    ) : null
  }
  <PanelComponent
    datasets={datasets}
    filters={filters}
    layers={layers}
    layerClasses={layerClasses}
    layerOrder={layerOrder}
    layerBlending={layerBlending}
    mapStyle={mapStyle}
    mapStyleActions={mapStyleActions}
    mapStateActions={mapStateActions}
    interactionConfig={interactionConfig}
    removeDataset={onRemoveDataset}
    showDatasetTable={onShowDatasetTable}
    updateTableColor={onUpdateTableColor}
    showAddDataModal={onShowAddDataModal}
    showAddMapStyleModal={onShowAddMapStyleModal}
    uiStateActions={uiStateActions}
    visStateActions={visStateActions}
    panelMetadata={currentPanel}
    layerPanelListView={currentPanel.id === 'layer' && uiState.layerPanelListView}
  />
```

src/components/side-panel/layer-panel/add-layer-button.js
```diff
@@ -0,0 +1,133 @@
...
+
+function AddLayerButtonFactory() {
+  const AddLayerButton = props => {
+    const {datasets, onOptionSelected} = props;
+    const [showAddLayerDropdown, setShowAddLayerDropdown] = useState(false);
+
+    const toggleAddLayerDropdown = () => {
+      setShowAddLayerDropdown(!showAddLayerDropdown);
+    };
+
```
macrigiuseppe, line 90: import {useCallback} from 'react';

const toggleAddLayerDropdown =  useCallback(() => {
      setShowAddLayerDropdown(!showAddLayerDropdown);
}, [setShowAddLayerDropdown, showAddLayerDropdown]);

src/components/side-panel/layer-panel/add-layer-button.js
```diff
@@ -0,0 +1,133 @@
...
+          <div>
+            <AddLayerMenu>
+              <DropdownContainer>
+                <DropdownList
+                  displayOption={d => d.label}
+                  options={Object.values(datasets)}
+                  onOptionSelected={dataset => {
+                    onOptionSelected(dataset);
+                    setShowAddLayerDropdown(false);
+                  }}
```
macrigiuseppe, line 109: create a callback for this

src/components/side-panel/layer-panel/add-layer-button.js
```diff
@@ -0,0 +1,133 @@
...
+                />
+              </DropdownContainer>
+            </AddLayerMenu>
+          </div>
+        }
+      >
+        <Button
+          tabIndex={-1}
+          onBlur={() => setShowAddLayerDropdown(false)}
```
macrigiuseppe, line 118: create a callback for this as well

src/components/side-panel/layer-panel/panel-view-list-toggle.js
```diff
@@ -0,0 +1,105 @@
...
+    const {layerPanelListViewMode, toggleLayerPanelListView} = props;
+
+    const toggleListView = listView => toggleLayerPanelListView(listView);
+    const options = useMemo(
+      () =>
+        TOGGLE_OPTIONS.map(opt => {
+          return layerPanelListViewMode === opt.value
+            ? {...opt, isActive: true}
+            : {...opt, isActive: false};
+        }),
```
macrigiuseppe, line 82: TOGGLE_OPTIONS.map(opt => ({...opt, isActive: layerPanelListViewMode === opt.value}),

src/components/side-panel/layer-panel/add-layer-button.js
hodoje: This doesn't make sense since useCallback is used for things that are passed from parent and showAddLayerDropdown is part of the state of AddLayerButton.

src/components/side-panel/layer-panel/add-layer-button.js
hodoje: Same as above though I could refactor this into a separate function.

src/components/side-panel/layer-panel/add-layer-button.js
hodoje: Similar to above but maybe this is just too simple for a separate function.

src/components/side-panel/layer-panel/add-layer-button.js
macrigiuseppe: In general is always a good practice to use memoization (useCallback, useMemo).

We should try to avoid creating inline objects (style, ...etc) and functions (onClick, ...etc). To improve performance we use  `useMemo` and `useCallback` which can take any dependencies (not just props)

A good read is this: https://itnext.io/6-tips-for-better-react-performance-4329d12c126b

I created an example to show you how to take advantage of `useCallback` appropriately.
Example: https://codesandbox.io/s/interesting-cherry-szu70?file=/src/App.js , if you toggle line 12 with 13 will notice the difference. 
When the `onToggle` is created with `useCallback` the `Button` component will not have unnecessary re-renders. I also use `Button = React.memo(...)` to simulate your specific use case with styled-components

src/components/side-panel/layer-panel/add-layer-button.js
macrigiuseppe: this is not for readability, the recommendation is based on performance

src/components/side-panel/layer-panel/add-layer-button.js
macrigiuseppe: same

src/components/side-panel/layer-panel/add-layer-button.js
hodoje: Wow, haven't looked at it from this perspective, so when a child just calls the function passed from the parent (and the parent function can change depending on props/state) we should use useCallback.
I'm used to using useCallback inside a child where the function depends on the props.
Thanks for this, and also I think this is also a good read: https://blog.logrocket.com/rethinking-hooks-memoization/ where sprinkling useMemo everywhere might affect performance since useMemo needs to do some checks always and useMemo is run on each render to do the checks.

src/components/side-panel/layer-panel/add-layer-button.js
hodoje: Yeah, read the comment above. Thanks!

macrigiuseppe (approved):

src/components/side-panel/layer-panel/add-layer-button.js
```diff
@@ -0,0 +1,142 @@
...
+function AddLayerButtonFactory() {
+  const AddLayerButton = props => {
+    const {datasets, onOptionSelected} = props;
+    const [showAddLayerDropdown, setShowAddLayerDropdown] = useState(false);
+
+    const toggleAddLayerDropdown = useCallback(() => {
+      setShowAddLayerDropdown(!showAddLayerDropdown);
+    }, [showAddLayerDropdown, setShowAddLayerDropdown]);
+
+    const toggleSelectedOption = useCallback(
```
hodoje, line 91: @macrigiuseppe I've noticed right now that after adding useCallback to toggleSelectedOption and onBlur I lose the feature of closing the dropdown after clicking on the option or outside of the dropdown. Will look into it in order to try and keep both useCallback and the functionality itself working. If not able I will ditch the useCallback in that case.

src/components/side-panel/layer-panel/add-layer-button.js
```diff
@@ -0,0 +1,142 @@
...
+      dataset => {
+        onOptionSelected(dataset);
+        setShowAddLayerDropdown(false);
+      },
+      [onOptionSelected]
+    );
+
+    const onButtonBlur = useCallback(() => {
```
hodoje, line 99: @macrigiuseppe I've noticed right now that after adding useCallback to onButtonBlur I lose the feature of closing the dropdown after clicking outside of the dropdown. Will look into it in order to try and keep both useCallback and the functionality itself working. If not able I will ditch the useCallback in that case.

heshan0131:
The dataset selector dropdown should look the same as the datasource selector shown below

Screen Shot 2022-01-11 at 10 42 37 AM

heshan0131:
disable sortable handle in the order by dataset view
Screen Shot 2022-01-12 at 9 54 46 AM

heshan0131 (approved):

PR #1711, opened 2022-02-15
[Feat]: Technical: Translate redusers (composers and combined-updaters) to typescript

HeimEndyd:


.eslintrc
```diff
@@ -19,12 +19,8 @@
     "context": true,
     "jestPuppeteer": true
   },
-  "parser": "babel-eslint",
-  "plugins": [
-    "babel",
-    "prettier",
-    "react"
-  ],
+  "parser": "@babel/eslint-parser",
```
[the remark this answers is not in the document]
HeimEndyd: Should I make a separate PR for this change. Or is it a recommendation for future PR?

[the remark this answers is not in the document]
ibgreen: No, it is fine for this PR. 

But for future, always try to make clearly focused PRs that make one change or one type of change. It is not always possible but just before adding reviewers to a PR think about how to make it easy for reviewers.

src/reducers/composer-helpers.ts
```diff
@@ -22,49 +22,54 @@ import Console from 'global/console';
...
 }
-
-export function merge_(obj) {
+/** Returns a reducer function that merges props with state */
+export function merge_<State, Props>(obj: Props): (state: State) => State & Props {
   return state => ({...state, ...obj});
 }
 
-export function pick_(prop) {
+export function pick_<State, Prop extends keyof State>(
```
heshan0131, line 59: You should instantiate the 2nd generic type in from of the function it is first used. so pick is better defined this way

```js
export function pick_<Prop extends string>(
  prop: Prop
): <Value>(fn: (p: Value) => Value) => <State extends Record<Prop, Value>>(state: State) => State {
  return fn => state => ({...state, [prop]: fn(state[prop])});
}
```

src/reducers/composer-helpers.ts
```diff
@@ -22,49 +22,54 @@ import Console from 'global/console';
...
 }
 
-export function compose_(fns) {
+export function compose_<State>(fns: Array<(s: State) => State>): (s: State) => State {
   return state => fns.reduce((state2, fn) => fn(state2), state);
 }
-
-export function merge_(obj) {
+/** Returns a reducer function that merges props with state */
+export function merge_<State, Props>(obj: Props): (state: State) => State & Props {
```
heshan0131, line 55: You should instantiate the 2nd generic type in from of the function it is first used. so `merge_` is better defined this way
```js
export function merge_<Props>(obj: Props): <State>(state: State) => State {
  return state => ({...state, ...obj});
}
```

src/reducers/composer-helpers.ts
```diff
@@ -22,49 +22,54 @@ import Console from 'global/console';
...
   return arr => arr.map(a => (a.id === item.id ? item : a));
 }
 
-export function findById(id) {
+export function findById<X extends {id: string}>(id: string): (arr: X[]) => X | undefined {
   return arr => arr.find(a => a.id === id);
 }
 
-export function map_(fn) {
+export function map_<X, Y>(fn: (e: X) => Y): (arr: X[]) => Y[] {
```
heshan0131, line 73: `map_` is meant to be ran on state transforms
```js
export function map_<X>(fn: (state: X) => X): (arr: X[]) => X[] {
  return arr => arr.map(e => fn(e));
}
```

dariaterekhovaae (approved):

heshan0131 (approved):

PR #1717, opened 2022-02-17
[Feat]: Technical: Translate redusers (map-state and map-style) to typescript

HeimEndyd:


src/reducers/map-state-updaters.ts
```diff
@@ -159,12 +198,17 @@ export const resetMapConfigUpdater = state => ({
...
  * @public
  */
 export const receiveMapConfigUpdater = (
-  state,
-  {payload: {config = {}, options = {}, bounds = null}}
+  state: MapState,
+  {
+    // @ts-expect-error
+    payload: {config = {}, options = {}, bounds = null}
+  }: {
```
heshan0131, line 208: return type of this function should be `MapState`

heshan0131 (approved):

src/reducers/map-style-updaters.ts
```diff
@@ -43,12 +42,66 @@ import {loadMapStyles, loadMapStyleErr} from 'actions/map-style-actions';
...
+};
+
+export type MapboxStyleUrl = string;
+
+export type BaseMapStyle = {
+  id: string;
+  label: string;
+  url: string;
+  icon: string;
+  style?: Object;
```
dariaterekhovaae, line 67: Please avoid Object and any types for future

dariaterekhovaae (approved):

PR #1722, opened 2022-02-21
UN-12 Technical: Translate redusers (main files) to typescript

HeimEndyd:


src/reducers/index.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Uber Technologies, Inc.
```
heshan0131, line 1: the `d.ts` file is converted and the `reducers/index.js` file is deleted here

HeimEndyd: Actual transformations with the index file can be viewed in the second commit, with the name "[Moved types from d.ts files](https://github.com/keplergl/kepler.gl/pull/1722/commits/86d7c441e9095287080ab0cc6dfbe5c9f050e990)"

dariaterekhovaae (approved):

src/reducers/index.ts
HeimEndyd: I changed the composition of the PR, now the changes to the index file are displayed correctly.

PR #1729, opened 2022-03-02
[Chore]: Components/common 1st part

dariaterekhovaae:


src/components/common/error-boundary.tsx
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
+import React, { ErrorInfo } from 'react';
```
heshan0131, line 21: hmm the white space is added here. Does the prettier not been ran?

dariaterekhovaae: Looks like current configuration of prettier checked only js files, I'll update it and add all of the changes to this pr

dariaterekhovaae: I added all of the ts and tsx files corrected by prettier to this pr, so the files number a little bit increased

src/components/common/modal.tsx
```diff
@@ -137,19 +140,24 @@ export const ModalFooter = ({cancel, confirm, cancelButton, confirmButton}) => {
...
-    close: PropTypes.bool,
-    onConfirm: PropTypes.func,
-    onCancel: PropTypes.func,
-    confirmButton: PropTypes.object,
-    confirmButtonLabel: PropTypes.string,
-    cancelButton: PropTypes.object,
-    cancelButtonLabel: PropTypes.string,
-    cssStyle: PropTypes.arrayOf(PropTypes.any)
-  };
+interface ModalDialogProps {
```
heshan0131, line 143: export this?

src/components/common/icons/base.tsx
```diff
@@ -80,6 +82,7 @@ export default class Base extends Component {
         height={svgHeight}
         style={style}
         className={`${predefinedClassName} ${className}`}
+        onClick={() => {}}
```
heshan0131, line 85: define this function outside:
```
function nop() {};
onClick={nop}
```

src/components/common/line-chart.tsx
```diff
@@ -111,12 +146,11 @@ function LineChartFactory() {
           <HorizontalGridLines tickTotal={3} />
           <LineSeries
             style={{fill: 'none'}}
```
heshan0131, line 148: why is scrokeWidth removed?

src/components/common/line-chart.tsx
```diff
@@ -111,12 +146,11 @@ function LineChartFactory() {
...
-            strokeWidth={2}
             color={color}
             data={series}
-            onNearestX={enableChartHover ? onMouseMove : null}
+            onNearestX={enableChartHover ? onMouseMove : undefined}
           />
-          <MarkSeries data={hoveredDP ? [hoveredDP] : []} color={color} size={3} />
+          <MarkSeries data={hoveredDP ? [hoveredDP] : []} color={color} />
```
heshan0131, line 153: why is `size={3}` removed?

src/components/common/range-plot.tsx
```diff
@@ -82,7 +102,6 @@ export default function RangePlotFactory(RangeBrush, HistogramPlot, LineChart) {
         range={range}
         value={value}
         width={width}
```
heshan0131, line 104: rangeBrush uses `height` in `getHandlePath`

src/components/common/range-slider.tsx
```diff
@@ -186,12 +206,12 @@ export default function RangeSliderFactory(RangePlot) {
...
             if (e.key === 'Enter') {
               update(e);
-              ref.current.blur();
+              (ref.current as any).blur();
             }
           }}
           onBlur={update}
           flush={key === 'value0'}
-          size={this.props.inputSize}
+          elementSize={this.props.inputSize}
```
heshan0131, line 214: can we just use `inputSize` instead here?

src/components/common/time-range-slider.tsx
```diff
@@ -30,11 +30,15 @@ import TimeRangeSliderTimeTitleFactory from 'components/common/time-range-slider
...
 
-const StyledSliderContainer = styled.div`
+interface StyledSliderContainerProps {
+  isEnlarged?: boolean;
+}
+
+const StyledSliderContainer = styled.div<StyledSliderContainerProps>`
   align-items: flex-end;
-  display: flex;
+  edisplay: flex;
```
heshan0131, line 39: typo

src/components/common/time-range-slider.tsx
```diff
@@ -30,11 +30,15 @@ import TimeRangeSliderTimeTitleFactory from 'components/common/time-range-slider
...
   flex-direction: row;
-  justify-content: space-between;
+  eejustify-content: space-between;
```
heshan0131, line 41: typo

heshan0131 (approved):
Some small issues. Please address all before landing

src/components/common/line-chart.tsx
dariaterekhovaae: According to types from @types/react-vis package LineSeries doesn't have such property

src/components/common/line-chart.tsx
dariaterekhovaae: The same as previous: according to types from @types/react-vis package MarkSeries doesn't have such property

src/components/common/icons/base.tsx
```diff
@@ -73,13 +75,16 @@ export default class Base extends Component {
     const fillStyle =
       Array.isArray(colors) && totalColor && getStyleClassFromColor(totalColor, colors);
 
+    const nop = () => {};
```
heshan0131, line 78: Sorry, I mean outside of the component scope. in render it will start gets instantiated every render.

Can we put it in defaultProps?


```js
 static defaultProps = {
    height: null,
    onClick: nop
}

```

src/components/common/line-chart.tsx
heshan0131: It is in the code, the type definition might be wrong
https://github.com/uber/react-vis/blob/master/packages/react-vis/src/plot/series/line-series.js#L80

src/components/common/line-chart.tsx
heshan0131: it is used, and also in documentation http://uber.github.io/react-vis/documentation/series-reference/mark-series

the type might be wrong

PR #1732, opened 2022-03-04
[Chore]: Technical: Translate utils (color and data)

HeimEndyd:


src/utils/color-utils.ts
```diff
@@ -38,7 +41,7 @@ export function hexToRgb(hex) {
   return [r, g, b];
 }
 
-export function isHexColor(hex) {
+export function isHexColor(hex: string) {
```
heshan0131, line 44: Use type guard here `isHexColor(hex:string): hex is HexColor`

src/utils/data-utils.ts
```diff
@@ -336,11 +343,10 @@ export const arrayMove = (array, from, to) => {
 /**
  * Get the value format based on field and format options
  * Used in render tooltip value
- * @type {typeof import('./data-utils').getFormatter}
  * @param format
  * @param field
  */
-export function getFormatter(format, field) {
+export function getFormatter(format: any, field?: Field): FieldFormatter {
```
heshan0131, line 349: `format: string | Record<string, string>`

heshan0131 (approved):

src/utils/color-utils.ts
HeimEndyd: this function works differently, it returns a RegExpExecArray used to get the RGB component of the hex color. 
Should I refactor this?

heshan0131: oh okay, nvm. disregard my comment. this should `isHexColor(hex: string): RGBColor | null`

heshan0131 (approved):

PR #1748, opened 2022-03-21
[Chore]: Technical: Translate layer factory

HeimEndyd:


src/constants/color-ranges.ts
```diff
@@ -22,6 +22,15 @@ import colorbrewer from 'colorbrewer';
 import {HexColor} from 'reducers';
 import {VizColorPalette} from './custom-color-ranges';
 
+export type ColorRange = {
+  name: string;
+  type: string;
+  category: string;
+  colors: HexColor[];
+  reversed?: boolean;
+  colorMap?: Map<string, HexColor> | object;
```
heshan0131, line 31: add 
```js
colorLegends?: Record<HexColor, string>;
```

src/layers/base-layer.ts
```diff
@@ -182,8 +180,7 @@ class Layer<LayerConfig extends LayerBaseConfig = LayerBaseConfig> {
   id: string;
   // TODO: define meta
   meta: {};
-  // TODO: define visConfigSettings
-  visConfigSettings: {};
+  visConfigSettings: Partial<LayerVisConfigSettings>;
```
heshan0131, line 183: Each layer has different propertues in visConfig and visConfigSettings. We should make the type here more generic and add specific types for individual layer's `layer.visConfigSettings`

```js
visConfigSettings: Record<string, VisConfig>;
```

heshan0131: then in ArcLayer:

```
type ArcLayerVisConfigSettings = {
  opacity: VisConfigNumber;
  thickness: VisConfigNumber;
  colorRange: VisConfigColorRange;
  sizeRange: VisConfigRange;
  targetColor: VisConfigColorSelect;
};

export default class ArcLayer extends Layer {
  visConfigSettings: ArcLayerVisConfigSettings;
...
}
```

HeimEndyd: I have considered your suggestions. This is how I see the implementation of ArcLayer

`type ArcLayerVisConfigSettings = {
  opacity: VisConfigNumber;
  thickness: VisConfigNumber;
  colorRange: VisConfigColorRange;
  sizeRange: VisConfigRange;
  targetColor: VisConfigColorSelect;
};

export default class ArcLayer extends Layer {
  declare visConfigSettings: ArcLayerVisConfigSettings;
...
}`

src/layers/base-layer.ts
```diff
@@ -178,13 +176,14 @@ function* generateColor(): Generator<RGBColor> {
...
-class Layer<LayerConfig extends LayerBaseConfig = LayerBaseConfig> {
+class Layer {
   id: string;
   // TODO: define meta
   meta: {};
-  // TODO: define visConfigSettings
-  visConfigSettings: {};
-  config: LayerConfig;
+  visConfigSettings: {
+    [key in keyof Partial<LayerVisConfig>]: ValueOf<LayerVisConfigSettings>;
```
heshan0131, line 184: Can we just put `[key: string]` here? people can create custom layers and assign custom visConfig props that is not part of the standard `LayerVisConfig`. This should be allowed by BaseLayer

PR #1749, opened 2022-03-21
[Chore]: Technical: Translate arc-layer

HeimEndyd:


src/layers/arc-layer/arc-layer.ts
```diff
@@ -52,6 +109,10 @@ export const arcVisConfigs = {
 };
 
 export default class ArcLayer extends Layer {
+  declare visConfigSettings: ArcLayerVisConfigSettings;
+  declare config: ArcLayerConfig;
```
heshan0131, line 113: declare `ArcLayerCongi` and `ArcLayerMeta` here

src/layers/arc-layer/arc-layer.ts
```diff
@@ -18,15 +18,72 @@
...
+};
+
+export type ArcLayerVisualChannelConfig = LayerColorConfig & LayerSizeConfig;
+export type ArcLayerConfig = Merge<
+  LayerBaseConfig,
+  {columns: ArcLayerColumnsConfig; visConfig: ArcLayerVisConfig}
+> &
+  ArcLayerVisualChannelConfig;
+
+export type ArcLayerData = {
```
heshan0131, line 75: define type of `ArcLayerData`

PR #1750, opened 2022-03-21
[Chore]: Technical: Translate components/common final part

dariaterekhovaae:


src/components/common/animation-control/animation-controller.ts
```diff
@@ -47,14 +58,8 @@ function AnimationControllerFactory() {
    * Current time is a point. An array of sorted time steps are provided,
    * animate a moving point jumps to the next step
    */
-  class AnimationController extends Component {
+  class AnimationController extends Component<AnimationControllerProps> {
     // TODO: convert the entire component to use hooks in the next PR
-    static defaultProps = {
```
heshan0131, line 52: is there a reason we can't keep `defaultProps`? I find it a lot more sane then adding values wherever we deconstruct `this.props`

src/components/common/animation-control/animation-controller.ts
```diff
@@ -118,8 +124,9 @@ function AnimationControllerFactory() {
     };
 
     _resetAnimtionByTimeStep = () => {
+      const {steps = []} = this.props
```
heshan0131, line 127: default value for `steps` should `null`
add below to this function
```js
if (!steps) return;
```

src/components/common/animation-control/animation-controller.ts
```diff
@@ -200,7 +207,7 @@ function AnimationControllerFactory() {
     }
 
     _nextFrameByTimeStep() {
-      const {steps, value} = this.props;
+      const {steps = [], value} = this.props;
```
heshan0131, line 210: same as above default value for `steps` should be `null`

src/components/common/animation-control/animation-controller.ts
dariaterekhovaae: The problem of defaultProps is that typescript doesn't see it. So sometimes I should duplicate it with deconstruction default values... But I'll keep it for the cases where it's not necessary

PR #1751, opened 2022-03-22
[Chore]: Technical: Setup for different Visual channels per layer

HeimEndyd:
I decided to use a generic class for typing settings in all classes that inherit. 
For example, this is how the "AggregationLayer" declaration will look like:

`export default class AggregationLayer extends Layer(LayerBaseConfig & LayerColorConfig & LayerSizeConfig)`
() - Must be angle brackets

### DIFF
--- a/src/layers/base-layer.ts
+++ b/src/layers/base-layer.ts
@@ -72,6 +72,7 @@ export type LayerColumns = {
 };
 export type VisualChannelDomain = number[] | string[];
 export type VisualChannelField = Field | null;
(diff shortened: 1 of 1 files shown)


src/layers/base-layer.ts
```diff
@@ -96,6 +97,27 @@ export type LayerBaseConfig = {
   };
 };
 
+export type LayerColorConfig = {
```
heshan0131, line 100: there is also `LayerStrokeColorConfig`

heshan0131 (approved):

PR #1753, opened 2022-03-27
[Chore]: Technical: Translate aggregation-layer

HeimEndyd:


src/layers/aggregation-layer.ts
```diff
@@ -19,17 +19,23 @@
...
   HIGHLIGH_COLOR_3D,
   CHANNEL_SCALES,
   FIELD_OPTS,
   DEFAULT_AGGREGATION
-} from 'constants/default-settings';
+} from '../constants/default-settings';
+import {Datasets} from '../reducers';
 
-export const pointPosAccessor = ({lat, lng}) => dc => d => [
+export const pointPosAccessor = ({lat, lng}: LayerColumns) => dc => d => [
```
heshan0131, line 38: define specific `AggregationLayerColumns` here

```js
type AggregationLayerColumns = {
  lat: LayerColumn;
  lng: LayerColumn; 
}

```

heshan0131 (approved):

Issue #1758, opened 2022-03-28
[Bug][Jupyter Widget] add_data with GeoDataFrame only works if the geometry column is called "geometry"


remi-sap:
**Describe the bug**
add_data using a GeoDataFrame fails if the geometry column is not called "geometry".
The geometry is often called "geometry" but this is not a given.

**To Reproduce**

```python
import pandas as pd
import geopandas
import keplergl
from shapely.geometry import Point
df = pd.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia'],
     'Country': ['Argentina', 'Brazil'],
     'PT': [Point(-34.58,-58.66), Point(-15.78, -47.91)]
})
point_gdf = geopandas.GeoDataFrame(df, geometry='PT')
w1 = keplergl.KeplerGl(height=500)
w1.add_data(data=point_gdf, name='cities')
```
Error Message:
`AttributeError: 'DataFrame' object has no attribute 'geometry'` 
with a stack pointing to the file keplergl.py line 44:
`df[name] = df.geometry.apply(lambda x: shapely.wkt.dumps(x))`

**Expected behavior**
No error message

**Environment (please complete the following information):**
 - Python version: 3.9
 - keplergl version 0.3.2

**Additional context**
Can be fixed with 1 key stroke :-)
replace 
`df[name] = df.geometry.apply(lambda x: shapely.wkt.dumps(x))`
by
`df[name] = gdf.geometry.apply(lambda x: shapely.wkt.dumps(x))`

I'll submit a pull request :-)

remi-sap:
Pull request for the fix by changing `df`to `gdf` is [here](https://github.com/keplergl/kepler.gl/pull/1759)

PR #1759, opened 2022-03-28
[jupyter] Fix for issue 1758

remi-sap:
add_data with GeoDataFrame should work regardless of the name of the geometry column


remi-sap:
Closes #1758

igorDykhta:

PR #1791, opened 2022-04-18
[Chore]: Technical: Translate deckgl-layers/layer-utils typesfix

HeimEndyd:


src/deckgl-layers/layer-utils/cpu-aggregator.ts
```diff
@@ -349,22 +380,21 @@ export const defaultElevationDimension: DimensionType<number> = {
...
-      updateSteps: UpdateStepsType[];
-      accessor: string;
-      getPickingInfo;
-      getSubLayerAccessor;
-    };
-  };
-  aggregationUpdater: {};
-
-  constructor(opts: {initialState?: CPUAggregatorState; dimensions?; aggregation?} = {}) {
+  dimensionUpdaters: DimensionType[];
```
heshan0131, line 388: the original type is correct `dimensionUpdaters` is a key value map

src/deckgl-layers/layer-utils/cpu-aggregator.ts
```diff
@@ -383,19 +413,17 @@ export default class CPUAggregator {
       },
       ...opts.initialState
     };
-    this.dimensionUpdaters = {};
-    this.aggregationUpdater = {};
 
-    this._addDimension(opts.dimensions || defaultDimensions);
-    this._addAggregation(opts.aggregation || defaultAggregation);
+    this.dimensionUpdaters = opts.dimensions || defaultDimensions;
```
heshan0131, line 417: remove the `_addDimension` will change the `dimensionUpdaters` to an array instead of an object

heshan0131 (approved):

heshan0131 (approved):

PR #1801, opened 2022-04-25
[Chore]: Technical: js to ts convertion components root modals

dariaterekhovaae:


src/components/common/dataset-label.tsx
```diff
@@ -31,15 +32,12 @@ const DatasetName = styled.div.attrs({
 `;
 
 interface DatasetLabelType {
-  dataset: {
```
btford, line 34: Is this deletion correct?

src/components/common/dataset-label.tsx
```diff
@@ -31,15 +32,12 @@ const DatasetName = styled.div.attrs({
...
-    color?: string;
-    label?: string;
-  };
+  dataset: KeplerTable;
 }
 
 const DatasetLabel = ({dataset}: DatasetLabelType) => (
   <CenterFlexbox>
-    <DatasetSquare className="dataset-clolor" color={dataset.color} />
+    <DatasetSquare className="dataset-clolor" backgroundColor={dataset.color} />
```
btford, line 40: Is the rename for this prop intentional? I see the implementation was also updated, but is this desired?

src/components/modals/add-map-style-modal.tsx
```diff
@@ -283,7 +288,7 @@ function AddMapStyleModalFactory() {
                       key={this.state.reRenderKey}
                       width={MapW}
                       height={MapH}
-                      mapStyle={inputStyle.url}
+                      mapStyle={inputStyle.url ? inputStyle.url : undefined}
```
btford, line 291: This looks wrong.

src/components/modals/export-map-modal/export-html-map.tsx
```diff
@@ -60,9 +67,26 @@ const BigStyledTile = styled(StyledType)`
   }
 `;
 
-function ExportHtmlMapFactory() {
+type IgnoreReturn<T extends (...args: any) => any> = T extends (...args: infer R) => any
```
btford, line 70: Is this pattern desirable?

src/components/modals/save-map-modal.tsx
```diff
@@ -73,13 +76,41 @@ const StyledSaveMapModal = styled.div.attrs({
...
+  ) => void;
+};
+
+export const MapInfoPanel: React.FC<MapInfoPanelProps> = ({
   mapInfo = {description: '', title: ''},
   characterLimits,
   onChangeInput
 }) => (
   <div className="selection map-info-panel">
-    <StyledModalSection className="save-map-modal-name">
```
btford, line 82: Is this deletion intentional?

src/components/common/dataset-label.tsx
dariaterekhovaae: Looks like param named "color" is used by library and has different type, so there was a name conflict. And I decided that renaming is the easiest way to avoid the conflict there

src/components/common/dataset-label.tsx
heshan0131: let't keep the original minimum type, so that `DatasetLabel` can be used as a generic component that does NOT require dataset to be a `KeplerTable`

src/components/modals/export-map-modal/export-html-map.tsx
```diff
@@ -26,6 +26,9 @@ import {EXPORT_HTML_MAP_DOC, EXPORT_HTML_MAP_MODES_DOC} from 'constants/user-gui
 import styled from 'styled-components';
 import {injectIntl} from 'react-intl';
 import {FormattedMessage} from 'localization';
+import {IntlShape} from 'react-intl';
+
+import {setUserMapboxAccessToken, setExportHTMLMapMode} from 'actions';
```
heshan0131, line 31: add a helper to `src/actions.ts` then used it to type all the action handlers in component
```js
export type ActionHandler<A extends (...args: any) => any> = (...args: Parameters<A>) => void;
```

use it here like this
```js
import {setUserMapboxAccessToken, setExportHTMLMapMode} from 'actions';

type ExportHtmlMapProps = {
  onChangeExportMapHTMLMode: ActionHandler<typeof setExportHTMLMapMode>;
  onEditUserMapboxAccessToken: ActionHandler<typeof setUserMapboxAccessToken>;
  ...
};
```

src/components/modals/export-map-modal/export-html-map.tsx
```diff
@@ -60,9 +67,22 @@ const BigStyledTile = styled(StyledType)`
   }
 `;
 
-function ExportHtmlMapFactory() {
+type ExportHtmlMapProps = {
+  onChangeExportMapHTMLMode: typeof setExportHTMLMapMode;
```
heshan0131, line 71:
```js
 onChangeExportMapHTMLMode: ActionHandler<typeof setExportHTMLMapMode>;
  onEditUserMapboxAccessToken: ActionHandler<typeof setUserMapboxAccessToken>;
```

src/components/modals/export-map-modal/export-map-modal.tsx
```diff
@@ -28,30 +27,34 @@ import {StyledExportMapSection} from './components';
...
-  options: PropTypes.object,
-  onEditUserMapboxAccessToken: PropTypes.func.isRequired,
-  onChangeExportData: PropTypes.func,
-  onChangeExportMapType: PropTypes.func,
-  mapFormat: PropTypes.string
-};
+interface ExportMapModalFactoryProps {
+  options?: {format: string};
+  config: any;
+  onEditUserMapboxAccessToken: typeof setUserMapboxAccessToken;
```
heshan0131, line 35:
```js
onEditUserMapboxAccessToken: ActionHanlder<typeof setUserMapboxAccessToken>;
onChangeExportMapHTMLMode?: ActionHandler<typeof setExportHTMLMapMode>;
```

heshan0131 (approved):

PR #1807, opened 2022-04-28
[Chore]: Technical: add types for side panel common

dariaterekhovaae:


src/components/side-panel/common/dataset-info.tsx
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
```
heshan0131, line 1: why is this recognized as new file? instead of rename

jagerts: fixed

heshan0131 (approved):

PR #1815, opened 2022-05-06
[Chore]: Technical: Translate deckgl-layers/cluster-layer

HeimEndyd:


src/deckgl-layers/cluster-layer/cluster-layer.ts
```diff
@@ -23,16 +23,22 @@ import {_AggregationLayer as AggregationLayer} from '@deck.gl/aggregation-layers
...
 
-import {LAYER_VIS_CONFIGS} from 'layers/layer-factory';
-import {SCALE_TYPES} from 'constants/default-settings';
-import {DEFAULT_COLOR_RANGE} from 'constants/color-ranges';
+import {LAYER_VIS_CONFIGS} from '../../layers/layer-factory';
+import {SCALE_TYPES} from '../../constants/default-settings';
+import {DEFAULT_COLOR_RANGE} from '../../constants/color-ranges';
 import ClusterBuilder, {getGeoJSON} from '../layer-utils/cluster-utils';
+import {RGBAColor} from '../../reducers';
+import {AggregationLayerProps} from '@deck.gl/aggregation-layers/aggregation-layer';
```
heshan0131, line 39: is this valid import from deck.gl? I think without a path resolver, you would always want to make deck.gl export `AggregationLayerProps` at top level and import from `'@deck.gl/aggregation-layers`

heshan0131 (approved):

PR #1820, opened 2022-05-12
Enable serving framed kepler.gl-jupyter over https

LangdalP:
Remove HTTP protocol specifier for stylesheets used in `bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js` to select the most appropriate protocol.

This is needed when kepler.gl is served on a website served via HTTPS. Today, this fails since chrome blocks loading of insecure resources when the page is served via HTTPS (Mixed content). However, Firefox and Safari are not as strict as chrome, so this fix is not needed for those browsers.

See image below to see what happened before this fix, included the error messages in the dev console at the bottom.

image

### DIFF
--- a/bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js
+++ b/bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js
@@ -100,11 +100,11 @@ function App() {
           <title>Kepler.gl Jupyter</title>
           <link
             rel="stylesheet"
-            href="http://d1a3f4spazzrp4.cloudfront.net/kepler.gl/uber-fonts/4.0.0/superfine.css"
+            href="//d1a3f4spazzrp4.cloudfront.net/kepler.gl/uber-fonts/4.0.0/superfine.css"
           />
           <link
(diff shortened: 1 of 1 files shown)


bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js
```diff
@@ -100,11 +100,11 @@ function App() {
           <title>Kepler.gl Jupyter</title>
           <link
             rel="stylesheet"
-            href="http://d1a3f4spazzrp4.cloudfront.net/kepler.gl/uber-fonts/4.0.0/superfine.css"
+            href="//d1a3f4spazzrp4.cloudfront.net/kepler.gl/uber-fonts/4.0.0/superfine.css"
```
heshan0131, line 103: Maybe just change this to `https://` instead?

afterthought: What is holding up this PR? I have an https jupyterhub environment that would benefit from this fix.

igorDykhta:

PR #1837, opened 2022-05-26
validate url of Add data modal

jagerts:
- add removed styles of Add data modal
- add validation to url input


examples/demo-app/src/utils/url.js
```diff
@@ -66,3 +66,11 @@ parseUri.options = {
     loose: /^(?:(?![^:@]+:[^:@\/]*@)([^:\/?#.]+):)?(?:\/\/)?((?:(([^:@]*)(?::([^:@]*))?)?@)?([^:\/?#]*)(?::(\d*))?)(((\/(?:[^?#](?![^?#\/]*\.[^?#\/.]+(?:[?#]|$)))*\/?)?([^?#\/]*))(?:\?([^#]*))?(?:#(.*))?)/
   }
 };
+
+/**
+ * Validates an url
+ * @param str
+ */
+export function validateUrl(str) {
+  return str?.match(/^(ftp|https?):\/\/+(www\.)?[a-z0-9\-\.]{3,}\.[a-z]{3}$/);
```
heshan0131, line 75: should also allow `http://`

examples/demo-app/src/utils/url.js
```diff
@@ -66,3 +66,11 @@ parseUri.options = {
     loose: /^(?:(?![^:@]+:[^:@\/]*@)([^:\/?#.]+):)?(?:\/\/)?((?:(([^:@]*)(?::([^:@]*))?)?@)?([^:\/?#]*)(?::(\d*))?)(((\/(?:[^?#](?![^?#\/]*\.[^?#\/.]+(?:[?#]|$)))*\/?)?([^?#\/]*))(?:\?([^#]*))?(?:#(.*))?)/
   }
 };
+
+/**
+ * Validates an url
+ * @param str
+ */
+export function validateUrl(str) {
+  return str?.match(/^(ftp|http|https?):\/\/+(www\.)?[a-z0-9\-\.]{3,}\.[a-z]{3}$/);
```
eduramirezh, line 75: This regex only allows URLs with an empty path (allows `https://uber.com` but not `https://uber.com/data.csv`).

Additionally, it removes the option of using localhost as a domain.
And TLDs with only 2 letters (.es, .cl, etc).

I think this PR should be reverted

eduramirezh: @jagerts @heshan0131

PR #1845, opened 2022-06-17
Fix demo freeze in $2+ Million Homes Built in Los Angeles Since 2006

jagerts:
Signed-off-by: Evgeny Zhgulev <evgeny.zhgulev@actionengine.com>

### DIFF
--- a/src/layers/scenegraph-layer/scenegraph-layer.ts
+++ b/src/layers/scenegraph-layer/scenegraph-layer.ts
@@ -114,8 +114,6 @@ export const scenegraphVisConfigs: {
   }
 };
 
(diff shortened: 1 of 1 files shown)


jagerts:
@heshan0131 

I removed default model usage, so now it doesn't freeze app when we choose 3DModel in "$2+ Million.." demo.

But if we will upload same Model via picker in side panel, it will freeze the app again.

I've tried to slice the data from dataset (currently this demo has over 200k nodes). In case we have only 1k nodes, the app doesn't freeze. So looks like this bug is about performance.

So, should we remove the default Model or may be reduce the dataset size for this demo..?

PR #1856, opened 2022-07-11
[Chore]: Technical: Isolate layers

dariaterekhovaae:


src/layers/src/aggregation-layer.ts
```diff
@@ -27,18 +27,18 @@ import Layer, {
   VisualChannelDescription,
   VisualChannels
 } from './base-layer';
-import {hexToRgb} from '../utils/color-utils';
-import {aggregate} from '../utils/aggregate-utils';
+import {hexToRgb} from 'utils/color-utils';
```
heshan0131, line 30: why is the relative path gets removed?

PR #1862, opened 2022-07-18
Chore: Fix lint script and issues

nrabinowitz:
As far as I can tell, `yarn lint` has not been checking files in `src` since [this change](https://github.com/keplergl/kepler.gl/commit/10868ecfb371e7547bf63daa205bdd62ebdd2c29#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63). This fixes the `yarn lint` script and addresses the lint failures it reveals.

Some of these issues can be resolved by upgrading `eslint`, `prettier`, and/or `eslint-plugin-prettier`, but for now this just gets lint green.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -44,6 +44,14 @@
     "consistent-return": 0,
     "comma-dangle": 1
   },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {
+        "no-undef": "off",
+      },
+    },
+  ],
   "settings": {
     "react": {
       "version": "detect"
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,6 +27,7 @@
     "max-len": 0,
     "react/no-did-mount-set-state": 0,
     "react/no-multi-comp": 0,
+    "react/sort-comp": 0,
     "no-use-before-define": "off",
     "prefer-spread": 1,
     "prefer-template": 1,
@@ -48,7 +49,8 @@
     {
       "files": ["*.ts", "*.tsx"],
       "rules": {
-        "no-undef": "off",
+        "no-undef": 0,
+        "no-unused-vars": 0,
       },
     },
   ],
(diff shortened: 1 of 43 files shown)


.eslintrc
```diff
@@ -43,6 +43,14 @@
...
     "comma-dangle": 1
   },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {
+        "no-undef": "off",
+      },
+    },
+  ],
```
nrabinowitz, line 53: See https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/TROUBLESHOOTING.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

package.json
```diff
@@ -61,7 +61,7 @@
     "typedoc": "typedoc --theme markdown --out typedoc --inputFiles ./src/reducers --inputFiles ./src/actions --excludeExternals --excludeNotExported --excludePrivate",
     "example-version": "babel-node ./scripts/edit-version.js",
     "prettier-all": "prettier --write '{src,examples,test}/**/*.{js,tsx,ts}'",
-    "lint": "yarn typescript && eslint src/*/src/*.{js,tsx,ts}  test webpack examples/*.js examples/**/src/*.{js,tsx,ts} website/*.js website/src --fix",
+    "lint": "yarn typescript && eslint src/*.js src/**/*.{js,tsx,ts} test webpack examples/*.js examples/**/src/*.{js,tsx,ts} website/*.js website/src --fix",
```
nrabinowitz, line 64: Note that for some reason just using `src` here caused `yarn lint` to run for several hours before I killed it - not sure why, but this seemed preferable.

nrabinowitz:
The build failed lint with 220 additional errors, which I don't understand. Need to investigate why my local env is so different from CI, and then decide whether it's worth fixing all those errors now.

jwasilgeo (approved):
I was able to `yarn`, `yarn lint`, and `yarn start` without any issues that I could identify. LGTM

heshan0131 (approved):

PR #1882, opened 2022-07-27
[Bug] Disable layer copy when layer is invalid

jwasilgeo:
- add allowDuplicate prop to layer header
- disable duplicate if layer is not valid

### DIFF
--- a/src/components/side-panel/layer-panel/layer-panel-header.tsx
+++ b/src/components/side-panel/layer-panel/layer-panel-header.tsx
@@ -65,6 +65,7 @@ type LayerPanelHeaderProps = {
   showRemoveLayer?: boolean;
   label?: string;
   layerType?: string | null;
+  allowDuplicate?: boolean;
   isDragNDropEnabled?: boolean;
   labelRCGColorValues?: RGBColor | null;
   actionIcons?: {
@@ -145,8 +146,10 @@ const StyledPanelHeaderHiddenActions = styled.div.attrs({
   opacity: 0;
   display: flex;
   align-items: center;
-  background-color: ${props => props.theme.panelBackground};
-  transition: ${props => props.theme.transition};
+  background-color: ${props =>
+    props.isConfigActive ? props.theme.panelBackgroundHover : props.theme.panelBackground};
+  transition: opacity 0.4s ease, background-color 0.4s ease;
+
   :hover {
     opacity: 1;
   }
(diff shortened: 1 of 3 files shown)


src/components/side-panel/layer-panel/layer-panel-header.tsx
```diff
@@ -138,15 +139,21 @@ const HeaderActionSection = styled.div<HeaderActionSectionProps>`
...
+})<StyledPanelHeaderHiddenActionsProps>`
   opacity: 0;
   display: flex;
   align-items: center;
-  background-color: ${props => props.theme.panelBackground};
-  transition: ${props => props.theme.transition};
+  background-color: ${props =>
+    props.isConfigActive ? props.theme.panelBackgroundHover : props.theme.panelBackground};
+  transition: opacity 0.4s ease, background-color 0.4s ease;
+
```
jwasilgeo, line 156: @heshan0131 these lines may need extra attention since there are theming improvements, but this upstreaming introduced some specific conditional formatting.

src/components/side-panel/layer-panel/layer-panel.tsx
```diff
@@ -140,6 +140,7 @@ function LayerPanelFactory(
       const {layer, datasets, isDraggable, layerTypeOptions} = this.props;
       const {config} = layer;
       const {isConfigActive} = config;
+      const allowDuplicate = typeof layer.isValidToSave === 'function' && layer.isValidToSave();
```
jwasilgeo, line 143: Originally this line was
```js
const allowDuplicate = typeof layer.isValidToSave === 'function' && layer.isValidToSave() && isValid;
```
but `isValid` now doesn't exist on the layer type definition. @heshan0131 do you think this diff will work as intended?

heshan0131: make a note in `[feat] Disable a layer after an error in Deck (#180)` to address it when cherrypick

jwasilgeo: ✅

src/components/side-panel/layer-panel/layer-panel-header.tsx
heshan0131: looks for to me. I made it so the transition only applies to opacity and background color

jwasilgeo: Makes sense, thanks for taking a look. Do you feel this PR is ready to go?

heshan0131 (approved):

PR #1902, opened 2022-08-04
chore: Updated filter-selector, item-selector, range-slider file typescript definitions

jwasilgeo:
upstreaming: "chore: Updated filter-selector, item-selector, range-slider file typescript definitions"


jwasilgeo:
@macrigiuseppe the upstreaming was mainly `.d.ts` file changes. I attempted to move over what I could by hand manually.

heshan0131 (approved):

PR #1927, opened 2022-08-15
[Chore] add prettier-check to git actions

igorDykhta:
- add prettier-check to git actions

- [ ] too many changes due to prettier, postpone pr
- [ ] bump prettier to support new ts version


package.json
```diff
@@ -68,7 +68,8 @@
     "docs": "babel-node ./scripts/documentation.js",
     "typedoc": "typedoc --theme markdown --out typedoc --inputFiles ./src/reducers --inputFiles ./src/actions --excludeExternals --excludeNotExported --excludePrivate",
     "example-version": "babel-node ./scripts/edit-version.js",
-    "prettier-all": "prettier --write '{src,examples,test}/**/*.{js,tsx,ts}'",
+    "prettier": "prettier --write '{src,examples,test}/**/*.{js,tsx,ts}'",
+    "run-prettier": "prettier-check '{src,examples,test}/**/*.js'",
```
heshan0131, line 72: now that we are moving to submodules, this should be changed to to include all the submodules src. and not the dist

`'{src,examples,test}/**/*.{js,tsx,ts}' '!src/*/dist/**/*.js'`

PR #1959, opened 2022-08-28
[Chore] bump deck to 8.6.0

igorDykhta:


heshan0131 (approved):

chrisgervang:
The `yarn.lock` lists deck 8.8. Does that mean package.json can be bumped to 8.8 as well?

igorDykhta:
@chrisgervang I don't think it should be 8.8 atm, but so far I haven't noticed any issues. The version should be backward compatible, but I frequently encounter deck/luma issues when we have multiple different versions requested by different modules.

PR #1967, opened 2022-09-01
[Chore]: Technical: Isolate components

dariaterekhovaae:


src/components/src/common/animation-control/floating-time-display.tsx
```diff
@@ -21,9 +21,9 @@
 import React, {useMemo} from 'react';
 import styled from 'styled-components';
 import classnames from 'classnames';
-import {Minus} from 'components/common/icons';
+import {Minus} from '../icons';
 import {DEFAULT_TIME_FORMAT} from '@kepler.gl/constants';
-import {CenterFlexbox} from 'components/common/styled-components';
+import {CenterFlexbox} from '@kepler.gl/constants';
```
heshan0131, line 26: styled-componets moved to constants? that seems a little odd

src/layers/src/icon-layer/icon-info-modal.tsx
```diff
@@ -21,9 +21,8 @@
 import React from 'react';
 import styled from 'styled-components';
 import {line} from 'd3-shape';
-import {CenterFlexbox} from 'components/common/styled-components';
 import {FormattedMessage} from '@kepler.gl/localization';
-import {Table} from 'components/common/icons';
+import {CenterFlexbox, Table} from '@kepler.gl/constants';
```
heshan0131, line 25: instead of moving `CenterFlexbox` to constant, can you just copy the implementation here. so we don't need to have layers depends on `CenterFlexbox`?

src/layers/src/scenegraph-layer/scenegraph-info-modal.tsx
```diff
@@ -20,7 +20,7 @@
 
 import React from 'react';
 import styled from 'styled-components';
-import {StyledTable as Table} from 'components/common/styled-components';
+import {StyledTable as Table} from '@kepler.gl/constants';
```
heshan0131, line 23: looks like the only place `StyledTable` is used is  in screnegraph-layer, you should just move the implementation in here, instead of moving it to constant

src/layers/src/cluster-layer/cluster-layer-icon.tsx
```diff
@@ -20,7 +20,7 @@
 
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import Base from 'components/common/icons/base';
+import {Base} from '@kepler.gl/constants';
```
heshan0131, line 23: can you make a copy of `Base` and put it in`layers/src/react` and keep the `Base` in `components/icons`? so that we don't move it to constant?

src/constants/src/icons/index.ts
```diff
@@ -0,0 +1,9 @@
+export {default as Base} from './base';
+// eslint-disable-next-line prettier/prettier
+export type {BaseProps} from './base';
+export {default as Upload} from './upload';
+export {default as Table} from './table';
```
heshan0131, line 5: Move table into @kepler.gl/layers

src/constants/src/icons/index.ts
```diff
@@ -0,0 +1,9 @@
+export {default as Base} from './base';
+// eslint-disable-next-line prettier/prettier
+export type {BaseProps} from './base';
+export {default as Upload} from './upload';
```
heshan0131, line 4: Move upload into `@kepler.gl/providers`

src/layers/src/cluster-layer/cluster-layer-icon.tsx
dariaterekhovaae: `Base` is imported by every other icon. So if we want at less one icon to be in `@kepler.gl/constants` then we need `Base` to be there as well

src/components/src/map/map-popover.tsx
```diff
@@ -22,12 +22,13 @@ import React, {useState} from 'react';
...
 import CoordinateInfoFactory from './coordinate-info';
-import {ArrowLeft, ArrowRight, Pin} from 'components/common/icons';
+import {Pin, ArrowLeft, ArrowRight} from '../common/icons';
 import {injectIntl, IntlShape} from 'react-intl';
 import {FormattedMessage} from '@kepler.gl/localization';
 import Tippy from '@tippyjs/react/headless';
-import {RootContext} from 'components';
+import {RootContext} from '../';
 import {LayerHoverProp} from '@kepler.gl/reducers';
+import {} from '@kepler.gl/constants';
```
heshan0131, line 31: remove this line?

test/helpers/mock-state.js
```diff
@@ -57,7 +57,7 @@ import {
 import tripGeojson, {tripDataInfo} from 'test/fixtures/trip-geojson';
 import {processCsvData, processGeojson} from '@kepler.gl/processors';
 import {MOCK_MAP_STYLE} from './mock-map-styles';
-import {getUpdateVisDataPayload} from 'components/geocoder-panel';
+import {} from '@kepler.gl/components';
```
heshan0131, line 60: delete this line?

heshan0131 (approved):

PR #1976, opened 2022-09-13
[Fix] "load from storage" and "Share" modals fix

dariaterekhovaae:

### DIFF
--- a/examples/webpack.config.local.js
+++ b/examples/webpack.config.local.js
@@ -69,7 +69,7 @@ function makeLocalDevConfig(env, EXAMPLE_DIR = LIB_DIR, externals = {}) {
   workspaces.forEach(workspace => {
     // workspace =  "./src/types",  "./src/constants", etc
     const moduleName = workspace.split('/').pop();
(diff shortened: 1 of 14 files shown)


examples/webpack.config.local.js
```diff
@@ -69,7 +69,7 @@ function makeLocalDevConfig(env, EXAMPLE_DIR = LIB_DIR, externals = {}) {
   workspaces.forEach(workspace => {
     // workspace =  "./src/types",  "./src/constants", etc
     const moduleName = workspace.split('/').pop();
-    resolveAlias[`@kepler.gl/${moduleName}`] = join(SRC_DIR, `${moduleName}/src`);
+    resolveAlias[`@kepler.gl/${moduleName}`] = join(SRC_DIR, `${moduleName}/dist`);
```
heshan0131, line 72: this function makes local dev setup, we should loads file from `src` instead of `dist`, changing this will not affect the production build

heshan0131 (approved):

PR #1991, opened 2022-09-26
[Feat] allow changing dataset color with color picker

igorDykhta:


igorDykhta:
This functionality is already in os kepler, upstreaming it to add a diff

heshan0131 (approved):

PR #1992, opened 2022-09-26
[Chore] onClose handler for dataset's color picker

igorDykhta:


igorDykhta:
Everything else is already in os kepler

heshan0131 (approved):

PR #2007, opened 2022-10-14
[Feat] Replaced filter enlarged with view: side | enlarged | minified

igorDykhta:


src/actions/src/vis-state-actions.ts
```diff
@@ -758,20 +759,24 @@ export function toggleLayerAnimationControl(): Merge<
 
 export type EnlargeFilterUpdaterAction = {
```
heshan0131, line 760: should this be renamed to `SetFilterViewUpdaterAction`

heshan0131 (approved):

PR #2035, opened 2022-11-18
[fix] include CenterFlexbox in common components

epeterson320:
Not sure if this should be considered [fix] or not, but upgrading from 2.5.5 seems to have broke importing `<CenterFlexbox>`.

Signed-off-by: Eric Peterson <git@ericp.co>


igorDykhta (changes requested):
You need to export CenterFlexbox from src/components/src/common as well

igorDykhta (approved):

PR #2049, opened 2022-12-01
[fix] correct provider downloadMap type

epeterson320:
This fixes the return type of the `downloadMap()` method in `Provider`.

It was `Promise<{ map: SavedMap; format: string }>`, but `SavedMap` itself has a `map` property. When an implementation of `downloadMap()` returns an object like `{ map: { map: { ... } }, format: ... }`, loading maps works, but the "Save"/"Save As" functionality breaks. I haven't figured out why, but I tested manually and confirmed that flattening the nested `map` property fixes the re-saving, without breaking the loading.

Signed-off-by: Eric Peterson <git@ericp.co>

### DIFF
--- a/src/cloud-providers/src/provider.ts
+++ b/src/cloud-providers/src/provider.ts
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import Upload from './upload';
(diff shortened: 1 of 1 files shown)


igorDykhta (approved):
Correct. I've checked dropbox-provider and it returns {map: SavedMap; format: string}>

PR #2078, opened 2023-01-06
[Feat] Add an `array` field type

### LINKED ISSUE #1637: Filters : multiple attributes per field?
In my data layer, i have some fields who contain an array with multiple category
for example : `["Non-hazardous waste", "Hazardous waste", "Inert waste"]`

The data parsing system  does not seem to handle multiple category in the same field

I havent found an alternative yet in the documentation but maybe there is a workaround ?

lutangar:
> fix #1637 and supersede #1666

Allow to import features with multiple values properties and convert them to an `array` type. 

These fields can then be picked as filters or shown in the tooltip.

### DIFF
--- a/src/constants/src/default-settings.ts
+++ b/src/constants/src/default-settings.ts
@@ -329,6 +329,7 @@ export const ALL_FIELD_TYPES = keyMirror({
   boolean: null,
   date: null,
   geojson: null,
+  array: null,
   integer: null,
   real: null,
   string: null,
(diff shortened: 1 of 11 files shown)


lutangar:
> __Review required__
> At least 1 approving review is required by reviewers with write access. 

Hello there, could you kindly have a look at this ? I'd be happy to give this some more work if needed.

lutangar:
Superseded by #2120

PR #2079, opened 2023-01-09
[Feat] refactored AnimationControl to handle both layer and minified filter playback

igorDykhta:
- [x] fix ts inconsistencies
- [x] test
- [ ] time.ts type issues


src/utils/src/time.ts
```diff
@@ -0,0 +1,101 @@
...
+    // @ts-expect-error not in anim
+    timeSteps,
+    defaultTimeFormat,
+    timeFormat,
+    timezone
+  } = animationConfig;
+
+  return {
+    // @ts-expect-error
+    value: toArray(currentTime),
```
heshan0131, line 58: @macrigiuseppe is this correct? if the type of `Timeline` is `number | [number, number]`, shouldn't it be `value: currentTime` instead?

src/utils/src/time.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export const getTimelineFromAnimationConfig = (animationConfig: AnimationConfig): Timeline => {
+  const {
+    currentTime,
+    domain,
+    speed,
+    isAnimating,
+    // @ts-expect-error not in anim
+    timeSteps,
```
heshan0131, line 50: can you add `timeSteps` type of `AnimationConfig` as optional prop?

heshan0131 (approved):

src/utils/src/time.ts
macrigiuseppe: given the current type this is correct. the Timeline structure is done to represents either a filter or an animation config. I do think we should enforce value to be an array of two elements because we expect to be as such in 

aninmation-controller.js
```
 _resetAnimationByDomain = () => {
      const {domain, value, animationWindow} = this.props;
      // interim solution while we fully migrate filter and layer controllers
      const setTimelineValue = this.props.updateAnimation || this.props.setTimelineValue;
      if (Array.isArray(value)) {
        if (animationWindow === ANIMATION_WINDOW.incremental) {
          setTimelineValue([value[0], value[0] + 1]);
        } else {
          setTimelineValue([domain[0], domain[0] + value[1] - value[0]]);
        }
      } else {
        setTimelineValue(domain[0]);
      }
    };
```

PR #2115, opened 2023-01-30
[Feat] Add list toggle to filters

igorDykhta:
- create a AddByDatasetButton
- Display flter by list or dataset
- Add filter button shows typeahead and dataset selection
- Move add buttons to right of title to filter / base map panel


src/components/src/side-panel/side-bar.tsx
```diff
@@ -24,14 +24,14 @@ import {ArrowRight} from '../common/icons';
 
 type CollapseButtonProps = {
   isOpen: boolean;
-  onClick: () => void;
+  onClick: (e: React.MouseEvent<HTMLDivElement>) => void;
 };
 
 type SideBarProps = {
```
heshan0131, line 30: this type should be exported from `components/index.ts`

src/components/src/side-panel/panel-tab.tsx
```diff
@@ -37,7 +37,7 @@ export type PanelItem = {
 type PanelTabProps = {
```
heshan0131, line 37: this type should be exported from components/index.ts

src/components/src/index.ts
```diff
@@ -100,6 +100,7 @@ export {default as DatasetTagFactory} from './side-panel/common/dataset-tag';
 
 export {default as FilterManagerFactory} from './side-panel/filter-manager';
 export {default as FilterPanelFactory} from './side-panel/filter-panel/filter-panel';
+export {default as AddFilterButtonFactory} from './side-panel/filter-panel/add-filter-button';
```
heshan0131, line 103: missing some type export here:

```
MapStyleSelectorProps,
LayerGroupSelectorProps,
MapManagerProps,
SourceDataCatalogProps,
SourceDataSelectorProps,
DatasetTitleProps,
DatasetInfoProps,
DatasetTagProps,
CustomPanelsProps
```

src/components/src/side-panel/common/types.ts
```diff
@@ -1,8 +1,21 @@
...
+};
+
+export type PanelMeta = {
+  id: string;
+  label: string;
+  iconComponent: React.ElementType;
+  onClick: null;
+};
+
 export type DatasetInfoProps = {
```
heshan0131, line 19: this should be 
`dataset: MiniDataset`?

heshan0131 (approved):

src/components/src/side-panel/common/types.ts
igorDykhta: As I remember it caused ts issues with related components, as something was using fields from KeplerDataset, not available in MiniDataset. I'll check again.

heshan0131: If that's case, we should add `fields: MiniField[]` to `MiniDataset`

PR #2122, opened 2023-01-31
[Chore] export LayerTypeListItem type

igorDykhta:

### DIFF
--- a/src/components/src/index.ts
+++ b/src/components/src/index.ts
@@ -241,9 +241,6 @@ export {
   LayerTitleSectionFactory
 } from './side-panel/layer-panel/layer-panel-header';
 
-import {LayerTypeListItemType} from './side-panel/layer-panel/layer-type-list-item';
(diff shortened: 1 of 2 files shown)


src/components/src/index.ts
```diff
@@ -241,6 +241,9 @@ export {
   LayerTitleSectionFactory
 } from './side-panel/layer-panel/layer-panel-header';
 
+import {LayerTypeListItemType} from './side-panel/layer-panel/layer-type-list-item';
+export type {LayerTypeListItemType};
```
heshan0131, line 245: `export type {LayerTypeListItemType} from '../side-panel/layer-panel/layer-type-list-item';`

heshan0131 (approved):

src/components/src/index.ts
igorDykhta: yeah, tried this one as well.
for some reason moving this line down fixes the breaking issue from prettier.

PR #2144, opened 2023-03-03
[fix] improves handling of "interpolate" mapbox colors during basemap switching

igorDykhta:
- Fixes automatic background color selection when switching to "No Basemap" from a basemap that has a mapbox "interpolate" color for the background layer.


ilyabo (approved):
lgtm 👍

PR #2164, opened 2023-03-22
[Feat] enhance mouse selection toolset

igorDykhta:
The basic idea is to extend the current map selection toolset so that users can do more with mouse selection.


ilyabo (approved):
lgtm 👍

PR #2170, opened 2023-03-26
[Feat] Unlocked split map viewports 

igorDykhta:
## PR Changes

- Adds the ability to have "unsynced" viewports when in split map mode. They can also optionally be set to lock to each other's zooms (regardless of lat/lon or 3D pitch, bearing, etc).
- New switch toggle controls for this were added to the legend panel.
- Works in 2D/3D.
- Changes MapPopover behavior to only ever show 1 instance of the MapPopover. It is only shown in the map that the mouse is hovering over, regardless of split map mode or single map mode.
- Adds dual geocoders for split + unsynced viewports.


src/components/src/map/map-control-panel.tsx
```diff
@@ -97,41 +120,76 @@ function MapControlPanelFactory() {
...
+      return (
+        <StyledMapControlPanel
+          className="map-control-panel"
+          style={{
+            transform: `scale(${scale})`,
+            marginBottom: '8px !important'
+          }}
+        >
+          {mapState?.isSplit ? (
+            <StyledMapControlPanelHeaderSplitViewportsTools>
```
jwasilgeo, line 145: Thanks for bringing this one in, @igorDykhta. When I test the deployed PR demo or run it locally myself, I can't get these conditional controls to render. If I force this to `true` it would look like this in a split maps view:

Screenshot 2023-03-26 at 10 46 33 PM

I'm not sure why yet, but in the demo app code it seems like the parents of this component might now be passing around props that are a flattened `mapState` instead of when this was originally written to pass in the full `mapState`. 🤔 Either way, `mapState` seems to evaluate to `undefined` now when I try to debug.

igorDykhta: @jwasilgeo should be fixed

PR #2172, opened 2023-03-28
[Feat] Drag and drop interaction for split map

igorDykhta:
1. replace [react-sortable-hoc](https://github.com/clauderic/react-sortable-hoc) with [dnd-kit](https://github.com/clauderic/dnd-kit). React-sortable-hoc is no longer actively maintained and its developers strongly encourage new consumers to adopt dnd-kit instead of adopting react-sortable-hoc.
2. Add support for the split map mode, allowing dragging and dropping layers from the side panel to each of the two map panels.

### DIFF
--- a/src/components/package.json
+++ b/src/components/package.json
@@ -32,6 +32,10 @@
   "dependencies": {
     "@deck.gl/core": "^8.6.0",
     "@deck.gl/react": "^8.6.0",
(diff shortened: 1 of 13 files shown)


ilyabo (approved):
lgtm 👍 🚀

PR #2174, opened 2023-03-28
[fix] hide side panel close button when data preview is open

igorDykhta:
- added a new showCollapseButton boolean type flag in uiState.


ilyabo (approved):
lgtm 👍

PR #2179, opened 2023-03-30
[feat] Create layer correctly from saved layer config

igorDykhta:
This PR ensures we load the layer correctly when using `createLayerFromConfig`. We can now pass a parsed and loaded layer to the method

### DIFF
--- a/test/node/reducers/vis-state-test.js
+++ b/test/node/reducers/vis-state-test.js
@@ -1186,6 +1186,175 @@ test('#visStateReducer -> DUPLICATE_LAYER', t => {
   t.end();
 });
 
+// TODO: not sure if this test is correct, it basically checks if the loaded visState is empty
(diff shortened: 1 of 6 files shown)


test/node/reducers/vis-state-test.js
```diff
@@ -1186,6 +1186,175 @@ test('#visStateReducer -> DUPLICATE_LAYER', t => {
   t.end();
 });
 
+// TODO: not sure if this test is correct, it basically checks if the loaded visState is empty
+// if we load configuration through the schema manager all layers will be changing from having visualChannels to visualConfig
```
ilyabo, line 1190: Not sure where the test comes from, but maybe the goal is to make sure that an invalid layer wouldn't be loaded?

ilyabo (approved):

PR #2182, opened 2023-03-30
[fix] support Polygon and LineString mode in idToPolygonGeo

igorDykhta:
- support Polygon and LineString mode in idToPolygonGeo


ilyabo (approved):
lgtm 👍

PR #2183, opened 2023-03-30
[Fix] intervals rendered incorrectly in time widget

igorDykhta:
- fix E.g. The loaded data has datetime column within one day. However, when specify Interval “24 hours”, the blue area in the selection area is rendered incorrectly outside.


ilyabo (approved):
lgtm 👍

PR #2199, opened 2023-04-16
[Feat] Add display format setting for table/tooltip

igorDykhta:
Add a global “display formatter” that will be used to format the numeric and date-time values.

This global fomatter will be used as a “default formatter”, so users can set up and enable/disable the display format quickly. Besides, users will be able to customize the display format for each component on top of the global formatter (e.g. customize the display format for each column in the data table)

* Table View
  * User can set display format for all integer/float/date columns at once
  * User can overwrite display format for each column

* Tooltip
  * Use field.displayFormat instead of visState.interactionConfig.tooltips
  * User can overwrite display format for each field


ilyabo (approved):
lgtm 👍

PR #2201, opened 2023-04-18
[fix] Empty cells with date time data are filled with 'Invalid date'

igorDykhta:
Empty cells with date time data are filled with 'Invalid date' --> A empty cell should be displayed


ilyabo (approved):
👍

PR #2231, opened 2023-05-21
[Fix] Feature Action Panel menu and editing tooltip are cut-off in dual map mode 

igorDykhta:
1) shape selection/filter menu showing under the map on the right when in dual map mode. 
The fix uses a react portal to change where that menu is inserted in the DOM. Instead of being nested inside the map on the left's DOM, it's appended to our context. That way it displays over both maps instead of being hidden by the left map.

2) switch the tooltip's position that shows on hover for shapes on the map. Too close to the right side - show the tooltip on the cursor's left side, etc.

### DIFF
--- a/src/components/src/editor/editor.tsx
+++ b/src/components/src/editor/editor.tsx
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 import React, {Component, CSSProperties, KeyboardEvent} from 'react';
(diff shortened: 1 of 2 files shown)


ilyabo (approved):
lgtm 👍 

Side note: maybe we should consider using [floating-ui](https://floating-ui.com/) for tooltips and popovers  in kepler as well like in Studio

PR #2252, opened 2023-06-13
[feat] H3 Layer: default text label anchor to middle position

igorDykhta:
- set H3 layer default text label anchor to middle
- update vis state updaters to rely on a layer's default text label config (if present) when adding a new text label config


ilyabo (approved):
lgtm 👍

PR #2321, opened 2023-09-18
[Feat] Add support to Arrow/GeoArrow format

lixun910:
### Description

GeoParquet is a file format while GeoArrow is a memory format. Both can be saved as a file e.g. .parquet and .arrow. Arrow memory is both zero-copy and has constant-time access, so it could be a very efficient memory format that allows different programs (javascript, C++, WebAssembly, Rust, Python) to exchange data.

This PR is to support Arrow/GeoArrow format in Kepler.gl specifically to `addDataToMap()` function. For example, with this PR, one can call the following code to add a GeoArrow file to kepler.gl:

``` js
    // use @loaders.gl/arrow to load and parse the arrow file
     parse(fetchFile(arrowUrl), ArrowLoader, {
        worker: false,
        arrow: {shape: 'arrow-table'}
      }).then((arrowTable: ApacheArrowTable) => {
        const parsedData = processArrowTable(arrowTable);
        keplerGlDispatch(
          addDataToMap({
            datasets: {data: parsedData},
            options: {}
          })
        );
      });
```
For geometry columns, the raw arrow data will be preserved in visState.datasets, and it can be used to e.g. pass raw data to WebAssembly programs or webworkers efficiently. When rendering the geometries, this PR adds the function `parseGeometryFromArrow()` to convert arrow geometry to GeoJSON feature for `GeoJsonLayer` in Kepler.gl. 

This is unfortunate because there is another "copy" of the raw geometries in the format of GeoJSON, and the deck.gl could convert these GeoJSON features back to binary raw geometries again for GPU rendering (see: https://deck.gl/docs/api-reference/layers/geojson-layer#using-binary-data). (Maybe this is worth another PR to e.g. add a new ArrowLayer).

The arrow geometry types supported: (see https://github.com/geoarrow/geoarrow/blob/main/extension-types.md#extension-names):
- [x] geoarrow.wkb
- [x] geoarrow.wkt
- [x] geoarrow.point
- [x] geoarrow.multipoint
- [x] geoarrow.linestring
- [x] geoarrow.multilinestring
- [x] geoarrow.polygon
- [x] geoarrow.multipolygon

performance:
arrow vs geojson vs csv

### Other: support drag-n-drop a GeoParquet/GeoArrow file in Kepler.gl

The current version of kepler.gl uses loaders.gl/arrow v3. However, in this version, the arrow loader couldn't return the correct data: the raw arrow data of each arrow column is returned without the metadata to kepler.gl. This has been fixed in the latest loaders.gl/arrow v4 (see: https://github.com/visgl/loaders.gl/blob/2577ca735878b521f07a556f26ce8ee457a7ad9f/modules/arrow/src/lib/parse-arrow-in-batches.ts#L29).

To support drag-n-drop a GeoParquet/GeoArrow file in Kepler.gl, there are two more tasks that need to be done in other PRs:
- upgrade loaders.gl to v4
- support dnd GeoParquet/GeoArrow file in file-handler.ts


src/utils/src/arrow-utils.ts
```diff
@@ -0,0 +1,341 @@
...
+ * @returns dataset containing `fields` and `rows` or null
+ */
+export function processArrowTable(arrowTable: ArrowTable): ProcessorResult | null {
+  if (!arrowTable) {
+    return null;
+  }
+  const metadata = arrowTable.schema.metadata;
+  // get geometry columns if metadata has key 'geo'
+  let geometryColumns = [];
+  if (metadata.get(ARROW_GEO_METADATA_KEY) !== undefined) {
```
kylebarron, line 77: Note that this is a geoparquet construct, not a geoarrow construct. As of yet, GeoArrow has no table-level metadata. The GeoArrow metadata sets extension metadata on a specific _field_, which you can check by reading `ARROW:extension:name` https://github.com/geoarrow/deck.gl-layers/blob/509547593e50cfd79aa9bc784970679747ea8acc/src/utils.ts#L33-L44

lixun910: Thanks, @kylebarron! Will update by checking `ARROW:extension:name` for each field.

lixun910:
Close this: as suggested by Ib, I will create another PR to add related functions in loaders.gl.

src/utils/src/arrow-utils.ts
kylebarron: Well, anytime you have data that comes from parquet, it will have the `geo` key. So it's not that you _shouldn't_ check the `geo` key, just to be clear that "pure geoarrow" data won't have that

PR #2323, opened 2023-09-19
[chore] Upgrade to react 18

igorDykhta:
- Upgrade to react 18


ilyabo:
The React upgrade might bring deck/mapbox syncing lag. Do you experience it after the upgrade?  If so, we should probably also upstream https://github.com/foursquare/kepler.gl/pull/723

jwasilgeo (approved):

PR #2333, opened 2023-09-22
[Bug] Allow adding data from cloud providers without private storage

paperfront:
* Pass all cloud providers to the load data modal
* Add test
* Add mock provider that doesn't support private storage


paperfront:
This issue came up in the context of trying to add a cloud provider that provides datasets to users of the app, but that doesn't support users saving their personal data to that provider. In the current code, only providers with private storage are being passed to the load-data-modal, but it seems that this is a bug

igorDykhta:
Closing
- All cloudProviders are passed unfiltered into a CloudListProvider

PR #2385, opened 2023-10-20
[Feat] support GeoArrow format

lixun910:
### Description

GeoArrow support in Kepler.gl will enable efficient loading of big data. For example, loading 1 million polygons takes ~2 seconds with Arrow format vs. ~20 seconds with GeoJson format:

arrow-polygons

Screenshot 2023-10-20 at 10 59 59 AM

GeoParquet is a file format while GeoArrow is a memory format. Both can be saved as a file e.g. .parquet and .arrow. Arrow memory is both zero-copy and has constant-time access, so it could be a very efficient memory format that allows different programs (javascript, C++, WebAssembly, Rust, Python) to exchange data.

### Details

* A new Geoarrow layer

The GeoJsonLayer in deck.gl already has the capability of loading binary geometries directly, so this pull request inherits an ArrowLayer from kepler.gl's `GeoJsonLayer`.

The geometry types supported: (see https://github.com/geoarrow/geoarrow/blob/main/extension-types.md#extension-names):
- [x] geoarrow.point
- [x] geoarrow.multipoint
- [x] geoarrow.linestring
- [x] geoarrow.multilinestring
- [x] geoarrow.polygon
- [x] geoarrow.multipolygon

Picking and table view are supported.

* A column-wise data container

This PR adds a new column-wise data container, ArrowDataContainer, which implements the DataContainerInterface. This container is designed to efficiently use the data structure of the Arrow format.

* CPU Filtering

The GPU filtering in GeojsonLayer is compatible with ArrowLayer. For CPU filtering, to avoid filtering on the raw Arrow table and make a partial copy of the raw table, this PR adds a simple deck.gl layer extension to filter the Geoarrow layer based on the result of CPU filteredIndex. This could impact other functions that rely on a filtered dataset. Please help to check. Thanks!

### Other: support drag-n-drop a GeoParquet/GeoArrow file in Kepler.gl

The current version of kepler.gl uses loaders.gl/arrow v3. However, in loaders.gl/arrow version < 4.0.0, the arrow loader in batch didn't return the correct data. Instead, the raw arrow data of each arrow column is returned and stored directly (without the metadata) to kepler.gl. This has been fixed in latest loaders.gl/arrow v4 (see: https://github.com/visgl/loaders.gl/blob/2577ca735878b521f07a556f26ce8ee457a7ad9f/modules/arrow/src/lib/parse-arrow-in-batches.ts#L29).

One can call `processArrowTable()` directly to add arrow data to Kepler, e.g.: 
``` js
import {Table as ApacheArrowTable} from 'apache-arrow';

fetchFile(arrowUrl)
  .then(response => response.arrayBuffer())
  .then(buffer => {
    const arrowTable = ApacheArrowTable.from([buffer]);
    const parsedData = processArrowTable(arrowTable);
    this.props.dispatch(
      addDataToMap({
        datasets: {data: parsedData}
      });
  });
```

To support drag-n-drop a GeoParquet/GeoArrow file in Kepler.gl, there are two more tasks need to be done:
- [x] upgrade loaders.gl to v4
- [x] support dnd GeoParquet/GeoArrow file in file-handler.ts

### Test arrow files:

[flights.arrow.zip](https://github.com/GeoDaCenter/kepler.gl/files/13049616/flights.arrow.zip)
[polygons.arrow.zip](https://github.com/GeoDaCenter/kepler.gl/files/13049618/polygons.arrow.zip)

One can use ogr2ogr to convert e.g. GeoJson file to Arrow file:
```sh
ogr2ogr test.arrow test.geojson -f Arrow
```


lixun910:
@kylebarron Hi Kyle, could you also help to take a look? Thanks!

src/utils/src/dataset-utils.ts
```diff
@@ -582,6 +587,80 @@ export function analyzerTypeToFieldType(aType: string): string {
...
+  }
+}
+
+/**
+ * Convert arrow data type to analyzer type
+ *
+ * @param arrowType the arrow data type
+ * @returns corresponding type in `AnalyzerDATA_TYPES`
+ */
+export function arrowDataTypeToAnalyzerDataType(
```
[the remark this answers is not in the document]
lixun910: I think type-analyzer is infering data types from CSV data, so it is not a good place for this function.

lixun910:
Use GeoArrow utils from loaders.gl/arrow v4.0.1. Support drag-n-drop GeoArrow file in kepler.gl, see the screen recording:

arrow-kepler

ibgreen (approved):

src/layers/src/arrow-layer/arrow-layer.ts
```diff
@@ -0,0 +1,351 @@
...
+    }
+    if (lat > newBounds[3]) {
+      newBounds[3] = lat;
+    }
+  }
+
+  return newBounds;
+}
+
+export default class ArrowLayer extends GeoJsonLayer {
```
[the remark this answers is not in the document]
ibgreen: GeoArrowLayer?

webpack/umd.js
```diff
@@ -108,6 +108,11 @@ const LIBRARY_BUNDLE_CONFIG = env => ({
             ]
           ]
         }
+      },
+      {
+        test: /\.mjs$/,
```
ibgreen, line 113: Comment

ibgreen (approved):

jest.setup.js
```diff
@@ -30,3 +30,14 @@ jest.mock('@kepler.gl/utils', () => ({
   hasPortableWidth: jest.fn(),
   hasMobileWidth: jest.fn()
 }));
+
+// TextEncoder / TextDecoder APIs are used by arrow, but are not provided by
+// jsdom, all node versions supported provide these via the util module
+if (
```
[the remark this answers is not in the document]
lixun910: Good point! Yes, this can be replaced by adding `require('@loaders.gl/polyfilles')` in jest.setup.js

src/processors/src/file-handler.ts
```diff
@@ -50,6 +61,44 @@ const JSON_LOADER_OPTIONS = {
...
+  progress?: {rowCount?: number; rowCountInBatch?: number; percent?: number};
+  metadata?: Map<string, string>;
+};
+
+/**
+ * check if table is an ArrowTable object
+ * @param table - object to check
+ * @returns {boolean} - true if table is an ArrowTable object type guarded
+ */
+export function isArrowTable(table: any): table is ArrowTable {
```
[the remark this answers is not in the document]
lixun910: All functions in arrow-utils.ts have been moved to loaders.gl, so maybe it's not worth adding a new `arrow-utils.ts` file for this function?

lixun910:
@ibgreen @heshan0131 see a test of progressive rendering using GeoArrow, loaders.gl v4 in kepler.gl
https://github.com/GeoDaCenter/kepler.gl/pull/5

src/layers/src/geoarrow-layer/geoarrow-layer.ts
```diff
@@ -0,0 +1,330 @@
...
+    }
+    if (lat > newBounds[3]) {
+      newBounds[3] = lat;
+    }
+  }
+
+  return newBounds;
+}
+
+export default class GeoArrowLayer extends GeoJsonLayer {
```
heshan0131, line 42: Any reason we can't use the existing geojson layer but created a new layer type to support geoarrow? All the configuration option is the same.

lixun910: Thanks, Shan! After moving the arrow utils to loaders.gl, I think we could handle the arrow format as a special case in Kepler's GeoJsonLayer (or making GeoJsonLayer arrow compatible).  Let me think about it.

src/layers/src/base-layer.ts
```diff
@@ -1428,6 +1428,16 @@ class Layer {
     // implemented in subclasses
     return () => null;
   }
+
+  getColumnAccessor(dataContainer?: DataContainerInterface): (...args: any[]) => any {
```
ibgreen, line 1432: Do the return types need to be `any`? Can we be more precise?

src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -183,15 +193,31 @@ export default class GeoJsonLayer extends Layer {
   declare config: GeoJsonLayerConfig;
   declare visConfigSettings: GeoJsonVisConfigSettings;
   declare meta: GeoJsonLayerMeta;
-  dataToFeature: GeojsonDataMaps;
+
+  dataToFeature: GeojsonDataMaps | BinaryFeatures[];
+  dataContainer: DataContainerInterface | null;
+
+  isArrow: boolean;
```
ibgreen, line 200: Nit: Move static initializations out of constructor?
```suggestion
  isArrow: boolean = false;
```

src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -399,8 +455,26 @@ export default class GeoJsonLayer extends Layer {
     return this;
   }
 
+  isLayerHovered(objectInfo: any): boolean {
+    return this.isArrow
+      ? isLayerHoveredFromArrow(objectInfo, this.id)
+      : super.isLayerHovered(objectInfo);
```
ibgreen, line 461: Does this mean that we could push the isArrow checks down into `super` and get support in all layers?

src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -429,11 +504,15 @@ export default class GeoJsonLayer extends Layer {
     const pickable = interactionConfig.tooltip.enabled;
     const hoveredObject = this.hasHoveredObject(objectHovered);
 
-    return [
-      new DeckGLGeoJsonLayer({
+    const {data, ...props} = dataProps;
+    const deckLayerData = this.isArrow ? data : [data];
```
ibgreen, line 508: - Why is non-arrow data now wrapped in an array?
- The goal must be to remove the use of `isArrow`, presumably rely on container as much as possible?

src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -429,11 +504,15 @@ export default class GeoJsonLayer extends Layer {
...
+    const deckLayers = deckLayerData.map((d, i) => {
+      return new DeckGLGeoJsonLayer({
         ...defaultLayerProps,
         ...layerProps,
-        ...data,
+        ...props,
+        data: d,
+        id: this.isArrow ? `${this.id}-${i}` : this.id,
```
ibgreen, line 515: Why an index? Is this related to batching? Maybe that should be a separate PR?

src/layers/src/layer-utils.ts
```diff
@@ -34,3 +37,83 @@ export function assignPointPairToLayerColumn(pair: FieldPair, hasAlt: boolean) {
...
+  // there could be multiple deck.gl layers created from multiple chunks in arrow table
+  // the objectInfo.layer id should be `${this.id}-${i}`
+  if (objectInfo?.picked) {
+    const deckLayerId = objectInfo?.layer?.props?.id;
+    return deckLayerId.startsWith(layerId);
+  }
+  return false;
+}
+
+export function getHoveredObjectFromArrow(
```
ibgreen, line 81: Could part of this (getting an object from a table) somehow go into data container interface?

src/processors/src/data-processor.ts
```diff
@@ -388,21 +392,56 @@ export function processKeplerglDataset(
...
+export function processArrowTable(arrowBatches: arrow.RecordBatch[]): ProcessorResult | null {
+  if (arrowBatches.length === 0) {
+    return null;
+  }
+  const arrowTable = new arrow.Table(arrowBatches);
+  const fields: Field[] = [];
+
+  // parse fields
+  arrowTable.schema.fields.forEach((field: arrow.Field, index: number) => {
+    const isGeometryColumn = field.metadata.get('ARROW:extension:name')?.startsWith('geoarrow');
```
ibgreen, line 410: Nit: This means we have knowledge about the GeoArrow extensions both here and in loader.gl/utils? Maybe hard to avoid duplication, but it is always good to try to centralize knowledge about some aspect in one part of the code. I.e. if we could use one of the utils from loaders...

src/processors/src/file-handler.ts
```diff
@@ -50,6 +61,44 @@ const JSON_LOADER_OPTIONS = {
   ]
 };
 
+/**
+ * ProcessFileDataContent
+ * @typedef {Object} ProcessFileDataContent
```
ibgreen, line 66: Nit: We should use typescript types instead of JSDoc types. At minimum remove the types in the comment.

Additionally if you move the docs for the fields into `/** ... */` comment above each field in the type then the vscode will show that on hover.

src/utils/src/data-container-interface.ts
```diff
@@ -60,6 +60,21 @@ export interface DataContainerInterface {
    */
   column(columnIndex: number): Generator<any, void, any>;
 
+  /**
+   * Returns the column object at the specified index.
+   * @param columnIndex Column index.
+   * @returns The column object at the specified index.
+   */
+  getColumn?(columnIndex: number): any;
```
ibgreen, line 68: Avoid return type `any` if at all possible. Usually preferable to use `unknown` and force the app to state what type it is expecting via a cast.

src/utils/src/dataset-utils.ts
```diff
@@ -586,6 +588,80 @@ export function analyzerTypeToFieldType(aType: string): string {
   }
 }
 
+/**
+ * Convert arrow data type to kepler.gl field types
+ *
+ * @param arrowType the arrow data type
+ * @returns corresponding type in `ALL_FIELD_TYPES`
+ */
+export function arrowDataTypeToFieldType(arrowType: ArrowDataType): string {
```
ibgreen, line 597: Just a thought: I wonder if this could go together with the ArrowContainer? I.e. consolidate arrow knowledge as much as possible.

ibgreen (commented):
Just re-read the code and it is looking great. Added some thoughts, but they can be addressed later or not at all.

heshan0131 (approved):

src/layers/src/base-layer.ts
lixun910: I think we don't need these placeholders in base-layer. The geoColumnAccessor and geoFieldAccessor in geojson-layer should work without these placeholders. I will try to consolidate them if needed when supporting arrow format on other layers.

src/layers/src/geojson-layer/geojson-layer.ts
lixun910: Yes, I think we can do that, so we can support arrow format in other layers. We can do that later with new PRs.

src/layers/src/geojson-layer/geojson-layer.ts
lixun910: It's because arrow table can have multiple chunks/batches, and a deck.gl layer is created for each chunk.

src/layers/src/geojson-layer/geojson-layer.ts
lixun910: Yes, this is related to batching in arrow table. Deck.gl will return this id for hovering.

src/layers/src/layer-utils.ts
lixun910: Yes, I believe this can be consolidated when working on supporting arrow in other layers.

src/processors/src/data-processor.ts
lixun910: Yes, agree. Will expose this in loaders.gl/utils and call it from here.

PR #2396, opened 2023-10-23
[fix] effect related fixes (split maps, shadows, timeline)

igorDykhta:
- enable effects only for the first view (split/swipe views)
- extend deck lighting effect to output uniform shadow during nighttime
- disable sunlight for nighttime
- fixes to animation time picker


test/browser/components/effects/effect-time-configurator-test.js
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2023 Uber Technologies, Inc.
```
ilyabo, line 1: maybe remove Uber copyright?

ilyabo (approved):

PR #2409, opened 2023-10-25
[feat] Support customized ref in useDimensions

igorDykhta:
- Update useDimensions to support customized ref

### DIFF
--- a/src/utils/src/observe-dimensions.ts
+++ b/src/utils/src/observe-dimensions.ts
@@ -125,7 +125,7 @@ export default function useDimensions<T extends Element>(
       unobserveDimensions(current);
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [throttleDelay]);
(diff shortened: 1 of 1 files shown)


src/utils/src/observe-dimensions.ts
```diff
@@ -95,9 +95,10 @@ function getSize(node, entry): Dimensions | null {
...
  */
 export default function useDimensions<T extends Element>(
+  nodeRef?: RefObject<T>,
   throttleDelay = DEFAULT_THROTTLE_DELAY
 ): [RefObject<T>, Dimensions | null] {
-  const ref = useRef<T>(null);
+  const ref = nodeRef ?? useRef<T>(null);
   const [size, setSize] = useState(null);
 
   useEffect(() => {
```
ilyabo, line 104: We should probably add `ref?.current` to the dependencies of `useEffect` so that the observers get reattached in case the ref element changes.

ilyabo (approved):

PR #2461, opened 2023-11-30
MapLibre basemap

birkskyum:
closes #2460 cc @ibgreen 

This makes the initial move to maplibre, to get the ball rolling, but it's not complete yet as outlined below. The demo app appear to work just fine. The examples folder doesn't use the local build, but instead a pinned kepler.gl version, so I haven't tried all of them yet.

New styles (and thumbnails) will be needed to replace the "uberdata" ones, since the mapbox styles with their proprietary mapbox:// syntax aren't supported. That is an area I'll need your help with completing. I made a quick darkblue style with maptiler, which is shown in the image below, and carto also has some free ones available (added the dark-matter one).

The GPU layers should work as well now in supported browsers, because maplibre v3 has webgl2 support, and a webgl1 fallback.

Screenshot 2023-11-30 at 20 24 27

PR #2463, opened 2023-11-30
pin luma.gl version to 8.5.21, to avoid mismatch

birkskyum:
Stumbled upon a version mismatch with luma as part of #2461 , because it's pinned in the "resolutions" part of package.json. Thus just update to the latest 8.x release.


ibgreen (approved):
LGTM

birkskyum:
Rebased - after loaders.gl was updated to 4.0 alpha.3 on master branch.

PR #2469, opened 2023-12-07
Remove uber-licence

chrisgervang:
For #2468

Perhaps we want to fork this project into OpenVis. Seems useful!


chrisgervang:
Going to fix my DCO

PR #2470, opened 2023-12-07
Remove uber-licence

chrisgervang:
For #2467 

Perhaps we want to fork this project into OpenVis. Seems useful!


chrisgervang:
Alternatively the CLI says it accepts a custom license.. so we may be able to keep it.

ibgreen (approved):

chrisgervang:
Rather than remove it, I'd prefer to go with copying it in #2472

chrisgervang:
Fixed in #2472

PR #2472, opened 2023-12-07
Add file license header script

chrisgervang:
For #2467. Alternatively to #2470 , we can keep the script that adds license headers


LICENSE-HEADER
```diff
@@ -0,0 +1,2 @@
+Copyright contributors to the kepler.gl project
+SPDX-License-Identifier: MIT
```
chrisgervang, line 2: Note that order matters due to an assumption made in [uber-licence](https://github.com/uber/uber-licence/blob/e7a9dc5058201e1ddc8f5d4d4ebcaca66a755b0f/license-fixer.js#L53)

chrisgervang: Forked it over, removed the parts causing issues, and made other necessary changes to get it to work (like ES modules and deps)

chrisgervang:
This is good to go. Once merged I can run `yarn run babel-node ./scripts/license-header/bin --license ./scripts/license-header/FILE-HEADER --legacy ./LICENSE` in #2471 and it'll migrate anything I missed.

chrisgervang:
Yes, but I need to do it within a few minutes of the header change PR to avoid messing up someone working on other PRs. I'll block off time today to land these.

PR #2474, opened 2023-12-11
[Fix] basemap frozen when incrementally loading GeoArrow

lixun910:
Description: It seems the basemap refreshing could be frozen when testing incremental loading GeoArrow file using https://kepler.gl/demo, which is different than the rendering behavior when testing on local dev env (see screenshot in https://github.com/keplergl/kepler.gl/pull/2459):

incremental-issue

This PR is to fix this issue by slightly delaying the nextFileBatch action to avoid the racing condition between basemap rendering and deck.gl geojson layer rendering:

incremental-issue-fix


test/node/reducers/vis-state-test.js
```diff
@@ -4924,8 +4924,11 @@ test('#visStateReducer -> LOAD_FILES', async t => {
   // test LOAD_FILE_TASK success
   const [task2, ...more2] = drainTasksForTesting();
   t.equal(more2.length, 0, 'should ceate 1 task');
-  t.equal(task2.type, 'UNWRAP', 'should return an UNWRAP task');
-  t.ok(task2.payload instanceof Promise, 'task 2 payload should be a Promise');
+  t.equal(task2.type, 'Chain(DELAY)', 'should return an Chain(DELAY) task');
+  t.equal(task2.payload, 10, 'should delay 10 millisecond of the next task');
+  // TODO: FIXME task2 is a chain task with a 10 millisecond delay and a UNWRAP_TASK task
+  // Here the payload of UNWRAP_TASK is a Promise. However, it seems we cannot test it here:
+  // t.ok(task2.payload instanceof Promise, 'task 2 payload should be a Promise');
```
lixun910, line 4931: @heshan0131 Hi Shan, do you know how to test the chained task? Thank you!

ibgreen: It may be easier to just wrap the batch iterator returned from loaders.gl in a new iterator that sleeps say 50ms between each batch. In fact I think there already is a `batchDebounceMs` option in loaders.gl that inserts such waits.

lixun910: Thanks, @ibgreen! That sounds reasonable. I just tested: adding a delay between each batch works the same. See this draft PR here: https://github.com/visgl/loaders.gl/pull/2836 But I am not sure how to pass the `batchDebounceMs` option to the `parseArrowInBatches()` function. Can you help to take a look? Thanks!

ibgreen: Looking around but getting lost in the code. Is loading going through the ArrowLoader in `file-handler.ts`? The general pattern would be to just add it to the loader options root:

```ts
load(url, [loaders...], {
   batchDebounceMs: 100,
   csv: {...}
});
```

lixun910: Thanks, Ib! I can add the option in kepler's file-handler.ts, like this:
Screenshot 2023-12-11 at 4 13 14 PM

Then, how can I pass it down to the `parseArrowInBatches()` function in loaders.gl? Should I add option parameter for `parseInBatches` here  Line 60?: 
Screenshot 2023-12-11 at 4 16 01 PM

lixun910: I think I got it workign in https://github.com/visgl/loaders.gl/pull/2836. I will update the loader.gl version for this PR once it's been merged and tagged. Thanks for your help, @ibgreen !

src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import * as arrow from 'apache-arrow';
-import {BinaryFeatures} from '@loaders.gl/schema';
+import {BinaryFeatureCollection} from '@loaders.gl/schema';
```
ibgreen, line 22: Was this module not upgraded to 4.0? If not, this looks confusing if loaders 4.0 is exporting two such similar types...

ibgreen (approved):

src/layers/src/geojson-layer/geojson-layer.ts
lixun910: I think it’s because the "@loaders.gl/schema": "^4.1.0-alpha.2" was missing in package.json. Should be fixed in this PR.

test/node/reducers/vis-state-test.js
lixun910: TODO: need a new release of loader.gl to complete this PR.

ibgreen:
Ok let me try to get a alpha.4 out sooner rather than later.

ibgreen (commented):
You want `4.1.0-alpha.4`

Issue #2475, opened 2023-12-12
[Bug] json map uploading problem


maxbobkov:
Hi! When trying to upload a valid json map into the kepler.gl/demo app, the process freezes when uploading bar progress reaches 100% and Done label appears.

1. Go to https://kepler.gl/demo
2. Try to upload any valid json map (previously exported from kepler.gl)
3. 100% uploading bar progress and nothing after that

**Expected behavior**
The uploaded map with the data should appear

**Screenshots**
<img width="1428" alt="kepler gl screenshot" src="https://github.com/keplergl/kepler.gl/assets/10603061/74d8bbbe-9a1b-4a63-8897-81bd5f01270d">

**Desktop (please complete the following information):**
I've tried different ways:
- OS: Mac OS 11.7.1/Ubuntu 20.04.6 LTS
 - Browsers: Chrome, Firefox

thanks for any help!

jtamerius:
Any idea when this might be fixed?

maxbobkov:
Thank you for fixing!

maxbobkov:
Thank you for fixing!

PR #2479, opened 2023-12-14
[fix] fix json map import 

igorDykhta:
Fix for #2475


jwasilgeo (approved):

ibgreen (approved):

lixun910:
Thanks for fixing it! I should add some test cases (later).

PR #2492, opened 2023-12-21
[chore] update readme, fix examples, show effects button

igorDykhta:
- update readme
- fix examples
- show effects button

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ overridden by the latest instance and reset to a blank state.
 
 By default, kepler.gl uses mapbox-gl.js to render its base maps. You can create a free account at [mapbox][mapbox] and create a token at [www.mapbox.com/account/access-tokens][mapbox-token].
 
-If you replaced kepler.gl default map styles with your own, and they are not Mapbox styles. `mapboxApiAccessToken` will not be reuiqred.
+If you replaced kepler.gl default map styles with your own, and they are not Mapbox styles. `mapboxApiAccessToken` will not be required.
 
 Read more about [Custom Map Styles][custom-map-styles].
 
--- a/examples/umd-client/index.html
+++ b/examples/umd-client/index.html
@@ -42,7 +42,7 @@
     <!--MapBox token-->
     <script>
       /**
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -73,7 +73,7 @@ You can add the script tag to your html file as it follows:
 or if you would like, you can load a specific version
 
 ```html
-<script src="https://unpkg.com/kepler.gl@0.2.2/umd/keplergl.min.js" />
+<script src="https://unpkg.com/kepler.gl@2.5.5/umd/keplergl.min.js" />
```
chrisgervang, line 76: v3 or continue using 2?

README.md
```diff
@@ -165,11 +160,10 @@ stored in `state.keplerGl.foo`.
 In case you create multiple kepler.gl instances using the same id, the kepler.gl state defined by the entry will be
 overridden by the latest instance and reset to a blank state.
 
-##### `mapboxApiAccessToken` (String, required*)
+##### `mapboxApiAccessToken` (String, required\*)
```
chrisgervang, line 163: Typo or intentional?

examples/umd-client/index.html
```diff
@@ -32,9 +45,9 @@
        * Provide your MapBox Token
        **/
       const MAPBOX_TOKEN = 'PROVIDE_MAPBOX_TOKEN';
```
chrisgervang, line 47: Do we use maplibre now?

chrisgervang (approved):

examples/umd-client/index.html
birkskyum: The mapbox token was used for different things

- **allow mapbox-gl library to do a renderer**. `maplibre-gl` doesn't need a token for this local operation, and can readily be used offline if need be with a set of downloaded tiles.
- **access mapbox styles from style api (and map tiles)**. MapLibre doesn't have any services. We use the free Carto styles by default which doesn't need a token. If a paid style is to be used, like Stadia maps or MapTiler, an apikey will be part of the url when it's added, so here also we don't need a similar api key.

PR #2494, opened 2023-12-21
Remove unused parts of translation files

birkskyum:
Adding a style previously requested an api key, but these fields are not active anymore since the api key is part of the style url, so the translations aren't needed either.


birkskyum:
There is a bit of confusion about how to add custom styles. This clarified the translations. There has for a long time been a bug related to recognizing some style urls, that I'd like to debug when this is in.

birkskyum:
@akre54 , this is a bit of a cleanup step related to the fix you made recently:
- The input field with the Mapbox token is removed (this was labelled step 2 of 3)
- Step 3 is renamed to Step 2.

Issue #2528, opened 2024-02-21
Tooltip not Working When Exporting Map HTML [Bug]


margootje123:
**Describe the bug**
I am running the Kepler GL Demo Locally. I have saved the HTML and then opened from my own folder. When opening the HTML locally, the tooltip function does not work. Not when I click on data on the map, and also not when I hover over the data on the map.

**To Reproduce**
1. Go to Share, Export Map, HTML, Allow users to edit map, and Export
2. Click on downloaded HTML in download folder
3. Hover over datapoints on map 
4. Nothing happens

**Expected behavior**
I expect the tooltip to pop up when hovering over the datapoints on the map.

**Desktop:**

 - Browser Chrome and Edge tried both
 - Version 3.0.0
 
<img width="864" alt="image" src="https://github.com/keplergl/kepler.gl/assets/160749297/3ced11ef-dca8-4715-b489-a0905bcd00ab">

AndriusGeo:
I just want to confirm that I also reproduce this error (windows 11, Chrome). Tooltips not working when map is exported as HTML.

I found a workaround and probably the cause of the problem. In HTML you can try to change kepler.gl version to older one (I tried 2.5.5) and for me tooltips started working:
![image](https://github.com/keplergl/kepler.gl/assets/1758436/ccdcd02e-17ad-4f45-83c0-2f8ea8c9a4a0)

Of course, this is by no means a workaround, but it does confirm (as far as I understand) that the bug is present in version 3.0.0. Maybe this will help in further investigating the error.

margootje123:
@AndriusGeo Thanks so much! Changing the version indeed resolved the issue for me as well. Seems to be a bug then with the latest version, fingers crossed this gets updated some time soon.

dafnikaravola:
Hey, I'm facing the same issue with @margootje123 (tried both Chrome & Opera), but the issue does not resolve with @AndriusGeo trick. Any other ideas?

liuwo23:
I found the issue is due to React version 

> Hey, I'm facing the same issue with @margootje123 (tried both Chrome & Opera), but the issue does not resolve with @AndriusGeo trick. Any other ideas?

This bug is caused by the inconsistency between the React version used in the project and the React version in the HTML template. I can fix it. Can someone review and approve the code?

IlyaSolomatin:
Hey, I have the same issue, any progress here?

liuwo23:
> Hey, I have the same issue, any progress here?

[...]

PR #2556, opened 2024-04-01
[Fix] Tooltip not working in exported HTML map

ilyabo:
Solves #2528 and #2536 (duplicate)

The fix updates the deps in the export map html file.

After fix, tooltip works in the exported map:

https://github.com/keplergl/kepler.gl/assets/351828/df39efdf-1c8b-4622-919d-ea7d653b19ac


ibgreen (approved):

lixiao321:
The tooltip doesn't work again in the downloaded html page. https://drlixiao.net/maps/syriacultural.html

PR #2594, opened 2024-07-23
[fix] fix for starting demo-app 

igorDykhta:
Fir for starting demo-app.

At the moment running `yarn start` produces the following error:

```
ERROR in ./node_modules/@loaders.gl/core/node_modules/@probe.gl/log/dist/utils/hi-res-timestamp.js 7:27
Module parse failed: Unexpected token (7:27)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|     let timestamp;
|     if (isBrowser() && window.performance) {
>         timestamp = window?.performance?.now?.();
|     }
|     else if ('hrtime' in process) {
```

Possibly related to #2593

### DIFF
--- a/examples/webpack.config.local.js
+++ b/examples/webpack.config.local.js
@@ -115,7 +115,11 @@ function makeLocalDevConfig(env, EXAMPLE_DIR = LIB_DIR, externals = {}) {
           test: /\.(js|ts|tsx)$/,
           use: ['source-map-loader'],
           enforce: 'pre',
-          exclude: [/node_modules\/react-palm/, /node_modules\/react-data-grid/]
+          exclude: [
+            /node_modules\/react-palm/,
+            /node_modules\/react-data-grid/,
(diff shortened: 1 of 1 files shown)


lixun910 (approved):

ibgreen (commented):
Any ideas 
- why is this suddenly needed? Is the build of the demo app not tested by CI?
- why isn't the CJS distribution in probel.gl / loaders.gl 4.x being picked up, is it some configuration issue?
- This will be a problem for all visgl libraries once we upgrade to v9 so kepler will be recompiling all of vis.gl at that point, even though CJS distributions are included.

echo292:
@ibgreen  Kepler.gl 3.0 release doesn't run as per current documentation. Do you have a list of steps to run the demo we can publish in the documentation?

Regards,

E

PR #2623, opened 2024-09-05
Fixed Webpack builds

mars-dlx:
Fix for #2593

* Updated Webpack to latest v4
* Removed duplicated babel loaders


heshan0131:
the fix for demo-app webpack config shouldn't be needed anymore after moving to esbuild. Website and kepler.gl-jupyter is still using webapck though

mars-dlx:
Rebased to master.

cc @heshan0131

PR #2634, opened 2024-09-09
[chore] Add several vis state mergers combineConfigs and improve TS

igorDykhta:
- Adds `combineConfigs` to some vis state saved configs that are non-array complex objects:
  1. `interactionConfig`
  2. `layerBlending`
  3. `overlayBlending`
  4. `editor`
  
  Each of these has inline comments (and unit tests) explaining the decisions made on how to combine values in different situations. For example, `interactionConfig.brush.size` gets aggregated into keeping the single largest value among all brush sizes.

- Improves and fixes some TS defs.

- Adds jest unit tests for newly added `combineConfigs`.


heshan0131:
Looks like there is a failed test

heshan0131 (approved):

Issue #1468, opened 2021-05-01
[Question] Label Font Bold


jlezard:
Hello 

Thanks for this cool project.

Sorry if this is not the right place to ask this.

Is there a way to set the Label Font to a Bold Font ? 

Many thanks,

Joe

macrigiuseppe:
@jlezard we are currently not supporting Bold 

@Pessimistress

Pessimistress:
@macrigiuseppe font weight can be set, though only per layer: https://deck.gl/docs/api-reference/layers/text-layer#fontweight

PR #2675, opened 2024-10-07
[Enhancement] add font weight option for Text Labels

alex-polic:
* work based on issue #1468 
* add font weight in list of settings for Text Labels
* integrate with localization
* hook with deck.gl fontWeight options for TextLabel

⚠️ FUTURE INTEGRATION NOTES:
- Currently, there is an option through deck.gl to change Font Family and Font Weight per layer, this PR only works with Font Weight
- The default font used for Text Labels in deck.gl is Monaco monospaced. This font does not provide full variety of weight options (only 2), which creates a problem visually (text doesn't seem bolded when weight is increased)
- It would be beneficial to either change the default font or allow users to use multiple fonts. Design and business input required before making adjustments

### DIFF
--- a/src/components/src/side-panel/layer-panel/text-label-panel.tsx
+++ b/src/components/src/side-panel/layer-panel/text-label-panel.tsx
@@ -97,6 +97,17 @@ function TextLabelPanelFactory(
                 </SidePanelSection>
                 <SidePanelSection>
                   <PanelLabel>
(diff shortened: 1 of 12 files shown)


igorDykhta:
Closing this stale PR - instead of a new adaptation https://github.com/keplergl/kepler.gl/pull/3408

PR #2678, opened 2024-10-07
[Feat] Dynamic map lib config

heshan0131:
* Support dynamic map lib


ilyabo (approved):
really cool!

ibgreen (approved):
Very nice. It would be really cool to support multiple basemaps, giving the user a choice, with little cost as they are dynamically loaded.

PR #2705, opened 2024-10-16
[feat] Updated time filter sync style

igorDykhta:
**Now:** 
Screen Shot 2024-10-16 at 5 00 13 PM

**Before:** 
Screen Shot 2024-10-16 at 5 04 05 PM

### DIFF
--- a/src/components/src/common/field-selector.tsx
+++ b/src/components/src/common/field-selector.tsx
@@ -4,7 +4,6 @@
 import classnames from 'classnames';
 import React, {Component, ComponentType} from 'react';
 import styled from 'styled-components';
-import classnames from 'classnames';
 import {createSelector} from 'reselect';
 
 import {Field} from '@kepler.gl/types';


src/components/src/side-panel/common/source-selector.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const StyledSourceContainer = styled.div`
+  background-color: #1c2233;
+`;
+
+SourceSelectorFactory.deps = [SourceDataSelectorContentFactory, FieldSelectorFactory];
+
+function SourceSelectorFactory(SourceDataSelectorContent, FieldSelector) {
+  const StyledFieldSelector = styled(FieldSelector)`
+    -webkit-border-bottom-right-radius: 4px;
```
ilyabo, line 18: Do we really need those prefixed styles? Won't `styled-components` add them?

ilyabo (approved):

PR #2724, opened 2024-10-31
[Bug] fix yaxis chat doesn't update

igorDykhta:
- fix yaxis chat doesn't update
- Sync time filter with layer animation

### DIFF
--- a/src/reducers/src/vis-state-updaters.ts
+++ b/src/reducers/src/vis-state-updaters.ts
@@ -42,6 +42,7 @@ import {
   isPlainObject,
   isObject,
   addNewLayersToSplitMap,
+  snapToMarks,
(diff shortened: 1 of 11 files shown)


src/reducers/src/vis-state-updaters.ts
```diff
@@ -1033,8 +1061,13 @@ export function setFilterUpdater(
   // pass only the dataset we need to update
   newState = updateAllLayerDomainData(newState, datasetIdsToFilter, newFilter);
 
+  // if (newFilter.syncedWithLayerTimeline) {
```
ilyabo, line 1064: maybe remove commented code?

ilyabo (approved):

PR #2725, opened 2024-10-31
[fix] Fixed synced filter domain and interval calculation

igorDykhta:
- Fixed synced filter domain and interval calculation


src/reducers/src/vis-state-updaters.ts
```diff
@@ -1061,10 +1066,6 @@ export function setFilterUpdater(
   // pass only the dataset we need to update
   newState = updateAllLayerDomainData(newState, datasetIdsToFilter, newFilter);
 
-  // if (newFilter.syncedWithLayerTimeline) {
```
ilyabo, line 1064: ah, I see, it's removed here

ilyabo (approved):

PR #2727, opened 2024-11-01
Update Code examples in API Reference Get Started page

indranildeveloper:
Update Code examples in API Reference Get Started page.


indranildeveloper:
@heshan0131 @ibgreen  if possible, I would love to have a review from you.

PR #2774, opened 2024-11-20
Updated imports for Kepler GL Reducers in docs

indranildeveloper:
Updated all the imports for Kepler GL Reducers to import from `@kepler.gl/reducers`


indranildeveloper:
@igorDykhta @ibgreen @heshan0131 Love to get a review from you.

PR #2775, opened 2024-11-20
Update the path to relative path in utils

indranildeveloper:
When built and used in an external web app, the `@kepler.gl/utils` was giving error because the application config was imported from `../../utils/src/application-config`. After building the `src` folder is missing so updated the imports to use the relative path to get rid of the error.

* Before in the build code:

Screenshot 2024-11-20 231658

* After in the build code:

Screenshot 2024-11-20 231841


indranildeveloper:
@igorDykhta @heshan0131 @ibgreen love to get a review.

PR #2777, opened 2024-11-20
[Feat] AI Assistant [2]

lixun910:
- Add geojson layer using field with quantile/quantize/ordinal/custom color scale [and first dataset]
- Add point layer
- Add arc layer
- Add heatmap
- Add h3 layer
- Filter field xxx on first dataset


lixun910:
keplergl-ai-2

lixun910:
Sometimes, the capability of the LLM model can leave you amazed by how intuitively it understands your queries and delivers unexpected yet accurate results. For instance, a model designed to respond to prompts like “Filter the data of price_pp using range [20, 50]” was able to interpret and act on a more abstract query “How many areas are between the values 14 and 41.05,” get the same outcome : ) can't wait to see real user experiences

Screenshot 2024-11-21 at 5 06 16 PM

Screenshot 2024-11-21 at 5 04 35 PM

lixun910:
😮  "create geojson layer with custom colors like Van Gogh’s starry night"

Screenshot 2024-11-26 at 1 57 11 PM

src/ai-assistant/src/tools/layer-utils.ts
```diff
@@ -0,0 +1,180 @@
+import {ColorMap, ColorRange, ColorLegends, COLOR_RANGES} from '@kepler.gl/constants';
```
ibgreen, line 1: Can these files be more thoughtfully named? This isn't just a another dump of utils is it? These are exposed function to let the assistant create layers?

ibgreen (approved):

PR #2798, opened 2024-11-28
[feat] duckdb plugin

ilyabo:


lixun910:
@ilyabo Looks super cool! Please please add some actions for querying the duckdb table, and I plan to add ai assistant to generate sql query from user prompts :)

PR #2819, opened 2024-12-07
[Feat] AI Assistant Query

lixun910:
- Allow user to query the variable using duckdb, which will be integrated in https://github.com/keplergl/kepler.gl/pull/2798

Screenshot 2024-12-07 at 1 14 48 AM


lixun910:
kepler-query

ilyabo:
awesome work, @lixun910!

I tried it with our ongoing work on the DuckDB SQL panel. I had to copy-paste the queries the assistant generated. Also, it didn't seem to have the loaded table names in the context, so I had to manually change that. I think we could try to improve that. I hope that we'll be able to move the plugin into the OS kepler soon, but we need to get this confirmed by the management at FSQ.
image

ilyabo:
@lixun910 ah, I just saw [your comment](https://github.com/keplergl/kepler.gl/pull/2798#issuecomment-2523991068) now. Yes, we should try to add actions for querying.

lixun910:
Thanks, @ilyabo! Got it. I hope your PR will be landing soon :) 
Can you approve this pr, so we can try and improve it?

ibgreen (approved):

examples/demo-app/esbuild.config.mjs
```diff
@@ -70,7 +68,8 @@ const config = {
   outfile: 'dist/bundle.js',
   bundle: true,
   define: {
-    NODE_ENV
+    'process.env.NODE_ENV': NODE_ENV,
+    'process.env': JSON.stringify(process.env)
```
lixun910, line 72: @igorDykhta Do we still need `process.env` ? The changes causes error when `yarn start:local` demo-app on my side:
Uncaught ReferenceError: process is not defined
    at default-settings.js:48:22
    at main.js:30:21
store.js:23 Uncaught ReferenceError: NODE_ENV is not defined
    at store.js:23:1
    at main.js:30:21

lixun910: Thanks, @igorDykhta! I just tested on master with clean build, and it shows the error, which refers to default-settings.js that `process.env` is still used: https://github.com/keplergl/kepler.gl/blob/b4c9e0ec4438195c0e0c68cd6172bc17b73b755a/examples/demo-app/src/constants/default-settings.js#L46-L55

lixun910: I see that we are using dotenv to handle the environment variables now. 
I need to create a .env from the .env.template to avoid the warning.
Should we add it to the README.md so that everyone knows the upcoming change? @igorDykhta

PR #2827, opened 2024-12-08
[chore] ts fixes

igorDykhta:
- ts fixes


src/utils/src/utils.ts
```diff
@@ -218,7 +218,7 @@ export function filterObjectByPredicate(obj, predicate) {
   );
 }
 
-export function isFunction(func): boolean {
+export function isFunction(func: any): boolean {
```
ilyabo, line 221: `func: unknown`  should also work here and would be more type-safe

ilyabo (approved):

PR #2867, opened 2024-12-27
[Fix] add_data() function not working in keplergl-jupyter

lixun910:
- Fix add_data() function not working in keplergl-jupyter v0.3.4
- Release v0.3.5


lixun910:
Thank you, @igorDykhta ! The new NPM_AUTH_TOKEN works when publishing keplergl-jupyter packages .

PR #2876, opened 2024-12-28
[fix] hide ColorBreaksDisplay when allBins is empty

igorDykhta:
- hide ColorBreaksDisplay when allBins is empty


src/components/src/side-panel/layer-panel/color-breaks-panel.tsx
```diff
@@ -194,7 +194,7 @@ function ColorBreaksPanelFactory(
               onApply={onApply}
               onCancel={onCilckCancel}
             />
-          ) : currentBreaks ? (
+          ) : currentBreaks && allBins.length > 1 ? (
```
ilyabo, line 197: Is allBins always defined?

ilyabo (approved):

PR #2565, opened 2024-04-20
[Feat] Kepler-Jupyter 0.3.4 with kepler v3

lixun910:
Summary:

- Update Kepler-Jupyter using kepler.gl 3.0.0, and upgrade keplergl python package to version 0.3.4-alpha0.
  - Get ready to publish keplergl 0.3.4 python package to PyPi and Conda
- Add new parameter: `use_arrow=True` (default is False) to allow load and render data faster using GeoArrow:

e.g.

```python
KeplerGl(data={'data_1': df}, config=config, use_arrow=True)
```

or
```python
map_1.add_data(df, 'data_1', use_arrow=True)`
```

Task:

- [x] build js for jupyter with kepler.gl v3
- [x] create python package
- [x] update dependencies
- [x] update documentation
- [x] Performance: load DataFrame or GeoPands DataFrame as Arrow table in the format of base64string to kepler-jupyter

Screenshot 2024-04-20 at 10 03 55 PM

PR #2901, opened 2025-01-09
[Fix] Problems to add data to maps through python code in kepler-jupyter

lixun910:
Revert the changes (which is probably added accidentally) in add_data() function from https://github.com/keplergl/kepler.gl/pull/2565/files#diff-9ad9c902c565677c69bdfc9666ed0fbfd12155854bd4339663e09bd475366ca1L143

Update version to 0.3.6

### DIFF
--- a/bindings/kepler.gl-jupyter/js/package.json
+++ b/bindings/kepler.gl-jupyter/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "keplergl-jupyter",
-  "version": "0.3.5",
+  "version": "0.3.6",
   "description": "This is a simple jupyter widget for kepler.gl, an advanced geo-spatial visualization tool, to render large-scale interactive maps.",
   "author": "Shan He",
   "license": "MIT",
--- a/bindings/kepler.gl-jupyter/keplergl/_version.py
+++ b/bindings/kepler.gl-jupyter/keplergl/_version.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
 # Copyright contributors to the kepler.gl project
 
-version_info = (0, 3, 5, 'final', 0)
+version_info = (0, 3, 6, 'final', 0)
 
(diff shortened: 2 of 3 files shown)


rbavery:
thanks for the fix @lixun910. I've tested this PR, geodataframes can be added but now two geodataframes cannot be added to the same map and be visualized. Tested https://github.com/keplergl/kepler.gl/blob/master/bindings/kepler.gl-jupyter/notebooks/GeoDataFrame.ipynb and only one of points_gdf or zipcode_gdf can be added at a time.

lixun910:
Thanks @rbavery !  In my test, I see only one layer been created from the first dataset. Is it the same on your side?

rbavery:
hmm you're right actually after a fresh jupyterlab restart I see the issue again on the first run. I'm trying to diagnose what causes me to successfully plot both layers and why on the first run after jupyterlab start this doesn't occur.

rbavery:
@lixun910 I found the culprit but I'm not sure what is the root cause

There's a cell in that notebook that shows the map before any layers are added. If we show the map before layers are added, only the first added layer is shown even if two datas are added

```
import keplergl
w1 = keplergl.KeplerGl(height=500)
w1
```

If instead we only make the map but don't show it before adding a layer, then we can show both layers and after adding the second layer both layers are plotted as expected

```
import keplergl
w1 = keplergl.KeplerGl(height=500)
w1
```

so I think somehow the first map instance `w1` is interfering with the state of subsequent maps.

rbavery:
the issue might be around here, `data_to_add` is a dict with keys for each layer. If the empty `w1` map is shown then only one layer is included in `data_to_add`

lixun910:
Ah, I see. I think it is by design: one needs to render the map with the new dataset in a new cell after calling `add_data()`, and the map in previous cell can keep the original state.

Screenshot 2025-01-09 at 7 47 01 PM

PR #2904, opened 2025-01-10
[Chore] esbuild website to speedup netlify build

lixun910:


lixun910:
website works fine with esbuild

Screenshot 2025-01-22 at 10 46 36 AM

PR #2919, opened 2025-01-18
[fix] fixes for style editor preview and error

igorDykhta:
- fix style preview 
- take into account basemap library switching in add-map-style-modal
- trigger new Map components with url, token or library change in order to clear cache of loaded styles. Otherwise 'style.load' events aren't triggered when input changes from correct > wrong > correct config.

Screenshot 2025-01-19 at 1 26 10 AM

Screenshot 2025-01-19 at 1 27 04 AM

Screenshot 2025-01-19 at 1 27 09 AM

### DIFF
--- a/src/reducers/src/map-style-updaters.ts
+++ b/src/reducers/src/map-style-updaters.ts
@@ -697,7 +697,6 @@ export const inputMapStyleUpdater = (
   // differentiate between either a url to hosted style json that needs an icon url,
   // or an icon already available client-side as a data uri
   const isUpdatedIconDataUri = updated.icon?.startsWith('data:image');
-  const isValid = true; // Boolean(updated.uploadedFile);
   const isMapboxStyleUrl =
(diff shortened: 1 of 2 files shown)


src/reducers/src/map-style-updaters.ts
```diff
@@ -697,7 +697,7 @@ export const inputMapStyleUpdater = (
   // differentiate between either a url to hosted style json that needs an icon url,
   // or an icon already available client-side as a data uri
   const isUpdatedIconDataUri = updated.icon?.startsWith('data:image');
-  const isValid = Boolean(updated.uploadedFile);
+  const isValid = true; // Boolean(updated.uploadedFile);
```
ilyabo, line 700: is this change intended?

ilyabo: if so, maybe remove `isValid` altogether?

ilyabo (approved):

PR #2940, opened 2025-01-27
(WIP) [feat] emulate geoarrow.wkb and geoarrow.linestring support in trip layer

igorDykhta:
**There is a blocker in DuckDB as 4th coord with time is dropped in both st_asWKB and st_asGeojson**

WIP

- [x] emulate geoarrow.wkb support in trip layer
- [x] emulate geoarrow.linestring support in trip layer
- [ ] fix: DuckDb drops 4th dimension with timestamps (!!!)
- [ ] figure out a way to prepare attributes for deck.gl's Path layer (sublayer of TripsLayer).


igorDykhta:
Still not supported and with conflicts. Closing.

PR #2946, opened 2025-01-28
[Chore] Add DeepSeek in Ai Assistant

lixun910:
- Add options to select DeepSeek models
- Add timer to monitor initial model connection


ilyabo (approved):
cool! 👍

ilyabo:
@lixun910 A suggestion (for a separate PR): I think it would be nice to save the api keys users enter in the assistant to local storage so that they don't have to enter them every time the open the app. What do you think?

lixun910:
The DeepSeek is down right now, and I can't test it. I will merge it now and fix any potential issue later.

PR #2952, opened 2025-01-28
[feat] DuckDb plugin: drag and drop file directly as table

igorDykhta:
- [x] can drag & drop a file into sql schema window
- [x] a file bypasses regular Kepler.gl processing and is ingested into DuckDb with registerFileHandle
- [x] create a table from registered file (csv, json, geojson, parquet)
- [x] create a table from registered arrow file
- [x] some indication that it's a drop zone
- [x] ingestion progress indicator
- [x] the new table can be queried and used to send data to kepler.gl
- [x] ingestion error messages

Screenshot 2025-02-28 at 4 58 12 AM

Screenshot 2025-01-29 at 1 31 53 AM

### DIFF
--- a/src/duckdb/src/components/schema-panel.tsx
+++ b/src/duckdb/src/components/schema-panel.tsx
@@ -79,7 +79,7 @@ function getSchemaSuggestion(result: {key: string; children: {key: string}[]}[])
 }
 
 type SchemaPanelProps = {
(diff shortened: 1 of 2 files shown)


ilyabo (approved):
nice!  👍 
we should add some indication that it's a drop zone, e.g. like in GIP lite:
image

PR #2953, opened 2025-01-29
[fix] arrow tables - save timestamps as iso date string

igorDykhta:
- convert numerical timestamps from arrow tables into iso date strings before saving.

### DIFF
--- a/src/schemas/src/dataset-schema.ts
+++ b/src/schemas/src/dataset-schema.ts
@@ -3,6 +3,7 @@
 
 import pick from 'lodash.pick';
 import {console as globalConsole} from 'global/window';
+import {Type as ArrowTypes} from 'apache-arrow';
 
 import {DATASET_FORMATS} from '@kepler.gl/constants';
 import {ProtoDataset, RGBColor, JsonObject} from '@kepler.gl/types';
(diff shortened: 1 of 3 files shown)


src/schemas/src/dataset-schema.ts
```diff
@@ -100,14 +101,40 @@ export const propertiesV1 = {
...
+const getAllDataForSaving = (dataContainer: DataContainerInterface): any[][] => {
+  const allData = dataContainer.flattenData();
+
+  if (dataContainer instanceof ArrowDataContainer) {
+    const numColumns = dataContainer.numColumns();
+
+    for (let columnIndex = 0; columnIndex < numColumns; ++columnIndex) {
+      const column = dataContainer.getColumn(columnIndex);
+      // TODO why arrow.Type.Timestamp is undefined?
+      if (column.type.typeId === 10) {
```
lixun910, line 119: Can we use ‘ arrow.Type.Timestamp’ enum instead of 10 here? I remember timestamp is not handled well in our initial implementation since it has timestamp, timestamp second, timestamp nanosecond etc.

lixun910 (approved):

PR #2955, opened 2025-01-29
[feat] banner with extra release info

igorDykhta:
- banner with extra release info
- once minified, the user won't see the banner again during next sessions

Screenshot 2025-01-29 at 3 23 51 PM


ilyabo (approved):
love it 👍

Issue #2967, opened 2025-01-30
[Bug][Jupyter Widget] non-serializable object is not handled in 0.3.6 Kepler release


rbavery:
**Describe the bug**
A pandas Timestamp object causes a big red warning when making a Kepler map. This used to work silently and I think the Timestamp object was either converted to a json serializable type or was ignored.

```python
/opt/conda/envs/wherobots/lib/python3.11/site-packages/jupyter_client/session.py:719: UserWarning: Message serialization failed with:
Out of range float values are not JSON compliant
Supporting this message is deprecated in jupyter-client 7, please make sure your message is JSON-compliant
  content = self.pack(content)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File /opt/conda/envs/wherobots/lib/python3.11/site-packages/IPython/core/formatters.py:406, in BaseFormatter.__call__(self, obj)
    404     method = get_real_method(obj, self.print_method)
    405     if method is not None:
--> 406         return method()
    407     return None
    408 else:

File /opt/conda/envs/wherobots/lib/python3.11/site-packages/keplergl/keplergl.py:293, in KeplerGl._repr_html_(self, data, config, read_only, center_map)
    288 config_to_add = self.config if config is None else config
    290 # for key in data_to_add:
    291 #     print(type(data_to_add[key]))
--> 293 keplergl_data = json.dumps({"config": config_to_add, "data": data_to_add, "options": {
    294                            "readOnly": read_only, "centerMap": center_map}})
    296 cmd = f"window.__keplerglDataConfig = {keplergl_data};"
    297 frame_txt = keplergl_html[:k] + "<body><script>" + \
    298     cmd + "</script>" + keplergl_html[k+6:]

File /opt/conda/envs/wherobots/lib/python3.11/json/__init__.py:231, in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    226 # cached encoder
    227 if (not skipkeys and ensure_ascii and
    228     check_circular and allow_nan and
    229     cls is None and indent is None and separators is None and
    230     default is None and not sort_keys and not kw):
--> 231     return _default_encoder.encode(obj)
    232 if cls is None:
    233     cls = JSONEncoder

File /opt/conda/envs/wherobots/lib/python3.11/json/encoder.py:200, in JSONEncoder.encode(self, o)
    196         return encode_basestring(o)
[...]

PR #2968, opened 2025-01-31
convert datetimes to str so they can convert to json

rbavery:
addresses https://github.com/keplergl/kepler.gl/issues/2967

I tested the MRE in the issue above locally and it works. Would probably be good to have handlign for other non-JSON serializable types as well.

I didn't find the root cause for why this wasn't an issue in earlier widget versions so I opted to make this type handling change in `_df_to_dict`. Let me know if you think there would be a better place to do this handling.

### DIFF
--- a/bindings/kepler.gl-jupyter/keplergl/keplergl.py
+++ b/bindings/kepler.gl-jupyter/keplergl/keplergl.py
@@ -49,7 +49,11 @@ def _df_to_dict(df):
     - dictionary: a dictionary variable that can be used in Kepler.gl
 
     '''
-    return df.to_dict('split')
+    df_copy = df.copy()
+    for col in df_copy.columns:
+        if pd.api.types.is_datetime64_any_dtype(df_copy[col]):
+            df_copy[col] = df_copy[col].astype(str)
+    return df_copy.to_dict('split')
 
 
 def _df_to_arrow(df: pd.DataFrame):
--- a/bindings/kepler.gl-jupyter/keplergl/keplergl.py
+++ b/bindings/kepler.gl-jupyter/keplergl/keplergl.py
@@ -50,9 +50,14 @@ def _df_to_dict(df):
 
     '''
     df_copy = df.copy()
+    # Convert all columns that aren't JSON serializable to strings
     for col in df_copy.columns:
-        if pd.api.types.is_datetime64_any_dtype(df_copy[col]):
+        try:
+            # just check the first item in the colum
+            json.dumps(df_copy[col].iloc[0] if len(df_copy) > 0 else None)
+        except (TypeError, OverflowError):
             df_copy[col] = df_copy[col].astype(str)
+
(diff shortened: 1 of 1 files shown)


lixun910:
Thanks @rbavery! 
Em, not sure what might cause this if this never happens before. The geopandas/pandas version has not been changed...
Can we have a more generic solution to avoid potential TypeError: Object of type Timestamp(or other types) is not JSON serializable, like this?
```python
    df_copy = df.copy()
    # Convert all columns that aren't JSON serializable to strings
    for col in df_copy.columns:
        try:
            # just check the first item in the colum
            json.dumps(df_copy[col].iloc[0] if len(df_copy) > 0 else None)
        except (TypeError, OverflowError):
            df_copy[col] = df_copy[col].astype(str)
    
    return df_copy.to_dict('split')
```
I can help to publish a new version to fix this issue.

lixun910 (approved):

rbavery:
Sure that seems better and thanks for the quick review.

rbavery:
@lixun910 sorry to correct my earlier statement, this issue has always been present in the official keplergl jupyter widget release. It was fixed in an internal fork we were using but then we switched to this official release.

It looks like we had handling for datetime specifically and do the conversion with strftime 

```
import datetime

class DateEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj,datetime.datetime):
            return obj.strftime("%Y-%m-%dT%H:%M:%S")
        else:
            return json.JSONEncoder.default(self,obj)

@widgets.register
class KeplerGl(widgets.DOMWidget):
    """An example widget."""
        data_to_add = data_to_json(self.data, None) if data == None else data_to_json(data, None)
        config_to_add = self.config if config == None else config

        keplergl_data = json.dumps({"config": config_to_add, "data": data_to_add, "options": {"readOnly": read_only, "centerMap": center_map}},cls=DateEncoder)

        cmd = """window.__keplerglDataConfig = {};""".format(keplergl_data)
        frame_txt = keplergl_html[:k] + "<body><script>" + cmd + "</script>" + keplergl_html[k+6:]
```

I'll check if strftime makes a difference over just calling `astype(str)` and if so add this handling for datetimes if that sounds good to you.

lixun910:
Got it. Yes, sounds good to me. We can land it as a fix and consider a more generic fix by adding or allowing passing more encoders to avoid the possible type error. Another way to fix it may be passing the raw arrow format from dataframe and handle the format and type in kepler.gl? We can work in other PRs. Thanks, @rbavery!

rbavery:
@lixun910 Agree! I checked and it doesn't make much diff to use `.astype(str)` on the column or `strftime`

```

df['Timestamp'].astype(str).iloc[0]
'2002-03-03 10:10:00'

vs

df['Timestamp'].iloc[0].strftime("%Y-%m-%dT%H:%M:%S")

'2002-03-03T10:10:00'
```


So I think this can be merged.

PR #2976, opened 2025-02-05
Rename ".env.template "  to ".env.template"

ironicbrew:
Current state causes ```git clone``` to throw on a windows computer:

```
invalid path '.env.template '
fatal: unable to checkout working tree
```

Removing the space from the ```.env.template``` file fixes the issue


ilyabo:
Thanks for the fix! @ironicbrew

PR #2986, opened 2025-02-07
Dockerize

fkauker-fsq:
Create a very simple docker container


fkauker-fsq:
You can run this by calling

```
docker build .
docker image ls
docker tag <hash> keplergl-latest
docker run -p 8080:8080 keplergl-latest
```

PR #2988, opened 2025-02-08
[fix] export geoarrow to CSV as geojson

igorDykhta:
- [x] geoarrow.polygon save / load as stringified geojson
- [x] geoarrow.linestring save / load as stringified geojson
- [x] geoarrow.point save / load as stringified geojson
- [x] hex wkb string save / load as stringified geojson

### DIFF
--- a/src/utils/src/data-utils.ts
+++ b/src/utils/src/data-utils.ts
@@ -23,7 +23,7 @@ import {isPlainObject} from './utils';
 
 export type FieldFormatter = (value: any, field?: Field) => string;
 
-// We need threat latitude differently otherwise marcator project view throws
+// We need threat latitude differently otherwise mercator project view throws
 // a projection matrix error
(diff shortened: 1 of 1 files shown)


src/utils/src/data-utils.ts
```diff
@@ -18,7 +21,7 @@ import {Field, Millisecond} from '@kepler.gl/types';
 import {snapToMarks} from './plot';
 import {isPlainObject} from './utils';
 
-export type FieldFormatter = (value: any) => string;
+export type FieldFormatter = (value: any, field?: Field) => string;
 
 // We need threat latitude differently otherwise marcator project view throws
```
ilyabo, line 26:
```suggestion
// We need threat latitude differently otherwise mercator project view throws
```

ilyabo (approved):

PR #3038, opened 2025-04-01
[Bug] on dataset change: this._rows[e] is undefined

Ritesh9876:
Steps to reproduce:
1.Select two datasets where at least one column is common, but the number of rows differs.

2. In a layer, remove some columns.

3. Change the dataset for the layer.

4. Check the console for errors.

Root Cause:
Kepler.gl's colorField, sizeField, and similar properties continue referencing fields from the previously used dataset. Since the datasets have different row counts, these fields attempt to access indices beyond the length of the previous dataset, causing the error when index > firstDataset.rows.length.

[sameple_2.csv](https://github.com/user-attachments/files/19554531/sameple_2.csv)
[sample_1.csv](https://github.com/user-attachments/files/19554533/sample_1.csv)
[dataset_changed_undefined.webm](https://github.com/user-attachments/assets/a22b5676-b834-44a7-816b-28e84edd4dc0)

### DIFF
--- a/src/reducers/src/vis-state-updaters.ts
+++ b/src/reducers/src/vis-state-updaters.ts
@@ -781,6 +781,44 @@ function validateExistingLayerWithData(dataset, layerClasses, layer, schema) {
     : null;
 }
 
+const validateAndReplaceFields = (state, layer, dataId) => {
+  const newSizeField = state.datasets[dataId].fields.find(
+    field => field.name === layer.config.sizeField?.name
+  );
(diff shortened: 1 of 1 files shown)


Ritesh9876:
Hii @igorDykhta @heshan0131 
Just checking in, would you have a moment to take a quick look at this PR?
If it's not something you'd like to move forward with, feel free to let me know and I can close it. Thanks!

PR #3051, opened 2025-04-20
[Bug] Draw on Map and Show Locale dropdowns active at the same time (right map controls)

indranildeveloper:
@igorDykhta @heshan0131 
In this PR, when we open the map draw dropdown, then the locale dropdown closes if it was opened, and vice versa.


indranildeveloper:
I will make a new PR with the updated code.

PR #3065, opened 2025-04-30
[Chore] drop lodash per-method packages in favor of the main lodash

### LINKED ISSUE #3046: [Bug] high severity vulnerabilities dependency lodash.pick
**Describe the bug**
Lodash per method packages are not maintained anymore, and their usage is [discouraged by the team](https://lodash.com/per-method-packages).
Kepler packages have multiple lodash per method packages as dependencies.  
`lodash.pick` specifically causes `npm audit` to raise warnings about multiple high severity vulnerabilities.

Workaround suggested here (https://github.com/lodash/lodash/issues/5809#issuecomment-1911616800) causes the basic template example from docs to fail with `An error in deck.gl: deck.gl: assertion failed..`

**To Reproduce**
- Just run `npm install --save kepler.gl @kepler.gl/components @kepler.gl/reducers` on an empty folder to see all the vulnerabilities
- To see the fail override workaround error on map:
  1. Start a new vite project with react redux: `npx degit reduxjs/redux-templates/packages/vite-template-redux my-app`
  2. Add kepler `npm install --save kepler.gl @kepler.gl/components @kepler.gl/reducers`
  3. Add override in `package.json`:
      ```
       "overrides": {
           "lodash.pick": "https://github.com/lodash/lodash/archive/refs/tags/4.17.21.tar.gz"
       }
     ```
  4. Follow https://docs.kepler.gl/docs/api-reference/get-started to render kepler on app
  5. `npm run dev`, if not presented with the error, try using the dual map button

**Expected behavior**
Should not depend on vulnerable package `lodash.pick`
Example on how to move away from that https://github.com/GeekyAnts/NativeBase/pull/5799

Does not concern device

**Smartphone (please complete the following information):**
Does not concern device

**Additional context**
The 3 kepler packages have in total 37 vulnerabilities (5 moderate, 32 high):
- `lodash.pick` is a direct dependency
- react-vis introduces `hoek`, `d3-color` (#1574)
- typedoc introduces `marked`
- react-palm introduces `node-fetch`
- @loaders.gl/polyfills introduces `request` and `tough-cookie`

Overriding these removes the warnings, but of course many are breaking changes, I have not tested the impact.

leverglowh:
- fix #3046: no one else in the dependency tree depend on `lodash.pick`
- import packages with `lodash/method` as suggested by docs to reduce weight

Lodash documentation on why we should drop them: https://lodash.com/per-method-packages


leverglowh:
> @leverglowh you have to update lodash versions in examples/demo-app package.json and yarn.lock to pass Netlify checks

I'm sorry, I didn't find any `lodash` dependency in demo-app's package.json..
It does depend on Kepler sub modules such as components, actions etc that when resolved still have old `lodash.method` dependencies.

How can I fix it?

leverglowh:
Ah, you're right, I missed it!
Submodules are resolved on the main lock file, but website being independent has its own!
It should be fixed now

PR #3083, opened 2025-05-07
[chore] keplergl-jupyter: reduce list of python build requirements

jorisvandenbossche:
I noticed that installing this package installs a bunch of packages when pip creates the isolated build environment (which happens since keplergl-jupyter only provides a source distribution, so it will "build" a wheel when installing).

Right now also packages like geopandas and pyarrow get installed in that isolated build environment, while those are not needed for building the wheel (only as runtime dependency in the environment in which you are installing keplergl into).

I _think_ only `jupyter-packaging` (and its dependencies like setuptools) is needed for running `setup.py` (also following https://github.com/jupyter/jupyter-packaging?tab=readme-ov-file#as-a-build-requirement). 
The actual runtime dependencies which includes the packages like geopandas and pyarrow, are still listed in ` is still the setup.py  (`install_requires`), so this change should not impact the end result of `pip install keplergl-jupyter`


igorDykhta:
`bindings/kepler.gl-jupyter` is deprecated and has been replaced by `bindings/python`.

PR #3106, opened 2025-05-14
[Chore] Better handle add data to map error and loading indicator

heshan0131:


heshan0131:
> Also, the main problem in Desktop App was FIT_BOUNDS **map state** action called as action task from **vis state updaters** from `createNewDatasetSuccessUpdater > applyMergersUpdater > postMergeUpdater`

Yes, here we still calls Action tasks on fit bounds. Just one less action task if it can be avoided

PR #3117, opened 2025-05-22
[Feat] Add OSM road tool to support point analysis on road networks

lixun910:
- add osm road tool, which allow AI to get OSM roads from current map view or the boundary of a dataset
- update instruction to guide point analysis on roads e.g. `Are the robbery points clustering on streets?`

The following screenshot shows an example of using OSM road tool with other tools to apply point analysis on road networks:
- Get the current map boundary to focus the analysis.
- Query and download the road network within this boundary.
- Perform a Local Moran’s I (LISA) analysis on the count of robberies near roads to detect clustering.

Screenshot 2025-05-21 at 10 42 08 PM


igorDykhta (approved):
:rocket:

PR #3118, opened 2025-05-22
[perf] disable strokes by default for polygons in geojson layer

igorDykhta:
Disable strokes by default for polygons in GeoJSON layer

Motivation:
When working with large datasets, the performance of the GeoJSON layer can degrade significantly.

- The additional stroke geometry increases memory usage.
- Stroke rendering uses screen-space sizing, which means that when zoomed out, a single screen fragment might require rendering thousands of stroke geometries, severely impacting performance (Overdraw).

- [x] save / export not affected

### DIFF
--- a/src/layers/src/geojson-layer/geojson-layer.ts
+++ b/src/layers/src/geojson-layer/geojson-layer.ts
@@ -625,7 +625,7 @@ export default class GeoJsonLayer extends Layer {
       // set both fill and stroke to true
       return this.updateLayerVisConfig({
         filled: true,
-        stroked: true,
+        stroked: false,
         strokeColor: colorMaker.next().value
       });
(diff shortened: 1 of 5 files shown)


lixun910:
Thanks, @igorDykhta one comment: Is it possible to add a condition to disable the stroke, for example, when the number of geometries exceeds 100K? When loading a dataset without default color schema, I think we still need the stroke color to show up the polygons?: Screenshot 2025-05-22 at 3 31 50 PM

igorDykhta:
> Thanks, @igorDykhta one comment: Is it possible to add a condition to disable the stroke, for example, when the number of geometries exceeds 100K? When loading a dataset without default color schema, I think we still need the stroke color to show up the polygons?

Good catch!

lixun910 (approved):
Thanks!

PR #3136, opened 2025-06-10
[wip] remove redux router use react-router-dom in demo app

heshan0131:


igorDykhta:
react-router was replaced with react-router-dom in React 19 upgrade https://github.com/keplergl/kepler.gl/pull/3514

Issue #3153, opened 2025-07-08
[Bug] Kepler DuckDB Preview - spaces in column name


igorDykhta:
Just tried to add a CSV with spaces in the col name to the kepler preview app and it crashed as DuckDB table load is run by default?

igorDykhta:
This is fixed in 3.1.9

PR #3156, opened 2025-07-09
[fix] DuckDB mode: space in column name breaks file import (#3153)

igorDykhta:
#3153

- fix for space in column name
- fix for complex table names that are already quoted
- tests for duckDB utils

### DIFF
--- a/src/duckdb/src/table/duckdb-table-utils.ts
+++ b/src/duckdb/src/table/duckdb-table-utils.ts
@@ -108,13 +108,13 @@ export function castDuckDBTypesForKepler(
 ): string {
   const modifiedColumns = columns.map(column => {
     const {name, type} = column;
-    const quotedName = quoteColumnName(name);
+    const quotedColumnName = quoteColumnName(name);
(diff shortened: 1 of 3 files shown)


src/duckdb/src/table/duckdb-table-utils.ts
```diff
@@ -60,6 +61,46 @@ export function getDuckDBColumnTypesMap(columns: DuckDBColumnDesc[]) {
...
+  // This handles cases like "table", "table with spaces", and "table.with.dots"
+  if (columnName.startsWith('"') && columnName.endsWith('"')) {
+    return columnName;
+  }
+
+  // If it contains dots and quotes, it might be a fully qualified name
+  if (columnName.includes('.') && columnName.includes('"')) {
+    // Simple heuristic: if it looks like "part1"."part2" or "part1"."part2"."part3"
+    // Split on dots and check if each part is quoted
+    const parts = columnName.split('.');
```
ilyabo: This will break if there's a . within quotes. Maybe just `return columnName;` in this case too? It's unlikely that anybody provides a partially quoted qualified name

PR #3178, opened 2025-08-13
[improvement] optimize speed of getCategoricalColorMap

igorDykhta:
- optimize speed of getCategoricalColorMap


lixun910 (approved):
LGTM. 👍  One small thing:

do we need to filter the null/undefined empty values? They are very common in user data (many empty values).

Issue #3174, opened 2025-08-06
Horizontal Scrollbars on Legend on editing fields and also on exporting image, mismatch of position of legend on exported image


GuntuAshok:
**Describe the bug**
Scrollbars on Legend on editing fields

**To Reproduce**
1. Go to https://kepler.gl/demo/county_unemployment
2. Click on Legend
3. Edit any legend field
4. Horizontal scroll will appear
5. Export image and the scrollbar color changes and also position of legend changes.

**Expected behavior**
Horizontal scroll is not a problem but there should be resize option horizontally also to avoid scrolls otherwise the exported image will get unnecessary scroll bars and it can't be avoided since there is no horizontal resize. Also the position and size of legend is also changed on exported image

**Screenshots**

<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/070a8957-224f-44c2-afad-6cb11874fc57" />

<img width="1545" height="854" alt="Image" src="https://github.com/user-attachments/assets/801425a3-9d7b-4220-8b8b-07b3feef4ad6" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Edge
 - Version Keplergl 3.1.9

indranildeveloper:
I will look into it.

igorDykhta:
@GuntuAshok could you check the latest kepler.gl/demo ? The legend issues should be fixed there

PR #3199, opened 2025-08-29
[fix] fixes for legend

igorDykhta:
for #3174 

- move _revert_ icon left to prevent bottom scrollbar from appearing / cutout by the right side of the legend
- ellipsis for long values
- fix for legend positioning during image export
- hide any legend scrollbars during image export


lixun910 (approved):
👍

PR #3207, opened 2025-09-08
[feat] custom zoom offset for raster tile layer

igorDykhta:
- custom zoom offset for raster tile layer


igorDykhta:
This is already in place

PR #3227, opened 2025-10-22
feat: extend bigInt casting to support UBIGINT, HUGEINT, UHUGEINT

lixun910:
- Add support for casting UBIGINT, HUGEINT, and UHUGEINT to DOUBLE
- Prevents BigInt issues in JavaScript when dealing with large integer types

From the DuckDB documentation:

- HUGEINT is a signed 16-byte (128-bit) integer type. 
- UHUGEINT is an unsigned 16-byte (128-bit) integer type.
- UBIGINT is an unsigned 8-byte (64-bit) integer type.


igorDykhta (approved):

lixun910:
Thanks @igorDykhta !

PR #3253, opened 2025-12-03
fix(jupyter): Fix initial viewport sizing and React 18 migration

### LINKED ISSUE #3252: keplergl-jupyter: Map doesn't fill viewport on initial render (requires resize event)
## Description

When using `keplergl` Python package (v0.3.7) with `_repr_html_()` to serve the map in a web page, the map renders small on initial page load and only fills the full viewport after manually resizing the browser window.

## Steps to Reproduce

1. Install keplergl: `pip install keplergl==0.3.7`
2. Generate HTML and serve it:
```python
from keplergl import KeplerGl
kepler_map = KeplerGl()
html = kepler_map._repr_html_()
# Serve html via Flask/FastAPI/etc
```
3. Open the page in browser
4. Map appears small (not filling viewport)
5. Resize browser window → map expands to fill viewport

## Expected Behavior

Map should fill the full viewport on initial page load without requiring a resize event.

## Root Cause Analysis

The bundled React component in `keplergl.html` creates a div with inline styles `width: 100vw; height: 100vh`, but the Kepler.gl component doesn't correctly calculate its dimensions on initial render. It relies on `window.resize` events to recalculate dimensions.

## Workaround

Inject a script to trigger a resize event after page load:

```python
resize_script = """<script>
    window.addEventListener('load', function() {
        setTimeout(function() {
            window.dispatchEvent(new Event('resize'));
        }, 100);
    });
</script>"""
html = html.replace("</body></html>", f"{resize_script}</body></html>", 1)
```

**Note**: Don't use `</body>` alone as replacement target - it appears inside minified JS string literals.

## Environment

- keplergl version: 0.3.7
- Browser: Firefox/Chrome (both affected)
- OS: macOS

## Suggested Fix

The `bindings/kepler.gl-jupyter/js/lib/keplergl/main.js` could dispatch a resize event after initial render, or the component could use ResizeObserver for more reliable dimension calculation.

yharby:
## Summary

This PR fixes two related issues in the kepler.gl-jupyter HTML export:

1. **Initial viewport sizing** - Map doesn't fill the viewport on initial page load, requiring a browser resize to trigger correct sizing
2. **React 18 compatibility** - Migrates from deprecated `ReactDOM.render` to the modern `createRoot` API

### Changes

**app.js:**
- Call `handleResize()` on component mount in the `useEffect` hook to set correct initial dimensions

**root.js:**
- Migrate from `ReactDOM.render` to React 18's `createRoot` API
- Add `DataLoader` component to handle data loading after Redux Provider mounts
- This ensures Provider subscriptions are active before dispatching `addDataToMap` actions

**main.js:**
- Remove data loading logic (now handled by DataLoader component in React lifecycle)
- Simplify to only handle DOM setup and trigger render

### Root Cause

The original code had two issues:

1. `handleResize()` was only called on the resize event, never on initial mount, so the map container started with 0x0 dimensions until a resize occurred

2. React 18's `createRoot().render()` is asynchronous, unlike the synchronous `ReactDOM.render()`. This caused a race condition where data was dispatched to Redux before the Provider's subscriptions were active, resulting in no layers being displayed

### Testing

- Tested locally with HTML export - map now fills viewport immediately on load
- Data/config loads correctly with layers displayed
- No console errors

Fixes #3252

### DIFF
--- a/bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js
+++ b/bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // Copyright contributors to the kepler.gl project
 
-import React, {useEffect, useState, useRef} from 'react';
+import {useEffect, useState, useRef} from 'react';
 import styled from 'styled-components';
 const ReactHelmet = require('react-helmet');
 const Helmet = ReactHelmet ? ReactHelmet.Helmet : null;
@@ -58,15 +58,12 @@ function App() {
 
     const width = rootElm.current.offsetWidth;
     const height = rootElm.current.offsetHeight;
-    const dimensionToSet = {
-      ...(width && width !== windowDimension.width ? {width} : {}),
-      ...(height && height !== windowDimension.height ? {height} : {})
-    };
-
-    setDimension(dimensionToSet);
+    if (width && height) {
+      setDimension({width, height});
+    }
   };
 
-  // in Jupyter Lab,  parent component has transition when window resize.
+  // in Jupyter Lab, parent component has transition when window resize.
   // need to delay call to get the final parent width,
   const resizeDelay = () => window.setTimeout(handleResize, 500);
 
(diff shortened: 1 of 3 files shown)


igorDykhta:

PR #3255, opened 2025-12-09
fix: Avoid Monaco AMD bundle when importing `@kepler.gl/duckdb`

ilyabo:
#### Problem
The root entrypoint of `@kepler.gl/duckdb` re-exported React components (e.g., `SqlPanel`) that pull the AMD build of `monaco-editor`. As a result, any import of `@kepler.gl/duckdb` (even for table/helpers only) dragged in Monaco AMD, which expects a global AMD loader (`define`/`require`). In ESM/Vite setups this triggers runtime errors like `ReferenceError: define is not defined` and clashes with apps that configure Monaco via the ESM build.

#### What this PR changes
- Decouples helpers from UI components:
  - Root entry now exposes only the plugin, tables, table utilities, and processors; it no longer imports UI/Monaco.
  - React UI components (SqlPanel, SchemaPanel, PreviewDataPanel, MonacoEditor, etc.) are exposed from a separate subpath: `@kepler.gl/duckdb/components`.
  - Table utilities are also available via `@kepler.gl/duckdb/table` for direct, UI-free imports.
- Updated `package.json` exports map to publish `.` (root), `./components`, and `./table` (with types), keeping helper imports free of Monaco AMD.

#### Rationale
- Lets ESM-first bundlers (e.g., Vite) consume `@kepler.gl/duckdb` without hitting AMD `define` errors.
- Prevents bundling the AMD Monaco build when only data-layer helpers are needed and avoids mixing Monaco configs for apps that already use the ESM build.
- Reduces accidental coupling between data utilities and UI.

#### Backwards compatibility
- Helper imports keep working and no longer pull Monaco:
  ```ts
  import {
    getDuckDBColumnTypes,
    getDuckDBColumnTypesMap,
    castDuckDBTypesForKepler
  } from '@kepler.gl/duckdb';
  ```
- UI consumers should import from the new subpath:
  ```ts
  import {SqlPanel} from '@kepler.gl/duckdb/components';
  ```
- Table-only consumers can target:
  ```ts
  import {getDuckDBColumnTypes} from '@kepler.gl/duckdb/table';
  ```

### DIFF
--- a/src/duckdb/package.json
+++ b/src/duckdb/package.json
@@ -18,9 +18,9 @@
       "require": "./dist/components/index.js"
     },
     "./table": {
-      "types": "./dist/table/duckdb-table-utils.d.ts",
-      "import": "./dist/table/duckdb-table-utils.js",
-      "require": "./dist/table/duckdb-table-utils.js"
+      "types": "./dist/table/index.d.ts",
+      "import": "./dist/table/index.js",
+      "require": "./dist/table/index.js"
     }
   },
   "keywords": [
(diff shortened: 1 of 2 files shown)


ilyabo:
`yarn start` runs now:

image

PR #3300, opened 2026-02-06
rollback change, and truncate tooltip

bdjulbic:
Rollback of #3298 
The previous fix for hiding geometry fields in tooltips wasn't covering all use cases.

Problem: Geometry fields loaded as strings via CSV (displayed as vector tilesets) show excessively long WKT values in map tooltips.
image

Solution: Apply truncation to tooltip values exceeding 60 characters, shortening them to 30 characters with ellipsis. This handles all long string values uniformly without needing to detect specific field types.
Screenshot 2026-02-06 at 14 52 58

### DIFF
--- a/src/components/src/map/layer-hover-info.tsx
+++ b/src/components/src/map/layer-hover-info.tsx
@@ -67,7 +67,7 @@ interface RowProps {
   url?: string;
 }
 
-const TOOLTIP_VALUE_MAX_LENGTH = 1024;
+const TOOLTIP_VALUE_MAX_LENGTH = 256;
 
 const Row: React.FC<RowProps> = ({name, value, deltaValue, url}) => {
   // Set 'url' to 'value' if it looks like a url


src/common-utils/src/data-type.ts
```diff
@@ -11,27 +11,6 @@ import {h3IsValid} from './h3-utils';
 
 const H3_ANALYZER_TYPE = 'H3';
 
-// Returns true if the value is likely a WKT geometry string (heuristic check).
-const WKT_PREFIX_RE =
```
bdjulbic, line 15: Rollbacking this change from previous PR 
it wasn't covering all usecases

instead option for truncating very long tooltip

test/node/utils/dataset-utils-test.js
```diff
@@ -89,42 +88,4 @@ test('datasetUtils.isHexWkb', t => {
 
   const validEWktNDR = '0020000001000013ff0000000000400000000000000040';
   t.ok(isHexWkb(validEWktNDR), 'A valid hex ewkb in NDR should be valid');
-  t.end();
-});
-
-test('datasetUtils.isWkt', t => {
```
bdjulbic, line 95: Rollbacking this change from previous PR 
it wasn't covering all usecases

instead option for truncating very long tooltip

igorDykhta (approved):

PR #3301, opened 2026-02-06
Fix PointLayer polygon filtering for GeoJSON column mode

TanishqChavan10:
Problem: Polygon/rectangle filtering failed because getPositionAccessor returned raw WKT/GeoJSON values instead of numeric coordinates.
Fix: Accessor now returns parsed coordinates (or null), caches parsed results, and polygon filtering supports MultiPoint.


igorDykhta:
Closing in favor of https://github.com/keplergl/kepler.gl/pull/3410

PR #3310, opened 2026-02-10
fix: polygon filter not working for point layers in GeoJSON column mode

### LINKED ISSUE #3278: [Bug] Can't filter geojson column mode points by polygon
**Describe the bug**
When filtering by drawing a polygon or rectangle, it has no effect on point layers with column mode geojson.
This is happening because
1. inside the function `geojsonPosAccessor` in the file `point-layer.ts`, `d` is normally an array, so accessing the position value with `d[geojson.fieldIdx]` works and returns a string "POINT (yy xx)".<br/>But when I click filter, and the function is called from `filter-utils.ts` line 387, `d` is an object with this structure: `{ index: 0, dataContainer: {...} }`, leading to pos being `undefined`.
2. Even if we arrive at line 387 in `filter-utils.ts` with the geojson value, say like this:
    ```typescript
    export const geojsonPosAccessor =
      ({geojson}: {geojson: LayerColumn}) =>
      d => {
        if (Array.isArray(d)) {
          return d[geojson.fieldIdx];
        } else if (typeof d === 'object' && d.dataContainer) {
          return d.dataContainer.valueAt(d.index, geojson.fieldIdx);
        }
      };
    ```
    it still expects an array of coordinates, throwing `Uncaught TypeError: pos.every is not a function` on the same line.
    A solution could be just extract the coordinates from the geojson string using regex, something like
    ```typescript
    // Vibe coding solution warning
    if (isPointGeometry) {
      // Parse "POINT (lng lat)"
      const match = geometry.match(
          /POINT\s*\(\s*([-+]?[0-9]*\.?[0-9]+)\s+([-+]?[0-9]*\.?[0-9]+)\s*\)/
      );
      if (!match) return;
      lng = parseFloat(match[1]);
      lat = parseFloat(match[2]);
    }
    ```
    But it's ugly, so here I am.
    Maybe we can add an ad hoc helper function inside the point layer class to get the coordinates?
    Thoughts?

**To Reproduce**
1. Create a simple csv like this:
    ```
    point
    POINT (7.63 45.04)
    POINT (10.07 45.82)
    ```
2. Import into kepler using add data modal
3. Change the layer type from geojson to point, and make sure it uses column "point" in geojson mode
4. Draw a rectangle around a point, right click it and select the layer
5. See error in console: _Uncaught TypeError: can't access property "every", pos is undefined_ on line 387 of `filter-utils.ts`

**Expected behavior**
It should correctly filter points without throwing errors.

**Screenshots**

Image

**Additional context**
Reproduced in all demo websites and locally.

pierreeurope:
## Summary

Polygon filter (draw-to-filter) doesn't work for point layers using GeoJSON column mode. Points inside the drawn polygon are not correctly filtered.

## Root Cause

Two issues in `geojsonPosAccessor` in `point-layer.ts`:

1. **Wrong data access pattern**: When called from `getPolygonFilterFunctor` in `filter-utils.ts`, `data` is an object `{ index, dataContainer }`, but the accessor treated it as an array (`d[geojson.fieldIdx]`), returning `undefined`.

2. **No coordinate parsing**: Even with correct data access, the accessor returned raw WKT strings (e.g. `"POINT (7.63 45.04)"`) instead of `[lng, lat]` coordinates that `isInPolygon` expects.

## Fix

- Curry `geojsonPosAccessor` with `dataContainer` (matching the pattern used by `pointPosAccessor` and `geoarrowPosAccessor`)
- Use `dataContainer.valueAt()` to properly access field values
- Parse WKT POINT strings into `[lng, lat, alt]` coordinate arrays
- Support GeoJSON objects with `coordinates` property
- Return `[NaN, NaN, 0]` for invalid/null data so the existing `Number.isFinite` check in filter-utils gracefully excludes them

Fixes #3278

### DIFF
--- a/src/layers/src/point-layer/point-layer.ts
+++ b/src/layers/src/point-layer/point-layer.ts
@@ -121,10 +121,34 @@ export const pointPosAccessor =
       altitude && altitude.fieldIdx > -1 ? dc.valueAt(d.index, altitude.fieldIdx) : 0
     ];
 
+/**
+ * Parse a WKT POINT string (e.g. "POINT (7.63 45.04)") into [lng, lat] coordinates.
+ * Returns null if the string cannot be parsed.
+ */
+export function parseWktPoint(wkt: string): [number, number] | null {
+  if (typeof wkt !== 'string') return null;
+  const match = wkt.match(/POINT\s*\(\s*([-\d.eE+]+)\s+([-\d.eE+]+)\s*\)/i);
+  if (!match) return null;
+  return [parseFloat(match[1]), parseFloat(match[2])];
+}
+
 export const geojsonPosAccessor =
   ({geojson}: {geojson: LayerColumn}) =>
-  d =>
-    d[geojson.fieldIdx];
+  (dataContainer: DataContainerInterface) =>
+  (d: {index: number}): number[] => {
+    const value = dataContainer.valueAt(d.index, geojson.fieldIdx);
+    if (value == null) return [NaN, NaN, 0];
+    // Handle WKT point strings (e.g. "POINT (7.63 45.04)")
+    if (typeof value === 'string') {
+      const parsed = parseWktPoint(value);
+      return parsed ? [parsed[0], parsed[1], 0] : [NaN, NaN, 0];
+    }
+    // Handle GeoJSON objects with coordinates
+    if (value.coordinates) {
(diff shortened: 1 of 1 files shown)


pierreeurope:
Closing in favor of #3321 which addresses the same issue with a more comprehensive approach.

PR #3312, opened 2026-02-10
fix: use average instead of sum as default y-axis aggregation for time series plot

### LINKED ISSUE #3122: [Bug] Time series animation mini plot line graph is taking sum of all values for a selected y-field
**Describe the bug**
Time series animation mini plot line graph is taking sum of all values for a selected y-field on time interval that's chosen automatically.

**To Reproduce**
1. Go to https://kepler.gl/demo/earthquakes
2. Click on Y Axis and select Magnitude
3. Now on y-axis of line plot you can see that the magnitude values are 2000, 4000 and 6000 which is unrealistic for an earthquake magnitude
4. What's happening is the y field (here magnitude) get aggregated (sum) over the entire year. Now the yearly time interval is chosen automatically.

**Expected behavior**
There should be an option to say define custom time interval. Also, the aggregation type must be given as an option for the user to choose from average, sum, min, max etc., Otherwise taking sum by default yields unrealistic values that are not usable in most cases.

**Screenshots**

Image

**Desktop (please complete the following information) **
 - OS: Windows 11
 - Browser Edge
 - Version 3.1.7

pierreeurope:
The initial plot type for time filter animations defaults to AGGREGATION_TYPES.sum, causing misleading y-axis values. For example, the earthquakes demo shows magnitudes of 2000/4000/6000 instead of actual values, because all magnitudes in each time bucket are summed rather than averaged. All other aggregation defaults in the codebase use average — this was an inconsistency. One-line fix in plot.ts. Fixes #3122

### DIFF
--- a/src/utils/src/plot.ts
+++ b/src/utils/src/plot.ts
@@ -698,6 +698,6 @@ export function getDefaultPlotType(filter, datasets) {
     interval,
     defaultTimeFormat,
     type: PLOT_TYPES.histogram,
-    aggregation: AGGREGATION_TYPES.sum
+    aggregation: AGGREGATION_TYPES.average
   };
 }
--- a/test/fixtures/state-saved-v0.js
+++ b/test/fixtures/state-saved-v0.js
@@ -901,7 +901,7 @@ export const mergedFilters = [
       interval: '5-second',
       defaultTimeFormat: 'L  LTS',
       type: 'histogram',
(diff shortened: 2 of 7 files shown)


igorDykhta:
This is implemented as part of time series 

Screenshot 2026-05-24 at 9 03 28 PM

Screenshot 2026-05-24 at 9 04 06 PM

PR #3313, opened 2026-02-10
feat: add TSV (tab-separated values) file support

### LINKED ISSUE #168: Add support to TSV files
It would be good add support to `.tsv` files, due to is a very clean format, and it avoid the problem we have when text data inside the file contains commas, for example:

image

To fix that, I use the method `tsvParseRows` from `d3-dsv` library that actually you use.

This is the result of the implementation, I stay tuned to send the PR.
xefalz09qj

**Additional context**

pierreeurope:
## Summary

Adds support for uploading `.tsv` (tab-separated values) files to kepler.gl.

**Closes #168**

### 1. File extension registration (`vis-state-selectors.ts`)
- Added `'tsv'` to `DEFAULT_FILE_EXTENSIONS` so the file upload dialog accepts `.tsv` files
- Added `'TSV'` to `DEFAULT_FILE_FORMATS` so TSV appears in the supported formats list

### 2. Auto-detect tab delimiter in `processCsvData` (`data-processor.ts`)
- When raw string data is passed to `processCsvData()`, the first line is checked for tab characters
- If tabs are found, uses `tsvParseRows` from `d3-dsv` instead of `csvParseRows`
- This ensures the programmatic API (`processCsvData(rawString)`) handles TSV data correctly

### How it works

**File upload path:** The `CSVLoader` from `@loaders.gl/csv` already supports automatic delimiter detection, so `.tsv` files are parsed correctly once the extension is accepted.

**Programmatic API path:** `processCsvData()` now auto-detects tab-delimited data and switches to `tsvParseRows` accordingly.

## Testing

Minimal change — only 2 files modified, 7 lines changed. The `CSVLoader` delimiter detection is well-tested in `@loaders.gl`, and `tsvParseRows` is a stable `d3-dsv` API.

### DIFF
--- a/src/processors/src/data-processor.ts
+++ b/src/processors/src/data-processor.ts
@@ -2,7 +2,7 @@
 // Copyright contributors to the kepler.gl project
 
 import * as arrow from 'apache-arrow';
-import {csvParseRows} from 'd3-dsv';
+import {csvParseRows, tsvParseRows} from 'd3-dsv';
 import {DATA_TYPES as AnalyzerDATA_TYPES} from 'type-analyzer';
 import normalize from '@mapbox/geojson-normalize';
 import {parseSync} from '@loaders.gl/core';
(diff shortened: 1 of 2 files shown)


igorDykhta:
This is going to fail when tabs appear inside fields in valid CSV files.

PR #3318, opened 2026-02-10
fix: prevent unnecessary re-renders caused by Container subscribing to entire Redux state

### LINKED ISSUE #2626: [Bug]
**Describe the bug**
I found that any state update rerenders almost all UI. This looks like some core state management issue.

**To Reproduce**
1. Enable "Highlight updates when components render." in React Dev Tool
2. Run time animation for word_flights

**Expected behavior**
Only affected components should rerender

**Screenshots**

https://github.com/user-attachments/assets/7ef8814b-ee9a-4acb-a088-79aecde9c524

 - OS: Windows
 - Browser: Chrome
 - Version: 128.0.6613.86

pierreeurope:
## Problem

Fixes #2626 — Any state change forces React to re-render the entire UI.

## Root Cause

In `container.tsx`, the `mapStateToProps` function was passing the **entire Redux state** as a prop:

```js
const mapStateToProps = (state, props) => ({state, ...props});
```

This meant that **every Redux state change** — regardless of which slice changed — produced a new `state` reference, causing the Container component to re-render. Since Container is the root of the kepler.gl component tree, this cascaded re-renders to `KeplerGl` and all child components (SidePanel, MapContainer, BottomWidget, etc.).

## Fix

Narrowed `mapStateToProps` to only select the specific kepler.gl instance state:

```js
const mapStateToProps = (state, props) => {
  const getState = props.getState || (s => s.keplerGl);
  const keplerGlState = getState(state);
  const id = props.id || 'map';
  return { instanceState: keplerGlState ? keplerGlState[id] : undefined };
};
```

Now the Container only re-renders when its own kepler.gl instance state changes, not on unrelated Redux updates. The `KeplerGl` component already has its own Redux connection via `keplerGlConnect` that reads state slices directly from the store, so passing the full state through Container was redundant.

## Impact

- Eliminates unnecessary re-renders of the entire UI on every state change
- Particularly noticeable with time animations and large datasets
- No API changes — this is an internal optimization

### DIFF
--- a/src/components/src/container.tsx
+++ b/src/components/src/container.tsx
@@ -20,7 +20,19 @@ export const ERROR_MSG = {
 
 const mapStateToProps = (state: any, props: ContainerProps) => ({state, ...props});
 const dispatchToProps = (dispatch: Dispatch<any>) => ({dispatch});
-const connector = connect(mapStateToProps, dispatchToProps);
+
+// Only re-render when the kepler.gl instance state for this component's id changes,
+// not on every Redux state change. This fixes #2626.
+const areStatesEqual = (next: any, prev: any, nextOwnProps: any) => {
+  const getState = nextOwnProps.getState || ((s: any) => s.keplerGl);
+  const id = nextOwnProps.id || 'map';
+  const nextKeplerState = getState(next);
+  const prevKeplerState = getState(prev);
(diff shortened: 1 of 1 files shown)


igorDykhta:
Doesn't work

PR #3332, opened 2026-02-12
fix(geojson-layer): enable wrapLongitude for antimeridian crossing polygons

### LINKED ISSUE #2668: [Bug] Layer with two polygons crossing -180/180 longitude are not showing at the same time
**Describe the bug**
One layer with two polygons is showing only one at a time at earth-size zoom levels.

**To Reproduce**
1. Go to https://kepler.gl/demo
2. Drop attached GeoJSON file [minimal-example.json](https://github.com/user-attachments/files/17147875/minimal-example.json)
3. Navigate map keeping -180/180 line at the center of screen
4. Scroll left/right
**Expected behavior**
Both polygons should be visible at the same time:
image

**Screenshots**
image
image

pierreeurope:
## Summary

Enable `wrapLongitude` on GeoJSON layers so that polygons crossing the -180/180 longitude boundary (antimeridian) are rendered correctly and can be displayed simultaneously.

## Problem

Two polygons on opposite sides of the antimeridian cannot be viewed at the same time (#2668). This is because the GeoJSON layer explicitly sets `wrapLongitude: false`, overriding the base layer's default of `true`.

## Fix

Change `wrapLongitude` from `false` to `true` in the GeoJSON layer's deck.gl layer configuration. This enables deck.gl's built-in antimeridian wrapping support, which correctly handles geometries that cross the -180/180 boundary.

The base layer (`base-layer.ts`) already uses `wrapLongitude: true` — the GeoJSON layer was inconsistently overriding it to `false` without a documented reason.

Fixes #2668


lixun910 (approved):
Thank you, @pierreeurope !

igorDykhta:
This caused issues before

Issue #3139, opened 2025-06-11
Mapbox Token Usage in Kepler.gl HTML Exports


chrisgervang:
Mapbox has changed their policies for discounted access tokens they're offering to our Open Visualization projects, and they've requested we rotate our tokens. In reviewing our token usage, I've noticed that the `kepler-jupyter` package [embeds a token](https://github.com/keplergl/kepler.gl/blob/ca628b5239e9f772d9903cf6ad6db496fe752041/bindings/kepler.gl-jupyter/js/lib/keplergl/components/app.js#L34) into all maps, including maps that are exported to HTML. Going forward, I think we should assume we will only be able to use tokens on sites we host.

Unfortunately, in my testing I've found that **kepler.gl isn't able to gracefully handle an invalid token, and so we're in a situation where old maps using mapbox styles will be broken once the tokens are rotated**. This will affect most exports since the default "light" style is a Mapbox style. This isn't ideal, so I'm hoping we can find a solution together to at least help people fix their HTML files, and to prevent this issue for new maps going forward.

Brainstorming solutions for old maps I've come up with a few options so far:
 - Ask affected users to make their own Mapbox account, and replace the tokens used in their HTML files (any valid token will work, and the free-tier works should be suitable).
 - Ask Mapbox to grant usage for this kepler use case forever (their service is advertised with an annotation, so maybe they'd do it -- but it's unlikely)
 - Cut a patch of kepler 2.5, 3.0, 3.1 that switches basemaps to a token-less alternative, like Carto, and show affected users how to update the version numbers in their HTML files.

Either way, going forward I'd recommend removing [these](https://github.com/keplergl/kepler.gl/blob/ca628b5239e9f772d9903cf6ad6db496fe752041/src/constants/src/default-settings.ts#L317) mapbox styles from Jupyter package, and finding another a free solution. I'd also recommend implementing a more graceful response to an invalid map style (like switching to the "No Basemap" option if there's an error, since no one can guarantee a URL will be live forever.

wenlanzhang:
it seems that the 'export as URL' one doesn't require Mapbox token? is it because it's using the MapLibre? or is it just doesn't notify? 
is the export as HTML one still using Mapbox token, but why does the documentation saying that 'Maplibre Base Map Renderer'  is finished, and was provided by [Birk Skyum](https://github.com/birkskyum).? 

Many thanks,

PR #3345, opened 2026-02-24
feat(kepler-jupyter): version 0.4.0

lixun910:
This pull request introduces a major refactor and modernization of the KeplerGL Jupyter Python package
- [x] use modern tools (`uv` for Python, `esbuild` for TypeScript)
- [x] use widget implementation based on `anywidget`
- [x] only need to publish to pypi (keplergl-jupyter npm package is no longer needed)

Note: the function `save_to_html()` will not be supported anymore:
- Old approach: The old binding shipped a complete standalone keplergl.html file (a full kepler.gl build) specifically for HTML export. This is separate from the widget rendering.
- New anywidget approach: The widget uses index.js and index.css which are designed for the Jupyter widget protocol, not standalone HTML. They expect the anywidget runtime and communication layer.
- Users will see a helpful message explaining that the method is no longer supported and directing them to use the "Share" button in the map config panel instead.


chrisgervang (approved):
Hey @lixun910, this looks great. I have a couple thoughts

I'd be interested in seeing if we could modernize pydeck in a similar way, and ideally get some help from you to do it given how great this looks.

As I recall, the "Share" button has an export HTML feature that's more or less feature equivalent to `save_to_html()` except that it must be triggered by the user in the UI. I agree with the decision to remove it given the complexity and lack of jupyter protocol support for this kind of feature. 

Also, I think removing it also resolves a long-standing issue outlined in #3139. Could you take a look and let me know if that is true? In practice, this means users will need to supply their own token to use Mapbox-provided tiles - mostly concerning satellite tiles. Mapbox is nice enough to donate a token for use in kepler's hosted apps and docs, and has requested we not embed them in exports going forward since people re-host them and increase costs.

We've received feedback over the years whenever satellite tiles or the html export has broken, my 2cents is to make a note of these intentional changes in an upgrade guide.

Have you considered making this a major release considering it comes with a breaking change to the API?

karosc:
Just noticing this PR now. I use kepler.gl to automatically generate map artifacts directly from python. It seems like that kind of workflow would no longer supported in this version of kepler for python?

@lixun910, is that correct? In other words, to generate keplergl map artifacts in this new version requires user interaction in a jupyter widget?

lixun910:
@karosc Yes, it is for now in the 0.4.0rc1. Is the `save_to_html` is critical feature for you?

karosc:
@lixun910 : Yes, I use the python library to generate map reports automatically outside of jupyter entirely. If the official version of keplergl won't support that functionality anymore, I'll need to resort to maintaining my own fork of the older version.

lixun910:
@karosc I see. Let me work on bring the save_to_html function to v0.4.0

lixun910:
@karosc Can you help to test if version 0.4.0rc2 works as you expected. Thanks! https://pypi.org/project/keplergl/0.4.0rc2/

karosc:
@lixun910, I can take a look tomorrow and let you know if I have any issue.

karosc:
@lixun910 here is my feedback:

## Possible Bugs
1. My map config does not seem to be working. I works fine with the older version of kepler. When I load up my geodataframes in this new version, they are loaded in the map, but the styles in my config are not applied. I cannot see where the problem is. Are you able to apply a visconfig successfully using the config dict?
2. [`serialize_geodataframe`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/serializers.py#L117-L136) breaks when the gdf is empty and thus `ga.as_geoarrow` cannot resolve the geometry type. In that case, I want the layer to still show up in my map (indicating that my analysis was run, but no meaningful outputs were found). Adding this kind of logic resolves the issue for me:
	```python
	geom_series = gdf.geometry
	if geom_series.notna().any():
	    geom_array = ga.as_geoarrow(geom_series)
	else:
	    # GeoArrow cannot infer a concrete geometry spec from an empty/all-null series.
	    geom_array = ga.as_geoarrow(geom_series, type=ga.wkb())
	```
3. [`_dataset_to_geojson`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L36-L59) does not work when a geodataframe as datetime values in it. This was not a problem in the last version. I can pre-process my own data prior to inserting to the html, but it feels like if I can successfully run `map.add_data`, I should be able to run `map.export_to_html`. Although it's a bit dangerous, you can change [`data.to_json()`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L41) to `data.to_json(default=str)`, which will just push the datetime values through str. Or we could build some kind of argument into `export_to_html` that would allow us to customize data serialization.

## Possible Improvements
1. I also think it would be nice if users could specify the app theme [here](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L188-L193). I personally use the mapbox light theme, and having the Kepler UI also be light would also be nice. Event being able to add a title to the html file other then kepler-gl by changing `appName` would be a nice touch.

I am happy to assist with a PR on any of these if you'd like.

PR #3371, opened 2026-04-10
feat: add tooltip toggle

bdjulbic:
Add a show tooltip toggle button for the geojson  layer. This property was already present on the point type layer
Screenshot 2026-04-10 at 20 03 59

Just for context in layers.ts allowHover default value is true
```
  allowHover: {
    type: 'boolean',
    defaultValue: true,
    label: 'layerVisConfigs.allowHover',
    group: PROPERTY_GROUPS.interaction,
    property: 'allowHover'
  },

```


src/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -699,7 +703,7 @@ export default class GeoJsonLayer extends Layer {
       opacity: visConfig.strokeOpacity
     };
 
-    const pickable = interactionConfig.tooltip.enabled;
+    const pickable = visConfig.allowHover;
```
bdjulbic: should I use here 
interactionConfig.tooltip.enabled && visConfig.allowHover?

igorDykhta (approved):

igorDykhta (approved):

PR #3380, opened 2026-04-20
fix: fixes related to deck.gl upgrade

igorDykhta:
- [x] 1) Fix polygon tool invisible after Light & Shadow effect 2) Exclude editor layers from shadow pass
- [x] Fix extruded layers stuck with wrong shading after effect removal
- [x] Replace PBR specular with simple Lambertian diffuse for 3D tile layers
- [x] Fix tile3d sublayer cache invalidation
- [x] Higher detail 3D tiles during video/image export
- [x] Fix export detection: Replace unreliable preserveDrawingBuffer check with deck.props._isExport flag, set explicitly in PlotContainer and ExportVideoModal.
- [x] Limit sun/ambient intensity slider max from 10 to 5.


lixun910 (approved):
👍

PR #3382, opened 2026-04-21
feat(kepler-jupyter): restore save_to_html() using kepler.gl UMD bundle from CDN

lixun910:
## Summary

- Restores `save_to_html()` for the new anywidget-based Python binding (`bindings/python`), which was previously stubbed out with a deprecation message
- The exported HTML loads kepler.gl and all dependencies (React, Redux, styled-components) from the **unpkg CDN** using the pre-built UMD bundle — no local JS compilation required
- Supports all data types: `pd.DataFrame`, `gpd.GeoDataFrame`, CSV strings, and GeoJSON dicts/strings
- Defaults `center_map=True` so exported maps automatically zoom to fit the data bounds
- Adds comprehensive test suite (`test_html_export.py`) covering data conversion, HTML generation, and the widget `save_to_html()` method
- users will now need to supply their own Mapbox token for satellite tiles and other Mapbox-hosted styles: add  `mapbox_token` option when load kepler.gl in notebook and save_to_html

### How it works

Python data is converted to a format the browser can process:
- **DataFrames** → CSV string → `KeplerGl.processCsvData()` in browser
- **GeoDataFrames** → GeoJSON (reprojected to EPSG:4326) → `KeplerGl.processGeojson()` in browser
- **GeoJSON dicts/strings** → passed through → `KeplerGl.processGeojson()`
- **CSV strings** → passed through → `KeplerGl.processCsvData()`

The generated HTML follows the same pattern as `src/utils/src/export-map-html.ts` (the "Share" button export), using `addDataToMap` to load data and config into the Redux store.

### Usage

```python
from keplergl import KeplerGl

map_1 = KeplerGl(data={'data_1': df})
map_1.save_to_html(file_name='my_map.html')

# With overrides
map_1.save_to_html(data={'data_1': df}, config=config, file_name='first_map.html')
```

## Test plan

- [x] Run `pytest tests/test_html_export.py` in `bindings/python/`
- [x] Verify exported HTML opens in browser and renders the map with data
- [x] Test with DataFrame, GeoDataFrame, and GeoJSON inputs
- [x] Confirm `center_map=True` zooms to data bounds
- [x] Confirm `read_only=True` hides the side panel

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bindings/python/keplergl/serializers.py
+++ b/bindings/python/keplergl/serializers.py
@@ -17,6 +17,14 @@ def _debug(msg):
         print(f"[keplergl-serializer] {msg}")
 
 
+def _arrow_table_to_base64(table: pa.Table) -> str:
+    """Serialize an Arrow table to a base64-encoded IPC stream."""
+    sink = pa.BufferOutputStream()
+    with pa.ipc.new_stream(sink, table.schema) as writer:
+        writer.write_table(table)
+    return base64.b64encode(sink.getvalue().to_pybytes()).decode("utf-8")
+
+
 def _try_parse_geojson(data: str) -> Optional[dict]:
     """Try to parse a string as GeoJSON. Returns parsed dict if valid GeoJSON, None otherwise."""
     import json
@@ -37,10 +45,11 @@ def _try_parse_geojson(data: str) -> Optional[dict]:
 def data_to_json(data: dict, widget) -> dict:
     """Serialize Python data for JavaScript consumption."""
     _debug(f"data_to_json called with {len(data)} datasets")
+    use_arrow = getattr(widget, '_use_arrow', False) if widget else False
     result = {}
     for name, dataset in data.items():
         _debug(f"  Serializing dataset '{name}' of type {type(dataset).__name__}")
-        result[name] = serialize_dataset(dataset, name)
+        result[name] = serialize_dataset(dataset, name, use_arrow=use_arrow)
         _debug(f"  Result format: {result[name].get('format')}")
     _debug(f"data_to_json result: {list(result.keys())}")
     return result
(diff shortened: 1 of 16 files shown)


lixun910:
ref: https://github.com/keplergl/kepler.gl/pull/3345

bindings/python/keplergl/serializers.py
```diff
@@ -51,32 +60,34 @@ def data_from_json(data: dict, widget) -> dict:
...
 
-def serialize_dataframe(df: pd.DataFrame, name: str) -> dict:
-    """Serialize DataFrame to JSON format for kepler.gl.
-    
-    We use JSON format instead of Arrow to avoid version mismatch issues
-    between the apache-arrow package used here and the one bundled with kepler.gl.
+def serialize_dataframe(df: pd.DataFrame, name: str, use_arrow: bool = False) -> dict:
+    """Serialize DataFrame for kepler.gl.
+
+    When use_arrow=True, serializes as Arrow IPC (base64-encoded) which is more
```
chrisgervang, line 84: Great addition! Some notebook environments, like Databricks, have a strict 20mb output size and arrow table efficiency will really help users hitting these limits.

bindings/python/keplergl/widget.py
```diff
@@ -9,9 +9,12 @@
 import traitlets
 
 from .serializers import data_to_json, data_from_json
+from ._html_export import export_map_html
 
 _STATIC_DIR = pathlib.Path(__file__).parent / "static"
 
+KEPLER_GL_CDN_VERSION = "3.2.6"
```
chrisgervang: Is this easy to remember to update? Could be good to pull out versions to central config files, or update docs

chrisgervang (approved):
Excellent!

I'm hoping we could bring sync support back to pydeck following the patterns used here.

bindings/python/keplergl/widget.py
lixun910: Good point! Thanks!

@copilot+claude-opus-4.6 Is it possible to pull out the latest stable version (not alpha) instead of hardcoded kepler.gl version?

chrisgervang: Latest could contain breaking changes, right? Could you glob to the latest minor release like 3.*?

lixun910: Agreed!  
Copilot is getting a bit too confident: It used to ask for approval via PR, but now it’s just committing straight to the branch. 😄

lixun910: - set DEFAULT_KEPLER_GL_CDN_VERSION to 3 (semver treats @3 as the range >=3.0.0 <4.0.0 and no prerelease) 
- add CI guard that fails the build if someone bumps the root package.json major version without updating the Python constant

PR #3386, opened 2026-04-24
feat: basic flow layer implementation

igorDykhta:
- flow layer with the latest [flowmap.gl](https://github.com/FlowmapBlue/flowmap.gl)

Screenshot 2026-04-25 at 2 52 22 AM


lixun910 (approved):

src/components/src/side-panel/layer-panel/layer-configurator.tsx
```diff
@@ -1124,6 +1124,67 @@ export default function LayerConfiguratorFactory(
...
+          >
+            <VisConfigSlider
+              {...layer.visConfigSettings.flowFadeAmount}
+              {...visConfiguratorProps}
+              disabled={!layer.config.visConfig.flowFadeEnabled}
+              label={false}
+            />
+          </LayerConfigGroup>
+          <LayerConfigGroup label={'layerVisConfigs.flow.display'} collapsible>
+            <VisConfigSwitch
```
ilyabo: I think it's probably better to use a drop down for the rendering mode like in the [example](https://flowmap-layer-demo.netlify.app/) rather than checkboxes, because there's e.g. no animated curved implementation. Also, other rendering modes might be added in future.

Image

docs/user-guides/c-types-of-layers/q-flow-layer.md
```diff
@@ -0,0 +1,95 @@
...
+
+Flow layers visualize movement between locations as aggregated origin-destination flows. They are useful for displaying migration patterns, commute data, trade routes, supply chains, and any dataset that represents movement between geographic points.
+
+The layer automatically clusters nearby locations at different zoom levels, draws flow lines proportional to magnitude, and renders location totals as circles sized by incoming/outgoing volume.
+
+Flow layers are powered by [flowmap.gl](https://github.com/FlowmapBlue/flowmap.gl).
+
+## Enabling the layer
+
+The flow layer is experimental and enabled by default. You can explicitly keep it enabled by setting `enableFlowLayer: true` in the application configuration:
```
ilyabo: did you mean `disabled`?

ilyabo (approved):

docs/user-guides/c-types-of-layers/q-flow-layer.md
ilyabo: but honestly, why not enable by default?

igorDykhta: Updated to "You can disable the layer by setting `enableFlowLayer: true` in the application configuration:"

PR #3389, opened 2026-04-24
fix(kepler-jupyter): load config issue + empty geometry + json encoder

### LINKED ISSUE #3388: kepler-jupyter: 0.4.0-rc3
To address the following issues:

## Possible Bugs
1. My map config does not seem to be working. I works fine with the older version of kepler. When I load up my geodataframes in this new version, they are loaded in the map, but the styles in my config are not applied. I cannot see where the problem is. Are you able to apply a visconfig successfully using the config dict?
2. [`serialize_geodataframe`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/serializers.py#L117-L136) breaks when the gdf is empty and thus `ga.as_geoarrow` cannot resolve the geometry type. In that case, I want the layer to still show up in my map (indicating that my analysis was run, but no meaningful outputs were found). Adding this kind of logic resolves the issue for me:
	```python
	geom_series = gdf.geometry
	if geom_series.notna().any():
	    geom_array = ga.as_geoarrow(geom_series)
	else:
	    # GeoArrow cannot infer a concrete geometry spec from an empty/all-null series.
	    geom_array = ga.as_geoarrow(geom_series, type=ga.wkb())
	```
3. [`_dataset_to_geojson`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L36-L59) does not work when a geodataframe as datetime values in it. This was not a problem in the last version. I can pre-process my own data prior to inserting to the html, but it feels like if I can successfully run `map.add_data`, I should be able to run `map.export_to_html`. Although it's a bit dangerous, you can change [`data.to_json()`](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L41) to `data.to_json(default=str)`, which will just push the datetime values through str. Or we could build some kind of argument into `export_to_html` that would allow us to customize data serialization.

## Possible Improvements
1. I also think it would be nice if users could specify the app theme [here](https://github.com/keplergl/kepler.gl/blob/dd403d45adbff279b75b21f0bb214bffb6c2576f/bindings/python/keplergl/_html_export.py#L188-L193). I personally use the mapbox light theme, and having the Kepler UI also be light would also be nice. Event being able to add a title to the html file other then kepler-gl by changing `appName` would be a nice touch.

I am happy to assist with a PR on any of these if you'd like.

_Originally posted by @karosc in https://github.com/keplergl/kepler.gl/issues/3345#issuecomment-4314742036_

lixun910:
This pull request enhances the flexibility and robustness of the kepler.gl Python bindings v0.4.0 rc3:

- fix: async loadData and loadConfig may avoid config when it is loaded before data
- feat: add app name and theme customization to KeplerGl widget and HTML export
- feat: enhance GeoDataFrame serialization with custom JSON encoder (for datetime)
- fix: handle serialization of empty and all-null geometries in GeoDataFrame

Screenshot 2026-04-24 at 12 45 09 PM


lixun910:
@karosc Can you help to verify these fixes and enhancement using version 0.4.0rc3 https://pypi.org/project/keplergl/0.4.0rc3/ ? Thank you!

karosc:
@lixun910 ,can you try running this notebook in a fresh venv with the pypi wheel installed? The lines on the jupyter widget show up a nice thick cyan color in the jupyter widget (as the config specifies), but in the html file they show up a thin blue color (for me anyway). FYI, I am using python 3.12 and a Firefox browser.

[GeoDataFrame links.ipynb](https://github.com/user-attachments/files/27089197/GeoDataFrame.links.ipynb)



image
image

lixun910:
@karosc Yes, it's a bug: the `processGeojson` processes the data and the geometry column becomes _geojson. So any layer config that references the original geometry column name (e.g., "geometry") needs to be updated to "_geojson".

Just publish 0.4.0rc4 with this fix. Can you help to test it again? Thanks!

lixun910:
@karosc can you help to take a look? Thank you!

karosc:
Apologies, got very busy at work this week. I can test tonight

karosc:
@lixun910, my config works again and the app_name argument is a very nice feature. 

I think this goes a long way for users like myself who create stand-alone reports and tools. Now I can easily title my analysis reports and have that show up in the UI. Great feature! Thanks again!

lixun910:
Thanks for helping verifying the fix! @karosc I will try to push out a release for version v0.40 soon.

igorDykhta (approved):
Great!

PR #3393, opened 2026-04-29
chore: upgrade react-map-gl to 8, maplibre-gl to 4

igorDykhta:
- Upgrade react-map-gl from ^7.1.6 to ^8.1.1 (aligning with the version used by @sqlrooms/deck) and maplibre-gl from ^3.6.2 to ^4.0.0 (required by react-map-gl v8).

Changes:

- [x] Update react-map-gl and maplibre-gl versions in root, @kepler.gl/components, and @kepler.gl/utils package.json files
- [x] Migrate imports from react-map-gl to react-map-gl/mapbox-legacy (required by v8 for mapbox-gl v1.x compatibility)
- [x] Replace removed MapboxMap type with MapInstance; remove generic param from MapLib
- [x] Remove @types/react-map-gl (v8 ships built-in TypeScript types)

Notes:
The mapbox-legacy endpoint is used because kepler.gl depends on mapbox-gl@1.13.1. It preserves the mapLib prop used for dynamic switching between mapbox and maplibre at runtime.


lixun910 (approved):
🚀

src/components/src/map-container.tsx
```diff
@@ -4,7 +4,7 @@
 // libraries
 import React, {Component, createRef, useMemo} from 'react';
 import styled, {withTheme, useTheme} from 'styled-components';
-import {Map, MapRef} from 'react-map-gl';
+import {Map, MapRef} from 'react-map-gl/mapbox-legacy';
```
lixun910, line 7: MapLibre GL v5 changed the Transform API — zoom, pitch, bearing, center became getter-only properties (values are set via setZoom(), setPitch() methods instead). But react-map-gl/mapbox-legacy's applyViewStateToTransform does tr.zoom = v.zoom (direct setter assignment), which throws.

We might need to replace `import {Map, MapRef} from 'react-map-gl/mapbox-legacy'` with both adapters: MapboxLegacyMap from react-map-gl/mapbox-legacy and MaplibreMap from @vis.gl/react-maplibre

Here is a PR https://github.com/keplergl/kepler.gl/pull/3395 
cc @igorDykhta

PR #3395, opened 2026-04-29
fix: integrate Maplibre support and update dependencies

lixun910:
MapLibre GL v5 (e.g. in sqlrooms/deck) changed the Transform API — zoom, pitch, bearing, center became getter-only properties (values are set via setZoom(), setPitch() methods instead). But react-map-gl/mapbox-legacy's applyViewStateToTransform does tr.zoom = v.zoom (direct setter assignment), which throws errors.

This PR replaces `import {Map, MapRef} from 'react-map-gl/mapbox-legacy'` with both adapters: 
- MapboxLegacyMap from `react-map-gl/mapbox-legacy` and
- MaplibreMap from `@vis.gl/react-maplibre`

Changes:
- Add Maplibre as a new Map component from @vis.gl/react-maplibre.
- Updated the MapComponent prop to allow selection between Mapbox and Maplibre.
- Updated package.json to include @vis.gl/react-maplibre and adjusted react-map-gl versioning (hubble.gl still pins older version)
- Cleaned up imports and adjusted map properties to accommodate the new library integration.


igorDykhta (approved):
Looks good and works. I haven’t found any new issues during testing.

PR #3410, opened 2026-05-03
fix: Fix PointLayer polygon filtering for GeoJSON column mode

igorDykhta:
- fix: Fix PointLayer polygon filtering for GeoJSON column mode
- fix highlighting in Point Layer in geojson mode
- instead of the stale PR here https://github.com/keplergl/kepler.gl/pull/3301
- Thanks @TanishqChavan10 for bringing up this issue!


lixun910 (approved):

leverglowh:
This should fix #3278 
Other stale pr (still open): #3321

PR #3412, opened 2026-05-03
fix: arcgis tile 3d model crash due to unknown coord system (post deck.gl upgrade)

igorDykhta:
- fix: arcgis tile 3d model crash due to unknown coord system (post deck.gl upgrade)
- this is a regression after deck.gl 9.3.1 upgrade


lixun910 (approved):
👍

PR #3436, opened 2026-05-13
feat: support MapLibre V5 in hubble.gl tool 

lixun910:
Hubble's ExportVideoPanelPreview internally calls `applyViewStateToTransform`, which directly assigns to the map's transform properties (tr.bearing = ..., tr.pitch = ..., etc.).

In MapLibre v4, transform.bearing and similar properties had writable setters, so this worked fine.

In MapLibre v5, these properties were changed to getter-only (read-only accessors). When the Mapbox adapter tries to write to them, it throws error.

The fix provide a compatibility shim: `getHubbleMapLib()` dynamically imports maplibre-gl and returns a patched version:

- Subclasses maplibre.Map as HubbleMapLibreMap
- After construction, patches the transform instance via patchMapLibreTransformForHubble()
- Redefines getter-only properties (bearing, pitch, zoom, padding, _center, _zoom) to add setters that delegate to the v5 setter methods (setBearing(), setPitch(), etc.)
- Stubs out internal methods (_constrain, _calcMatrices) that Hubble's adapter may call but don't exist on v5 transforms

### DIFF
--- a/src/components/src/modals/hubble-utils.ts
+++ b/src/components/src/modals/hubble-utils.ts
@@ -177,7 +177,11 @@ function patchMapLibreTransformForHubble(transform: any): void {
   if (!Object.getOwnPropertyDescriptor(transform, '_setZoom')) {
     Object.defineProperty(transform, '_setZoom', {
       value(this: any, zoom: number) {
-        this.setZoom(zoom);
+        if (typeof this.setZoom === 'function') {
+          this.setZoom(zoom);
+        } else {
+          this.zoom = zoom;
+        }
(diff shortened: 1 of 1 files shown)


lixun910:
hmm it seems the video quality is also bad (setting High quality) with maplibre v5:

kepler gl

igorDykhta:
Screenshot 2026-05-13 at 8 39 42 PM

chrisgervang:
Would you like to patch hubble.gl directly for native v5 compatibility?

PR #3438, opened 2026-05-15
feat: zoom and compass control

igorDykhta:
- zoom and compass control
- instead of the stale https://github.com/keplergl/kepler.gl/pull/3328

Screenshot 2026-05-15 at 6 58 11 PM


lixun910 (approved):
🆒

PR #3439, opened 2026-05-16
fix: prevent unnecessary UI re-renders

igorDykhta:
**NOT TO LAND - PERFORMANCE TESTS, TO FRAGILE TO LAND ANYWAY**

- for #2626 
- prevent unnecessary UI re-renders on 1) animation update 2) mouse move
- instead of stale non-functional https://github.com/keplergl/kepler.gl/pull/3318 

Screenshot 2026-05-16 at 4 43 26 AM


src/components/src/map/map-control.tsx
```diff
@@ -143,7 +143,66 @@ function MapControlFactory(
 
   MapControl.displayName = 'MapControl';
 
-  return MapControl;
+  const areMapControlPropsEqual = (prev: MapControlProps, next: MapControlProps): boolean => {
+    const keys = Object.keys(next) as (keyof MapControlProps)[];
+    for (const key of keys) {
```
ilyabo: wouldn't some `deepEqual` simplify this?

ilyabo (approved):

PR #3441, opened 2026-05-16
chore: Convert class components to functional components

igorDykhta:
- Convert class components to functional components
- instead of the stale, outdated https://github.com/keplergl/kepler.gl/pull/3137 PR

Here's the list of components updated in this adaptation of PR #3137:

src/components/src/effects/effect-panel.tsx — Converted from class to functional component. Removed PropTypes, added ActionHandler<> type wrappers, replaced class methods with useCallback hooks.

src/components/src/notification-panel.tsx — Converted from class to functional component. Removed React.memo wrapper workaround, changed return type from React.ComponentClass to React.FC.

src/components/src/notification-panel/notification-item.tsx — Converted from class to functional component. Replaced class state with useState, replaced componentDidMount with useEffect.

src/components/src/modals/add-map-style-modal.tsx — Converted from class to functional component. Replaced getDerivedStateFromProps with useEffect, replaced this._map with useRef, removed react-lifecycles-compat polyfill.

src/components/src/modals/data-table-modal.tsx — Converted from class to functional component. Replaced createSelector (reselect) with useMemo, replaced class methods with useCallback, replaced instance cache with useRef, removed @ts-expect-error on withTheme.

src/components/src/modals/export-data-modal.tsx — Converted from class to functional component. Replaced componentDidMount with useEffect, replaced class method _onSelectDataset with a local function.

### DIFF
--- a/src/components/src/effects/effect-panel.tsx
+++ b/src/components/src/effects/effect-panel.tsx
@@ -1,13 +1,12 @@
 // SPDX-License-Identifier: MIT
 // Copyright contributors to the kepler.gl project
 
-import React, {Component, ComponentType} from 'react';
+import React, {useCallback, ComponentType} from 'react';
 import styled from 'styled-components';
 import classnames from 'classnames';
-import PropTypes from 'prop-types';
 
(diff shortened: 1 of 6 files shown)


lixun910 (approved):
👍

PR #3451, opened 2026-05-20
feat: improvements to the trip layer

igorDykhta:
Improve Trip Layer
- Add 3D scenegraph model support to Trip layer (airplane, car, helicopter, etc.) with orientation controls (roll/pitch/yaw fields and manual adjustments)
- Add table column mode allowing Trip layer to work with flat CSV data (lat/lng/timestamp/id columns) without requiring pre-built GeoJSON
- Add text labels that follow animated position along the path, using the standard TextLabelPanel (removed the redundant TextLabelTooltipPanel)
- Add color-by-field visual channel support for scenegraph models, including an option to use trail color
- Add 11 built-in 3D model icons and a scenegraph model selector with custom URL input

Screenshot 2026-05-21 at 2 27 43 AM


lixun910 (approved):
😍

PR #3544, opened 2026-07-22
fix: heatmap issues on zoom out

igorDykhta:
## Issue
In 2D/top mode the heatmap layer disappears or renders mirrored/leaking when zoomed out (multiple world copies) or panned across the antimeridian.

## Fix
Render the heatmap quad over the data's extent, pinned to a single world copy (like other layers), instead of tying it to the screen corners — and disable `wrapLongitude` on the flat quad so it isn't re-wrapped/mirrored.

Before:
Screenshot 2026-07-22 at 11 59 37 PM

After: 
Screenshot 2026-07-22 at 11 59 43 PM


lixun910 (approved):
👍

PR #2283, opened 2023-06-23
[Bug][jupyter] Serialization error on GeoDataFrames with multiple geometry columns

KrausMatthias:
When passing a GeoDataFrame with multiple geometry columns, add_data fails with a serialization error, as currently only the "geometry" column is converted to WKT.

This uses [to_wkt](https://github.com/geopandas/geopandas/blob/main/geopandas/geodataframe.py#L1017) introduced by GeoPandas in version 0.9.0 which converts all columns of type geometry in the dataframe to WKT.

Besides the issue with multiple geometry columns this also fixes:

* Fixes #1758, Fixes #1759.
* Fixes #1240, as to_wkt in geopandas does a copy internally, which avoids overwriting the geometry column with the WKT string column.

PR #3590, opened 2026-07-28
fix: serialize GeoDataFrames with multiple geometry columns in python binding

igorDykhta:
Adapts the fix from #2283 (originally by @KrausMatthias) to the new `bindings/python` package.

## Problem

When a `GeoDataFrame` has more than one geometry-typed column, `serialize_geodataframe` called `pa.array()` directly on the non-active `GeoSeries` columns, raising:

```
ArrowInvalid: Could not convert POLYGON (...) with type Polygon: did not recognize Python value type
```

## Fix

- **Active geometry column** → encoded as GeoArrow (preserves rich geometry type for kepler.gl rendering)
- **Secondary geometry columns** → converted to WKT strings via `series.to_wkt()`, so they pass through Arrow cleanly and are visible as text attributes in the kepler.gl data table

## Tests

Three new regression tests in `TestGeoDataFrameSerialization`:
- GDF with two geometry columns (active + buffered)
- GDF where the active geometry is not the first geometry column
- GDF with a secondary geometry column that is all-null

Closes #2283

### DIFF
--- a/bindings/python/keplergl/serializers.py
+++ b/bindings/python/keplergl/serializers.py
@@ -114,24 +114,44 @@ def _serialize_dataframe_arrow(df: pd.DataFrame, name: str) -> dict:
     }
 
 
-def serialize_geodataframe(gdf: gpd.GeoDataFrame, name: str) -> dict:
-    """Serialize GeoDataFrame to GeoArrow format."""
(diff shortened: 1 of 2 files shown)


lixun910 (approved):
Thanks!

PR #3591, opened 2026-07-28
chore: upgrade Python bindings to kepler.gl 3.3.0-alpha.4 (React 19)

igorDykhta:
## Summary

- Bump all `@kepler.gl/*` packages to `3.3.0-alpha.4` and align peer deps: React 19, Redux 5, React-Redux 9, TypeScript 5.6.
- Replace UMD `<script>` tags in the standalone HTML export with an ESM import map + `es-module-shims`, since React 19 / Redux 5 / React-Redux 9 no longer publish UMD builds. The shim keeps exported maps working when opened as a local `file://` document.
- Fix Redux v5 TypeScript stricter typing in `store.ts` (`StoreEnhancer` cast).
- Update CI version-check to handle full semver strings (e.g. `3.3.0-alpha.4`) in addition to bare major tags.

## Test plan

- [x] `npm install && npm run build` in `bindings/python` — JS bundle compiles cleanly.
- [x] `npx tsc --noEmit` — no TypeScript errors.
- [x] `uv run pytest tests/ -v` — all 101 tests pass.
- [x] Open a saved `.html` export in a browser (both via HTTP and from `file://`) and verify the map loads with kepler.gl 3.3.0-alpha.4.

### DIFF
--- a/.github/workflows/build-publish-pypi.yml
+++ b/.github/workflows/build-publish-pypi.yml
@@ -23,9 +23,10 @@ jobs:
         run: |
           MAJOR=$(node -p "require('./package.json').version.split('.')[0]")
           CDN=$(grep -Po '(?<=DEFAULT_KEPLER_GL_CDN_VERSION = ")[^"]+' bindings/python/keplergl/_html_export.py)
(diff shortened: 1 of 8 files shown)


lixun910 (approved):
Thank you! 👍

PR #3596, opened 2026-07-30
feat: add map scale bar

igorDykhta:
## Summary

- Adds a `MapScale` component that renders a real-world distance scale bar at the bottom-left of each map panel
- Scale is calculated by unprojecting two screen points 100 px apart and measuring the geodesic distance between them, so it stays accurate at any zoom level and latitude
- Clicking the bar toggles between **km/m** and **mi/ft**; the preference is persisted in `localStorage`
- Scale bar is wired into the kepler.gl factory system (`MapScaleFactory`) and exported from `@kepler.gl/components`
- The loading indicator is offset upward to avoid overlapping the scale bar
- The feature is **opt-out** via `initApplicationConfig({ enableMapScale: false })`

Screenshot 2026-07-30 at 11 53 34 PM

### DIFF
--- a/src/components/package.json
+++ b/src/components/package.json
@@ -65,6 +65,7 @@
     "@mapbox/mapbox-sdk": "^0.15.3",
     "@math.gl/web-mercator": "^4.1.0",
     "@tippyjs/react": "^4.2.0",
+    "@turf/distance": "^7.2.0",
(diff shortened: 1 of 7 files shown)


lixun910 (approved):
👍

PR #3612, opened 2026-08-04
feat: add loading spinner to exported HTML maps

igorDykhta:
Adds a loading spinner to exported HTML maps so users see feedback while CDN bundles are downloading, instead of a blank white screen.

## Changes
- Added a full-screen loading overlay with a kepler.gl-branded spinner and "Loading map…" label — shown immediately on page open
- The spinner is automatically replaced by the map when React renders (no extra JS needed to dismiss it)
- Improved error handling: if the UMD bundle fails to load (network drop, CDN error), the spinner is replaced with a user-readable error message instead of a silent blank screen

## Related
- Follows up on #3610 (blank screen bug fix)


lixun910 (approved):
👍

PR #3653, opened 2026-08-21
chore: replace lodash with es-toolkit

### LINKED ISSUE #3626: Request migrate lodash to es-toolkit
How about migrate lodash to [es-toolkit](https://github.com/toss/es-toolkit) ?

 I noticed that eight published `@kepler.gl/*` packages carry lodash in `dependencies`, each importing via subpath (`lodash/pick`, `lodash/uniq`,
 etc.) across 47 source files — 13 distinct functions in total. Since these are browser-bundled, every consumer app includes the lodash slice.
 Swapping to es-toolkit/compat brought that slice from 16.2 kB to 5.3 kB gzip, about 67% smaller. The migration also drops eight `@types/lodash`
 entries, since es-toolkit ships its own types.

 I ran the Jest suite after applying the swap: 129 of 130 passed, 8 of 15 suites green — identical to the baseline before the change (the 7 failing
 suites all hit a missing `maplibregl-mapbox-request-transformer` module unrelated to lodash). The diff is 84 files, +125/−111, almost entirely
 mechanical one-line import changes. One manual fix is needed: `MemoizedFunction` in `cluster-utils.ts` needs a small inline type since es-toolkit
 doesn't export that type name.

I think this migration would bring significant benefits to users!

igorDykhta:
## Summary
- Swap `lodash/*` imports for `es-toolkit/compat` in the published packages, tests, website, and demo-app.
- Depend on `es-toolkit` as a regular `^1.39.10` dependency (not a peer) so hosts such as Studio can share one copy.
- Drop `@types/lodash`. `MemoizedFunction` in `cluster-utils.ts` is a small local type.

Closes #3626

## Test plan
- [x] `yarn typescript`
- [x] Jest
- [x] Node tape tests


lixun910 (approved):
👍

PR #3666, opened 2026-08-26
feat: add GeoHash layer

### LINKED ISSUE #989: New layer type: GeoHash
**Is your feature request related to a problem? Please describe.**
no

**Describe the solution you'd like**
I would like to display data based on GeoHash, which is a standard in several platforms. 

for instance a CSV like the following:
geohash,kpi1
dr5rg,12
dr5r8,42
dr5rv,566

**Describe alternatives you've considered**
I see that you provide also H3 layer type, it would be interesting to have also the geohash layer

**Additional context**
The new feature should be easy to implement.

igorDykhta:
## Summary
- Add a GeoHash layer (same pattern as A5/S2) that renders [geohash](https://en.wikipedia.org/wiki/Geohash) cells from a token column via deck.gl `GeohashLayer`.
- Auto-detects columns named `geohash`, `geo_hash`, or `geohash_id`. Optional and **enabled by default**; disable with `initApplicationConfig({enableGeohashLayer: false})`.

Closes https://github.com/keplergl/kepler.gl/issues/989

## Test plan
- [x] Load a CSV with a `geohash` column and confirm cells render (fill/stroke/3D).
- [x] Confirm the layer type is hidden when `enableGeohashLayer` is `false`.
- [x] Confirm auto-detect from `geohash` / `geo_hash` / `geohash_id` column names.

Screenshot 2026-08-26 at 4 54 47 AM


lixun910 (approved):
👍

Issue #322, opened 2019-01-09
render realtime data coming from websocket for example


rcarlomagno:
i think there are use case as mine, where data is not fixed, instead, it is generated in realtime
for example,in my case, i would like to view realtime data generated from users events (what and where users are doing)
could it be possible?

heshan0131:
We don't plan to support streaming data yet in our roadmap. But this is a valid usecase, we will put it in our backlog.

omegadeep10:
Currently we run a regular cron job to generate new CSV files every 15 minutes. We are using the single-file HTML version of Kepler, and we've configured it to load the CSV file dynamically, and refresh the page every 15 minutes.

It's not real-time, but it works well enough for our use-case. However, this is only feasible because we are only working with a relatively small amount of data (about 300k total rows across 4 files).

aakarsh:
Is there any update for this ? Is streaming data support coming soon ?

gohkokhan:
> Currently we run a regular cron job to generate new CSV files every 15 minutes. We are using the single-file HTML version of Kepler, and we've configured it to load the CSV file dynamically, and refresh the page every 15 minutes.
> @omegadeep10  Can you explain more on how you configured it to load the CSV file dynamically? I'm doing something similar right now. Cron job to get the latest data, and want to feed it into the HTML file. 
> It's not real-time, but it works well enough for our use-case. However, this is only feasible because we are only working with a relatively small amount of data (about 300k total rows across 4 files).

@omge

omegadeep10:
@gohkokhan 

1. We setup nginx to serve the HTML and CSV files. 
2. In the Kepler HTML file, I load the config JSON and any data files dynamically using `fetch`. 
3. I then process each CSV file using the built-in `KeplerGl.processCsvData` function, assigning it a predetermined id.
4. After that, I add the processed datasets and the config to the Kepler instance.

Please see this relevant section of the code. The full example is attached as well.

[example.kepler.zip](https://github.com/keplergl/kepler.gl/files/4431882/example.kepler.zip)

```<script>
      /**
       * Customize map.
       * Interact with map store to customize data and behavior
       */
      (function customize(keplerGl, store) {
        // store.dispatch(keplerGl.toggleSplitMap());
        // console.log(KeplerGl)

        // here we are loading data files dynamically from the same root (nginx configured correctly)
        // These data files are updated daily with a cron job
        var urlParams = new URLSearchParams(window.location.search);
[...]

PR #3667, opened 2026-08-26
feat: reload and poll remotely hosted datasets

igorDykhta:
## Summary
- Add a reload control and optional poll interval (10s–15m) for remotely hosted datasets, so Kepler can replace rows in place when the remote file changes.
- Unchanged files skip a re-parse via `If-None-Match` / 304. Layers stay up if field names match.
- `examples/live-data` serves a CSV that updates every 10s (`yarn start:live-data`).

Follow-up to #3662.

## Test plan
- [x] Load a remote CSV/GeoJSON URL, click reload, confirm rows update and the layer is kept
- [x] Set refresh to 10s against `yarn start:live-data` and confirm points move / snapshot increments
- [x] Unchanged remote file returns 304 and does not rebuild the table

https://github.com/user-attachments/assets/b42ee648-ae71-4a5e-9071-eb5e6607b814


lixun910 (approved):

igorDykhta:
For #322 (related)

PR #3692, opened 2026-09-02
fix(layers): trip layer color-by-field reads correct row in TABLE mode

lixun910:
It looks like a regression:

## Summary

Fixes the trip layer's color-by-field accessor in TABLE column mode.

In TABLE column mode, each feature's `properties.index` is the **feature index** (trip ordinal), not a row index into the data container. The color accessor was resolving field values via `dc.valueAt(featureIndex, fieldIdx)`, which reads the value at the feature-index row — the wrong trip's value whenever the first trip spans more rows than the feature count. When the first trip has many points, every trip resolves to the first trip's category (e.g. all green for an Uber-leading dataset).

## Change

Read field values from the trip's first row (`d.properties.values[0]`) in TABLE mode, matching the behavior of `efb072eb` that `9194faaf` reverted. GEOJSON mode keeps resolving by row index.

## Tests

Adds `trip-layer-color.spec.ts` covering the TABLE-mode color accessor. Verified passing on `master`:

```
Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
```

> Note: this is a cherry-pick of `d794b817` from the `xli-kepler-mcp` branch, which is not being merged for now.

### DIFF
--- a/src/layers/src/trip-layer/trip-layer-color.spec.ts
+++ b/src/layers/src/trip-layer/trip-layer-color.spec.ts
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: MIT
+// Copyright contributors to the kepler.gl project
+
+import TripLayer from './trip-layer';
+import KeplerTable from '@kepler.gl/table';
+import {ALL_FIELD_TYPES, SCALE_TYPES} from '@kepler.gl/constants';
+
+/**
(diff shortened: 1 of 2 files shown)


igorDykhta (approved):
Yes, a definite regression

PR #3695, opened 2026-09-04
feat(demo): bump kepler-assistant to 0.0.18 so anthropic works in the browser

lixun910:
## Summary

Bumps both assistant demos to `@openassistant/kepler-assistant@0.0.18`, and reworks the SQLRooms demo's model factory so anthropic works from the browser.

This supersedes the `^0.0.16` pin that was on this branch: 0.0.16 predates the anthropic fix, so pinning to it would not have resolved the report below. The 0.0.16 changes are still included — see the section at the bottom.

## The bug (jo-chemla's report)

Selecting Anthropic with an API key on `/demo` or `/demo-next` fails on every message:

```
Error: Failed to fetch
400 Bad request, strict-origin-when-cross-origin
```

Anthropic refuses a browser-origin call unless the request carries `anthropic-dangerous-direct-browser-access: true`. Without it the CORS **preflight** is rejected outright:

```
$ curl -si -X OPTIONS https://api.anthropic.com/v1/chat/completions \
    -H 'Origin: https://kepler.gl' \
    -H 'Access-Control-Request-Method: POST' \
    -H 'Access-Control-Request-Headers: authorization,content-type'
HTTP/2 400
(no access-control-allow-origin)
```

so the browser discards the real request and `fetch` rejects — that is the `Failed to fetch` above, and why the API's own response never appears. Listing the header in `Access-Control-Request-Headers` makes the same preflight return `HTTP/2 200` with `access-control-allow-origin: *`.

Two of the report's other conclusions do not hold, worth recording here: `/v1/chat/completions` **is** the correct endpoint for Anthropic's OpenAI-compatible layer, and `Authorization: Bearer <key>` **is** accepted there (a dummy key returns `401 authentication_error: Invalid Anthropic API key`, not an auth-format error). No localhost CORS proxy is needed — one request header is sufficient.

## What 0.0.18 does

`@sqlrooms/ai-core` builds every provider through a single `createOpenAICompatible(...)` client with no way to pass request headers, so kepler-assistant 0.0.18 builds the transport itself through the existing `AiSliceOptions.getCustomModel` hook (geodaai/kepler-assistant#6). That fixes `/demo`, where the packaged `AiAssistantPanel` owns the `createAiSlice` call, with no change on this side beyond the pin.

`/demo-next` builds its own store and passes its own `getCustomModel`, so kepler-assistant's store wiring is not on its path. Hence the change in this PR:

- `examples/demo-app-sqlrooms/src/components/assistant-model.ts` gains `getChatModel`, which keeps `getOpenAiModel` for openai and delegates every other provider to kepler-assistant's `getChatModel` from the `/integration` subpath, where the header lives. openai keeps its own client deliberately: it uses the OpenAI **Responses** API plus the `providerOptions.openai.store = false` middleware that carries encrypted reasoning between tool steps, and the other providers do not speak that API — routing openai through the OpenAI-compatible client would silently downgrade it.
- `examples/demo-app-sqlrooms/src/components/assistant.tsx` passes that as `getCustomModel`.
- The skill sub-agent path needed no change: `assistant-tools.ts` already falls back to kepler-assistant's `getModel`, which 0.0.18 fixed as well.

## What's in 0.0.16 (included in this branch)

The kepler-assistant release `v0.0.16` (commits `89feb08` + `4439335`) fixes the AI Assistant so first-time users can select a model:

Before:
Screenshot 2026-09-04 at 8 37 04 AM

After:
Screenshot 2026-09-02 at 5 33 55 PM

- **Before:** the stock `@sqlrooms/ai-core` `Chat.ModelSelector` returns `null` when no chat session exists. Since the demo-app creates the session lazily (only on the first message), a first-time user saw "Enter your OpenAI API key..." with no way to pick a provider/model.
- **After:** a sessionless model selector (`SessionlessModelSelector`) renders in the composer even before a session exists. It reads `ai.getSelectedModel()`, groups models by provider in a Radix Select, and on change either calls `setAiModel` (session exists) or `createSession(undefined, provider, model)` so the selection persists and the API-key hint updates to the chosen provider.

## Changes

- `examples/demo-app/package.json`: `@openassistant/kepler-assistant` `^0.0.15` -> `^0.0.18`
- `examples/demo-app/yarn.lock`: lockfile updated
- `examples/demo-app-sqlrooms/package.json`: `@openassistant/kepler-assistant` `0.0.17` -> `0.0.18`
- `examples/demo-app-sqlrooms/yarn.lock`: lockfile updated
- `examples/demo-app-sqlrooms/src/components/assistant-model.ts`: new `getChatModel` factory
- `examples/demo-app-sqlrooms/src/components/assistant.tsx`: pass it as `getCustomModel`
- `examples/demo-app-sqlrooms/test/assistant-model.test.cjs`: keep the new import external to the node-only test bundle, and assert the delegation

## Verification

- Both demos install `0.0.18` with `dist/index.js` checksum `0d781821084fed4b402e32466335c0736e60f8c5`, matching the published npm tarball.
- Both demos build clean (`node esbuild.config.mjs --build`), and the header string is present in both emitted bundles.
- `tsc`, `eslint`, and `prettier` clean on the changed files; `node --test examples/demo-app-sqlrooms/test/*.test.cjs` passes 16/16.

Note for anyone reproducing: verifying this needs a real Anthropic key, since the fix is specifically about what the browser is allowed to send. Without one, the check is that a deliberately invalid key yields the API's `401 authentication_error: Invalid Anthropic API key` instead of `Failed to fetch`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jo-chemla:
Probably related hence I did not want to open a new thread.

I wanted to test the keplergl AI assistant on the live demo apps, hence tested on both [/demo](https://kepler.gl/demo) and [/demo-next](https://kepler.gl/demo-next). I've opened the AI Assistant tab, went to settings to set Anthropic API key generated from https://platform.claude.com/settings/keys (tested with Personal and Legacy key, scoped to default workspace)

UI shows `Error: Failed to fetch`
Looking at browser network request tab I see  the requests sent by keplergl ai assistant always return `400 Bad request, strict-origin-when-cross-origin`. Is this expected? I've got an allow-cors plugin, so tested both with and without, but same result. Also searched the keplergl and [geodaai/openassistant](https://github.com/geodaai/openassistant) repos for CORS/cross origin mentions but could not find any. 

After trying to work around the issue with claude, it explained the following: 

> **What's going wrong**
> 
> - `demo-next` builds its assistant on the `@sqlrooms/ai` library. For every provider except OpenAI, that library sends the request in OpenAI format to `https://api.anthropic.com/v1/chat/completions` with an `Authorization: Bearer` header.
> - It never sends the `anthropic-dangerous-direct-browser-access: true` header, which Anthropic requires for calls made directly from a browser.
> - Without that header, Anthropic answers the browser's CORS preflight (the automatic `OPTIONS` check before the real request) with **400 and no `access-control-allow-origin`**. That is the 400 and the CORS error you're seeing. With the header listed in the preflight, I got 200.
> - The Allow-CORS plugin can't fix this. It rewrites response headers, but the preflight still comes back as 400, which makes Chrome drop the request. Even if it got past that, the real POST would get a 401 for the same missing header.
> - A fetch from `https://kepler.gl` straight to Anthropic, sent the way the demo sends it, failed with `Failed to fetch`.

Note the agent also wrote a minimal `anthropic-cors-proxy.mjs`, a node proxy that listens on localhost only, accepts requests only from `kepler.gl` and localhost pages, answers the preflight itself, and forwards the call to Anthropic as a normal server-side request. 

<details><summary>The workaround node proxy script</summary>
<p>

To use it:

1. The proxy is already running from this session. To start it yourself later, run `node C:\Users\jonathan\anthropic-cors-proxy.mjs`.
2. In the kepler.gl assistant settings, change the Anthropic base URL from `https://api.anthropic.com/v1` to `http://localhost:8787/v1`.
3. Turn off the Allow-CORS plugin, since it can interfere.


</p>
</details> 

<details><summary>Details</summary>
<p>

```js

// Local CORS proxy for using Anthropic from browser apps (e.g. kepler.gl AI assistant)
// that don't send the 'anthropic-dangerous-direct-browser-access' header.
// Usage: node anthropic-cors-proxy.mjs   -> set base URL to http://localhost:8787/v1
import http from 'node:http';
import https from 'node:https';

const PORT = process.env.PORT || 8787;
const UPSTREAM = 'api.anthropic.com';
// Only these pages may use the proxy
const ALLOWED_ORIGINS = [/^https:\/\/kepler\.gl$/, /^http:\/\/localhost(:\d+)?$/];

http
  .createServer((req, res) => {
    const origin = req.headers.origin || '';
    if (origin && !ALLOWED_ORIGINS.some(re => re.test(origin))) {
      res.writeHead(403).end('origin not allowed');
      return;
    }
    const cors = {
      'access-control-allow-origin': origin || '*',
      'access-control-allow-methods': 'GET, POST, OPTIONS',
      'access-control-allow-headers': req.headers['access-control-request-headers'] || '*',
      'access-control-allow-private-network': 'true',
      'access-control-max-age': '600',
      vary: 'Origin'
    };
    if (req.method === 'OPTIONS') {
      res.writeHead(204, cors).end();
      return;
    }

    // Forward as a server-side request: drop browser-only headers so Anthropic
    // doesn't treat it as a CORS call.
    const headers = {...req.headers, host: UPSTREAM};
    for (const h of Object.keys(headers)) {
      if (h === 'origin' || h === 'referer' || h.startsWith('sec-')) delete headers[h];
    }

    const upstream = https.request(
      {host: UPSTREAM, path: req.url, method: req.method, headers},
      up => {
        const out = {...up.headers};
        for (const h of Object.keys(out)) if (h.startsWith('access-control-')) delete out[h];
        res.writeHead(up.statusCode, {...out, ...cors});
        up.pipe(res);
        console.log(`${req.method} ${req.url} -> ${up.statusCode}`);
      }
    );
    upstream.on('error', e => {
      res.writeHead(502, cors).end(String(e));
    });
    req.pipe(upstream);
  })
  .listen(PORT, '127.0.0.1', () =>
    console.log(`Anthropic CORS proxy: http://localhost:${PORT}/v1 -> https://${UPSTREAM}/v1`)
  );


```

</p>
</details>

lixun910:
Thanks @jo-chemla ! I will take a look at it today. Will keep you updated.

lixun910:
@jo-chemla Could you help testing if this build fixes the issue: https://deploy-preview-3695--keplergl.netlify.app/demo Thank you!

jo-chemla:
Yes I can ask Claude Opus or Sonnet questions about the datasets on kepler after inputting in params my `sk-ant-api03-Rtm...` keys on this deployed instance, thanks!

ilyabo (approved):

PR #3699, opened 2026-09-05
feat: optional end time for time filter animation

### LINKED ISSUE #3198: [Feature Request] Support for Start/End DateTime Ranges in Animation
Hi Kepler team 👋

First off, huge thanks for the work you've done on this project — Kepler.gl is an incredibly powerful and elegant tool for visualizing large-scale geospatial data.

I'm reaching out to request a feature enhancement that would greatly improve temporal visualizations for service availability, lifecycles, coverage, and other time-span-based data.

🚀 Feature Request

Support animation of date/time ranges using start_time and end_time fields, allowing features to persist across multiple animation frames.

🎯 Use Case

Right now, Kepler.gl’s animation system only supports a single timestamp field for filtering and playback — a feature disappears when the animation frame does not match that exact timestamp.

However, many real-world use cases involve durations, such as:

Visualizing service coverage across time (start_date → end_date)

Tracking presence or activity windows

Showing infrastructure, deployments, or availability during time intervals

🔧 Proposed Behavior

Allow users to define two fields (e.g., start_datetime and end_datetime) and enable an animation mode where:

A feature is visible in all frames where the animation time falls within the [start, end] range

Optionally allow fading or styling based on whether the current frame is start/middle/end

This could be implemented either by:

A new “Time Range” animation mode, or

A toggle in the current animation filter for duration-based presence

🧠 Workarounds Tried

I've tried expanding each row into multiple rows using GENERATE_DATE_ARRAY(start, end) in BigQuery, with one row per day — which technically works, but results in massive datasets (millions of rows), difficult to upload or render due to memory limits.

Lowering the row count makes the animation jerky or incomplete.

🙏 Summary

I’d love to see a more scalable way to visualize spanning temporal data in Kepler.gl — especially one that doesn’t require exploding the dataset to millions of duplicates.

Thanks so much for considering this!

— A fan of dots that stick around ⏳🗺️

igorDykhta:
## Summary
Add an optional **End time** field on time filters so features stay visible for a `[start, end]` span during playback, instead of only at a single timestamp.

Closes [#3198](https://github.com/keplergl/kepler.gl/issues/3198).

- A feature is shown while the playback window overlaps `[start, end]`; missing end times are treated as still active.
- The timeline histogram counts a feature in every bin that falls inside that range.
- Instant (single-timestamp) time filters are unchanged.

## Test plan
- [x] Load a dataset with start and end timestamp columns (e.g. `examples/sf-start-end-times.csv`)
- [x] Add a time filter on the start field, then set **End time** in the filter panel
- [x] Confirm features persist across frames that fall inside their interval
- [x] Confirm the timeline bars fill the start–end span, not only the start instant
- [x] Clear End time and confirm single-timestamp playback is restored

### DIFF
--- a/docs/user-guides/e-filters.md
+++ b/docs/user-guides/e-filters.md
@@ -15,7 +15,7 @@ To add a column filter:
 ![choose a dataset](https://d1a3f4spazzrp4.cloudfront.net/kepler.gl/documentation/image29.png "choose a dataset")
 
 4. Your filter is applied to your map as soon as you specify the field and value.
-5. For timestamp fields, you can optionally choose an **End time** field so features stay visible for the whole `[start, end]` span during playback. See [Playback](./h-playback.md).
+5. For timestamp fields, you can optionally choose an **End time** field so features stay visible for the whole `[start, end]` span during playback. End time is hidden while the filter is synced across datasets. See [Playback](./h-playback.md).
 6. Delete a filter anytime by clicking the __trashcan__ to the right of the filter you wish to delete.
 
 __Note__: column filters apply to all layers in the same dataset on your map.
--- a/docs/user-guides/e-filters.md
+++ b/docs/user-guides/e-filters.md
@@ -15,7 +15,8 @@ To add a column filter:
 ![choose a dataset](https://d1a3f4spazzrp4.cloudfront.net/kepler.gl/documentation/image29.png "choose a dataset")
 
 4. Your filter is applied to your map as soon as you specify the field and value.
-5. Delete a filter anytime by clicking the __trashcan__ to the right of the filter you wish to delete.
+5. For timestamp fields, you can optionally choose an **End time** field so features stay visible for the whole `[start, end]` span during playback. See [Playback](./h-playback.md).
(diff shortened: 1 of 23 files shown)


lixun910 (approved):

jo-chemla:
Nice to see this feature! Quick question for features that only have a single time field and no end-time column: 

Qgis vector layer [temporal properties](https://docs.qgis.org/3.44/en/docs/user_manual/working_with_vector/vector_properties.html#temporal-properties) has loads of options, most standard being single datetime field, seperate fields for start/end datetimes (like this PR), but can also accumulate features over time. How would one do this with current kepler, do you recommend adding a column that sets same end datetime for all fatures (to features max datetime)? 

Also linked to how the temporal controller itself can now choose to accumulate features or not, see gif here in  [animating-a-temporal-navigation](https://docs.qgis.org/3.44/en/docs/user_manual/map_views/map_view.html#animating-a-temporal-navigation)

igorDykhta:
Hey @jo-chemla 

Could you check Incremental Time Window?

Screenshot 2026-09-13 at 6 41 00 AM

PR #3702, opened 2026-09-06
feat: warn before slow grid/hexagon aggregation

### LINKED ISSUE #2935: [Bug] Keple.gl freezes when layer is changed to grid layer
**Describe the bug**
kepler.gl freezes

**To Reproduce**
1. Go to demo
2. Upload [data](https://nextcloud2.datagon.systems/s/ZLsHKr9jEpriGtS)
3. Set layer to grid with resolution 0.1 km
4. kepler.gl freezes

**Expected behavior**
The layer shows on the map

 - OS: Manjaro Linux
 - Browser: Firefox
 - Version 134.0

**Additional context**

igorDykhta:
## Summary
- Show a confirm when changing grid size, hexagon radius, or switching to a grid/hexagon layer would produce about 100k+ occupied cells.
- Aggregation is unchanged: the warning is opt-in UX so a tiny radius on a large dataset is less likely to freeze the UI by surprise.

## Related issues
- Related to https://github.com/keplergl/kepler.gl/issues/2935 (`[Bug] Keple.gl freezes when layer is changed to grid layer`).

Screenshot 2026-09-07 at 1 20 20 AM

**This PR does not close #2935.** The freeze is still there if the user continues, or if a saved map already uses a small cell size.

## Not in this PR
- No worker / async aggregation (an earlier attempt was dropped).
- No hard cap on cell size or occupied-cell count.
- No spinner or cancel while aggregation is running.
- Loading a config that already has a small radius can still freeze on startup.

## Test plan
- [x] On a small dataset (<100k points), changing grid/hex radius never shows the warning.
- [x] On a large dataset, shrinking radius toward native resolution shows “The layer aggregation may be slow with up to N cells” with Cancel / Continue.
- [x] Cancel keeps the previous radius; Continue applies the new size (may be slow).
- [x] Switching layer type to grid/hexagon uses the same confirm when the estimate is over the threshold.
- [x] After Cancel or Continue, the layer is not stuck: slider and map stay in sync.

### DIFF
--- a/src/components/src/side-panel/layer-panel/aggregation-size-warning.tsx
+++ b/src/components/src/side-panel/layer-panel/aggregation-size-warning.tsx
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+// Copyright contributors to the kepler.gl project
+
+import React from 'react';
+import styled from 'styled-components';
+import {FormattedMessage} from '@kepler.gl/localization';
+import {Button} from '../../common/styled-components';
+
+const WarningBox = styled.div`
+  margin-top: 8px;
+  padding: 8px 10px;
+  border-radius: 2px;
+  background-color: ${props => props.theme.panelBackgroundHover};
+  color: ${props => props.theme.textColor};
+  font-size: 11px;
+  line-height: 16px;
+`;
+
+const WarningText = styled.div`
+  color: ${props => props.theme.warningColor || props.theme.errorColor || props.theme.textColorHl};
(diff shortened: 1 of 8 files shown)


igorDykhta:
Closing. Aggregation performance is improved in the latest deck. There is still a possibility of very slow aggregation or even a crash in the case of a world-spanning dataset with millions of points + a small aggregation cell size, but I guess it's fine for now for the majority of users.

PR #3703, opened 2026-09-07
feat: WIP run grid/hexagon aggregation on a web worker

igorDykhta:
## Summary
- Offload CPU grid and hexagon aggregation to a web worker for datasets of 20k+ points (count, sum, average, min, max), so pan/zoom stay responsive during the hitch that sync aggregation already has on large data.
- Keep the last cells on screen and show the existing loading indicator while the worker runs.
- Sample legend/color-domain bins so a huge occupied-cell count does not allocate millions of UI objects.

Median, mode, and stdev still run on the main thread. Stacked on #3702.

On master, 2M points at 0.01 km already completes with only a short freeze. This PR is about keeping the UI interactive during that work, and about the worse case (very small cells / occupied cells ≈ point count) that can still lock the tab.

Screenshot 2026-09-07 at 4 33 48 AM

## Related issues
- Related to https://github.com/keplergl/kepler.gl/issues/2935 (`[Bug] Keple.gl freezes when layer is changed to grid layer`).

## Test plan
- [ ] On a small dataset (<20k points), grid/hex aggregation behaves as before (sync, no spinner).
- [ ] On a large dataset (e.g. 500k–2M points), changing radius or switching to grid/hexagon shows “aggregating grid / hexagon cells” and the map still responds to pan/zoom.
- [ ] Previous cells stay at the old size until the new aggregation finishes (no cells scaling around the origin).
- [ ] Count / sum / average / min / max match the previous sync results; median / mode / stdev still work (sync).
- [ ] Hover tooltip still shows point count for worker-backed bins.

### DIFF
--- a/examples/demo-app/esbuild.config.mjs
+++ b/examples/demo-app/esbuild.config.mjs
@@ -47,6 +47,8 @@ const KEPLER_SRC_ALIASES = Object.fromEntries(
     'common-utils',
     'components',
     'constants',
+    'deckgl-arrow-layers',
+    'deckgl-layers',
     'duckdb',
     'layers',
     'localization',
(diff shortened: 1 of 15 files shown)


igorDykhta:
Postpone for now. The main problem is still data copying in the main thread.

PR #3704, opened 2026-09-07
fix: draw float32 raster tiles

javandres:
## Problem

A `float32` raster never draws through `RasterTileLayer`. The layer mounts, the
STAC metadata is read, and then nothing happens: no tiles are requested, no
error is logged, no warning is shown. There are **two** guards in the way, and
the first hides the second.

### 1. `getPixelRange` gives up on any data type without a fixed maximum

`dtypeMaxValue.float32` is `null` (as are `float64`, `int64`, `uint64`), so
`getPixelRange` returns before it can look anywhere else:

```js
const maxPixelValue = dtypeMaxValue[dtype];

// TODO check if this early return is expected
if (!maxPixelValue) {
  return null;
}

if (
  !Number.isFinite(maxPixelValue) &&        // unreachable
  typeof minRasterStatsValue === 'number' &&
  typeof maxRasterStatsValue === 'number'
) {
  return [minRasterStatsValue, maxRasterStatsValue];
}
```

The fallback for exactly this case — the band statistics carried by the STAC
metadata — is written on the next line and can never run: no entry in
`dtypeMaxValue` is both truthy and non-finite, so `!Number.isFinite(maxPixelValue)`
is unreachable. The statistics path is there and disabled.

`getDataSourceParams` then returns `null`, and the layer requests no tiles at all.

### 2. The float mask discards every valid pixel

`raster/webgl/texture/mask.ts` uses `Math.pow(2, 62)` as its "no bound" default:

```js
const inf = Math.pow(2, 62);   // 4.6e18
...
if (mask_value > mask_image_float.keepMax) discard;
```

A tile server writes the mask in the band's own data type and marks valid pixels
with that type's maximum — 255 for `uint8`, 65535 for `uint16`, and `FLT_MAX`
(3.4028235e38) for `float32`. kepler's own comment in `gpu-utils.ts` says as much:

```js
// In general, data masks are 0 for nodata and the maximum value for valid data,
// e.g. 255 or 65535 for uint8 or uint16 data, respectively
moduleProps.maskKeepMin = 1;
```

3.4e38 is greater than 4.6e18, so for a float32 mask the upper bound discards
every valid pixel and the tile disappears — again with no error.

### Reproduced

A float32 NDVI COG served by TiTiler 2.2.1, whose `/cog/stac` carries
`raster:bands[0].statistics` (`{"minimum": -0.0131, "maximum": 0.7516}`), loaded
as a `raster-tile` dataset:

| | `/cog/stac` requests | tile requests | pixels drawn |
|---|---|---|---|
| master | 1 | 0 | 0 |
| with fix 1 only | 1 | 4 | 0 |
| with both fixes | 1 | 4 | 409,045 |

The same COG rewritten as `uint16` draws on master, which is the workaround this
has been costing users.

## Fix

- **`getPixelRange`**: when the data type has no fixed maximum, use the STAC band
  statistics, and return `null` only when there are none — nothing to rescale
  against. Integer types keep their current `[0, dtypeMaxValue]` range.
- **`mask.ts`**: raise the unbounded sentinel to `FLT_MAX`, so "no upper bound"
  really is no upper bound for a float32 mask. Anything a mask can carry is now
  inside it.

## Not fixed here

`getRasterStatisticsMinMax` returns statistics only for the `singleBand` preset,
so a float32 **multi-band** composite (`trueColor` and friends) still has no
range and still will not draw. That looked like a separate decision about where
the range for a composite should come from, so this PR leaves it alone. Happy to
widen it if you would rather have both in one change.

## Tests

Four cases in `test/browser/layer-tests/raster-tile-layer-specs.js`:

- `getDataSourceParams float32 band with statistics` — fails on master
- `getDataSourceParams float32 band without statistics` — still `null`
- `getDataSourceParams integer band uses the data type range` — unchanged behaviour
- `float mask keeps the pixels a float32 mask marks as valid` — fails on master

Full browser suite: 3597 → 3600 passing, no failures. `tsc --noEmit` clean,
`eslint` clean on the touched files.

### DIFF
--- a/src/deckgl-layers/src/raster/webgl/texture/mask.ts
+++ b/src/deckgl-layers/src/raster/webgl/texture/mask.ts
@@ -4,7 +4,10 @@
 import type {Texture} from '@luma.gl/core';
 import {GetUniformsOutput, ShaderModule} from '../types';
 
-const inf = Math.pow(2, 62);
+// The "no bound" default for keepMin/keepMax. A tile server writes the mask in the band's own
+// data type and marks valid pixels with that type's maximum — 255 for uint8, 65535 for uint16,
+// FLT_MAX for float32 — so a smaller default discards every valid pixel of a float32 mask.
+const inf = 3.4028234663852886e38;
 
 function getUniforms(
   opts: {imageMask?: Texture; maskKeepMin?: number; maskKeepMax?: number} = {}
--- a/src/layers/src/raster-tile/raster-tile-utils.ts
+++ b/src/layers/src/raster-tile/raster-tile-utils.ts
@@ -156,17 +156,13 @@ function getPixelRange(
   const minPixelValue = 0;
   const maxPixelValue = dtypeMaxValue[dtype];
 
-  // TODO check if this early return is expected
-  if (!maxPixelValue) {
+  // Data types without a fixed maximum, such as floats, can only be rescaled from the
+  // statistics reported by the STAC metadata.
+  if (maxPixelValue === null || !Number.isFinite(maxPixelValue)) {
+    if (typeof minRasterStatsValue === 'number' && typeof maxRasterStatsValue === 'number') {
+      return [minRasterStatsValue, maxRasterStatsValue];
+    }
     return null;
-  }
-
-  if (
-    !Number.isFinite(maxPixelValue) &&
-    typeof minRasterStatsValue === 'number' &&
-    typeof maxRasterStatsValue === 'number'
-  ) {
-    return [minRasterStatsValue, maxRasterStatsValue];
   }
 
   return [minPixelValue, maxPixelValue];
--- a/test/browser/layer-tests/raster-tile-layer-specs.js
+++ b/test/browser/layer-tests/raster-tile-layer-specs.js
@@ -8,8 +8,10 @@ import {
   findAssetWithName,
   getUsableAssets,
   getRasterStatisticsMinMax,
-  filterAvailablePresets
+  filterAvailablePresets,
+  getDataSourceParams
 } from '@kepler.gl/layers';
+import {RasterWebGL} from '@kepler.gl/deckgl-layers';
 import {parseRasterMetadata} from '@kepler.gl/table';
 import {testCreateCases} from 'test/helpers/layer-utils';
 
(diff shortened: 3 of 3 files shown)


igorDykhta:
Hi @javandres Any chance you could provide the metadata for a Float32 raster tile like this so I could test it?

javandres:
> Hi @javandres Any chance you could provide the metadata for a Float32 raster tile like this so I could test it?

Thanks, @igorDykhta.  Here is metadata plus a reproduction that needs nothing of mine — it runs in the demo app against a
public COG, and I have just walked through it both ways to check it.

## Repro, zero setup

Copernicus DEM 30 m is public, single-band **float32**, and TiTiler reports statistics for it:

```
https://copernicus-dem-30m.s3.amazonaws.com/Copernicus_DSM_COG_10_N45_00_E006_00_DEM/Copernicus_DSM_COG_10_N45_00_E006_00_DEM.tif
```

In the demo app: **Add Data → Tileset → Raster Tile**, paste that `.tif`, **Add Tileset**.
`isCOGUrl` matches, so the form points metadata and tiles at `https://titiler.xyz` by itself —
there is nothing else to fill in. It is the same shape as the built-in **Africa Farms** example,
which is a `TCI.tif`, i.e. `uint8` and one of the types that already works.

The metadata preview fills in identically on both sides, so the STAC is parsed fine and the failure
is downstream of it:

```json
"raster:bands": [
  {
    "data_type": "float32",
    "statistics": {
      "mean": 1700.306884765625,
      "minimum": 244.705078125,
      "maximum": 4801.47314453125,
      "stddev": 743.3125
    }
  }
]
```

Then the map. Both captures are the demo app at 3.3.0-alpha.10, same viewport, same layer in the
tree; the only difference between them is `raster-tile-utils.ts` + `mask.ts` reverted to `master`.

**`master` (68e844d) — dataset added, layer listed, nothing drawn:**

1-master-dem-blank

**With this PR — the DEM draws:**

2-patched-dem-drawn


## The second guard, in one request

A tile that server actually returns for the same COG — `9/265/183` covers Mont Blanc, and
`return_mask=true` is what makes it two planes, band then mask:

```
$ curl -s 'https://titiler.xyz/cog/tiles/WebMercatorQuad/9/265/183.npy?url=https%3A%2F%2Fcopernicus-dem-30m.s3.amazonaws.com%2FCopernicus_DSM_COG_10_N45_00_E006_00_DEM%2FCopernicus_DSM_COG_10_N45_00_E006_00_DEM.tif&bidx=1&return_mask=true' -o tile.npy

>>> a = numpy.load('tile.npy')
shape (2, 256, 256)  dtype float32
band  min 0.0  max 3722.2239
mask  unique [-3.4028235e+38  3.4028235e+38]
```

The mask marks valid data with `+FLT_MAX` (3.4028235e38). `mask.ts` defaults `keepMax` to
`Math.pow(2, 62)` = 4.6e18, so `if (mask_value > keepMax) discard` throws away every valid pixel.
Worth noting it writes nodata as `-FLT_MAX` rather than 0, which `maskKeepMin = 1` already handles —
only the upper bound is wrong.

## The raster the description's numbers came from

Attached as `ndvi-float32.tif.zip` (230 KB) if you would rather serve it from your own TiTiler —
zipped only because GitHub will not take a `.tif`. Its `/cog/stac`, trimmed to what the layer reads:
[ndvi-float32.tif.zip](https://github.com/user-attachments/files/31930831/ndvi-float32.tif.zip)

```json
{
  "type": "Feature",
  "stac_version": "1.1.0",
  "stac_extensions": [
    "https://stac-extensions.github.io/projection/v1.1.0/schema.json",
    "https://stac-extensions.github.io/raster/v1.1.0/schema.json",
    "https://stac-extensions.github.io/eo/v1.1.0/schema.json"
  ],
  "id": "ndvi-float32.tif",
  "bbox": [-79.1, -3.05, -78.9, -2.85],
  "properties": {"proj:epsg": 4326},
  "assets": {
    "data": {
      "href": "http://derived-data/ndvi-float32.tif",
      "type": "image/tiff; application=geotiff",
      "raster:bands": [
        {
          "data_type": "float32",
          "scale": 1.0,
          "offset": 0.0,
          "sampling": "area",
          "nodata": "-inf",
          "statistics": {
            "mean": 0.3800666332244873,
            "minimum": -0.013052688911557198,
            "maximum": 0.751609206199646,
            "stddev": 0.14480004167502852,
            "valid_percent": 100.0
          }
        }
      ],
      "eo:bands": [{"name": "b1", "description": "gray"}],
      "roles": []
    }
  }
}
```

That one is TiTiler 2.2.1; the DEM above goes through `titiler.xyz`. Both carry
`raster:bands[0].statistics`, which is the range the patched `getPixelRange` reads.

igorDykhta (approved):

igorDykhta:
@javandres landed. Deployed to https://kepler.gl/demo Thanks!

PR #3719, opened 2026-09-11
feat: snap annotations to 3D surfaces

### LINKED ISSUE #3713: [Feat] Attribution - snap to a 3D point
Right now attribution is snapped to XY, ignores Z

igorDykhta:
## Summary
- Annotation snap points now store `[lng, lat, altitude]` and stay on 3D surfaces when the camera pitches.
- Dragging a snap handle uses deck.gl `pickObject({unproject3D: true})` to reconstruct XYZ from the surface under the cursor (3D tiles, extruded GeoJSON, raster terrain, Mapbox 3D buildings).
- Falls back to the ground plane when nothing is picked. Existing `[lng, lat]` configs still load.

Fixes #3713

## Test plan
- [x] Add an annotation and confirm it appears at map center without moving the camera
- [x] Drag the snap point onto extruded GeoJSON / 3D tiles / raster terrain, then pitch the camera — the marker should stay on the surface
- [x] Drag into empty sky — the marker should drop back to the ground plane
- [x] Reload a saved map that still has 2D `[lng, lat]` anchors

https://github.com/user-attachments/assets/8bc462d3-836a-421d-9ec9-2bba3bf7def6

### DIFF
--- a/src/components/src/annotations/annotation-overlay.tsx
+++ b/src/components/src/annotations/annotation-overlay.tsx
@@ -18,6 +18,7 @@ import AnnotationNode from './annotation-node';
 import AnnotationText from './annotation-text';
 import {
   MapViewport,
+  PickWorldPosition,
   isPointVisibleOnGlobe,
   movePoint,
   moveText,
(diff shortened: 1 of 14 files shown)


lixun910 (approved):
Nice 👍

PR #3721, opened 2026-09-11
feat: add SQLRooms integration and a separate experimental demo

ilyabo:
image

---

Add `@kepler.gl/sqlrooms` and a separate experimental demo for testing SQLRooms 0.29.0 alongside the original Kepler application. `examples/demo-app` and `website` match the pre-migration base commit exactly; the website continues to serve the original `/demo` app.

Run both from separate terminals at the repository root:
- `yarn start`: original demo on port 8080.
- `yarn start:sqlrooms`: experimental demo on port 8083 (`examples/demo-app-sqlrooms`).

The experimental app has its own dependency installation, lockfile, entry point, and build. It preserves Kepler's existing Redux workflows, loading dialogs, logo/version, map controls, exports, and provider integrations while composing the map, SQL, and assistant panels through SQLRooms. It adds a collapsible sidebar with a floating toggle over the map (no title/header row), Kepler colors/fonts, and removes the obsolete DuckDB preview callout. Sample fixtures are shared with the original demo. Different origins isolate browser storage for comparison.

Package and architecture:
- Import the adapter and schemas from SQLRooms commit `26d8e78e086cebdfc4eb6b4047c9035c5704a068`, retaining MIT attribution, and expose the former config package through the lightweight `/config` entry point.
- Add `/shell` for portaling the original sidebar into SQLRooms while retaining contexts and popup placement. Honor zero sidebar width in Kepler's selectors.
- Retain existing data ownership and saved-map formats in the experimental app. New functionality can register panels in its room layout; it does not convert every map to `KeplerSlice` or introduce a second database connection.
- Adapt async task execution and dataset synchronization to current Kepler APIs; add CJS/ESM/declaration exports and focused CI tests. Kepler foundations do not depend on the adapter.
- Port SQLRooms [#920](https://github.com/sqlrooms/sqlrooms/pull/920) and [#922](https://github.com/sqlrooms/sqlrooms/pull/922): preserve incomplete saved maps during asynchronous hydration, discard superseded restore results, duplicate pending maps from preserved config, and expose `isMapConfigPending(mapId)` plus dataset-reference/persistence helpers. Keep the local task runtime and dispatch integration.

See [Component overrides and possible core migration](https://github.com/keplergl/kepler.gl/blob/feat/sqlrooms-integration/src/sqlrooms/docs-dev/component-overrides-and-core-migration.md) for the override inventory and proposed core migration.

Validation:
[...]

PR #3727, opened 2026-09-14
Deploy SQLRooms demo at /demo-next alongside the website

ilyabo:
The SQLRooms demo currently runs separately from the published website. Build and publish it at `/demo-next` alongside the existing `/demo`, using the same Netlify publish directory and existing repo-controlled build entry point.

- Extend root `install:web` to install the SQLRooms demo's locked dependencies. The website's `build` composes its own `esbuild` with `build:demo-next`, which builds and copies the separate preview bundle into `website/dist/demo-next/` without installing dependencies. No Netlify dashboard or DNS changes are needed for the existing Git-connected deployment.
- Run SQLRooms demo unit tests in CI for PRs targeting `master` or `feat/sqlrooms-integration`; keep Chromium checks manual and the preview unlisted in marketing navigation.
- Add SPA rewrites and prefix the new app's assets, sample navigation, and cloud-share links with `/demo-next`. Standalone development and builds keep their existing `/demo` paths. A dedicated build identifier keeps dotenv/environment values from overriding the route prefix.
- Isolate experimental assistant sessions/settings under a separate storage key in the host-owned persistence configuration. This uses the assistant composition now in #3721 and needs no build-time package rewriting. Keep existing provider login storage and OAuth callbacks, including `/auth`.

Stacked on `feat/sqlrooms-integration` (#3721).

Validation: the complete website production build and standalone SQLRooms build pass; all 15 SQLRooms unit tests pass; the SQLRooms shared-room Chrome integration test and the deployment Chrome test pass for both entry points, `/demo-next` with/without a trailing slash, asset responses, sample navigation with a SQL query, nested reloads, and assistant storage isolation. Changed application files pass ESLint and the diff passes whitespace checks. The revised `install:web`, website-only build, preview build, and deployment browser check also pass with a conflicting `DEMO_BASE_PATH` environment value. Production build validation used placeholder provider configuration; live OAuth and AI-provider requests were not tested. The combined Netlify build will take longer and either app's build failure will block deployment.


website/package.json
```diff
@@ -7,13 +7,14 @@
...
-    "esbuild": "yarn build-clean && yarn build-static && node esbuild.config.mjs --build",
+    "esbuild": "yarn build-clean && yarn build-static && node esbuild.config.mjs --build && yarn build:demo-next",
     "start": "webpack serve --mode development --env prod --static-directory ./src/static --progress --open",
     "start-prod": "webpack serve --mode production --env prod --static-directory ./src/static --progress --open",
     "build-clean": "rm -rf ./dist && mkdir dist",
     "build-static": "cp -r ./src/static/* dist",
     "build-script": "NODE_OPTIONS=--openssl-legacy-provider webpack --mode production --env prod",
-    "build": "yarn build-clean && yarn build-static && yarn esbuild",
+    "build": "yarn esbuild",
+    "build:demo-next": "cd ../examples/demo-app-sqlrooms && yarn install --immutable && yarn build --demo-next && cp -R dist/demo-next ../../website/dist/",
```
igorDykhta: `yarn deploy` still only installs `examples/demo-app` and `website` (`install:web`). This nested `yarn install --immutable` runs in the website *build* step, so it likely misses Netlify’s usual `node_modules` cache and re-downloads DuckDB/Monaco on every deploy.

Two small splits would match the existing demo-app pattern:

1. Install SQLRooms from root `install:web` (alongside `install:example`), and keep `build:demo-next` as build + copy.
2. Leave `esbuild` as the website bundle only; have `build` run `yarn esbuild && yarn build:demo-next`. Folding SQLRooms into `esbuild` means a website-only rebuild pays for a second app.

The old `build` → `esbuild` double-clean is a good fix; worth keeping that part.

examples/demo-app-sqlrooms/README.md
```diff
@@ -1,7 +1,8 @@
 # SQLRooms Demo App
 
 An experimental version of the Kepler demo for testing SQLRooms 0.29.0. It runs
-beside the original `examples/demo-app`; the website still uses the original app.
+beside the original `examples/demo-app`. The website serves the original app at
+`/demo` and this app at `/demo-next` as separate bundles.
```
igorDykhta, line 5: Accurate for the website. Two spots below still read as if the SQLRooms app lives at `/demo`:

- L29: “Both apps support `/demo` and sample routes such as `/demo/earthquakes`.”
- L98: the SQL panel example is still `/demo/earthquakes?sql=...`

Standalone `yarn start` / `yarn build` should keep `/demo`; the website preview should show `/demo-next/...`.

examples/demo-app-sqlrooms/esbuild.config.mjs
```diff
@@ -182,10 +185,12 @@ const config = {
     '.woff2': 'file'
   },
   entryPoints: ['src/main.js'],
-  outfile: 'dist/bundle.js',
+  outfile: join(outdir, 'bundle.js'),
+  publicPath,
   bundle: true,
   define: {
     NODE_ENV,
+    'process.env.DEMO_BASE_PATH': JSON.stringify(demoNext ? '/demo-next' : '/demo'),
```
igorDykhta: `@dotenv-run/esbuild` also injects `process.env.*` from `../../.env`. A local or Netlify `DEMO_BASE_PATH` can clash with this `--demo-next` define (duplicate define, or the flag is ignored).

Worth a README note that `DEMO_BASE_PATH` should not be set in `.env`, or a dedicated identifier dotenv cannot override.

examples/demo-app-sqlrooms/test/duckdb-browser-entry.ts
```diff
@@ -221,7 +221,7 @@ export async function run() {
     assistantStore.getState().aiSettings.config.providers.openai.apiKey === 'test-only-key',
     'Closing assistant lost provider settings'
   );
-  const persisted = JSON.parse(localStorage.getItem('kepler-ai-assistant-state')!);
+  const persisted = JSON.parse(localStorage.getItem('kepler-sqlrooms-ai-assistant-state')!);
   check(
     persisted.state.ai.sessions.some((s: {id: string}) => s.id === sessionId),
     'Existing persistence key was not retained'
```
igorDykhta: This now asserts the new `kepler-sqlrooms-ai-assistant-state` key (good), but the message still says the existing persistence key was retained. Something like “SQLRooms assistant state was not persisted” would match the check.

igorDykhta (commented):
## Review

The `/demo` vs `/demo-next` split, SPA rewrites, asset prefix, and assistant storage isolation look right. A few deploy/docs tweaks would make this less brittle.

Inline comments cover the script wiring, README path mix-up, `DEMO_BASE_PATH` vs dotenv, and the stale persistence assertion.

Not in the diff, but related:

- Add `!examples/demo-app-sqlrooms/yarn.lock` next to the other example lockfile exceptions. It is already tracked, and `--immutable` is now on the production path.
- `test.yml` only runs for PRs targeting `master`, so `deployment-routes.test.cjs` will not run here. Fine to keep the Chromium check manual.
- The marketing site still only links to `/demo`. If this preview should stay unlisted, that is fine; otherwise a banner or header item would help people find `/demo-next`.

igorDykhta (approved):

PR #3728, opened 2026-09-15
fix: keep layer order when replacing datasets back to back

javandres:
# fix: keep layer order when replacing datasets back to back

## Summary

When two datasets are replaced with `replaceDataInMap` one right after the other, the layers of the
first dataset are merged back in **reverse order**. An app that refreshes several datasets in a row
sees its layers swap on every refresh — for example a heatmap meant to sit under a point layer ends
up on top of it.

## Cause

`prepareStateForDatasetReplace` parks the replaced dataset's layers in `layerToBeMerged`, and
`replaceDatasetDepsInState` records their order in `preserveLayerOrder`, which `mergeLayers` uses to
put them back.

The condition guarding that assignment only checks that `layerToBeMerged` is non-empty:

```ts
if (mergerOptions.toMergeProp !== undefined && replacedState[mergerOptions.toMergeProp]?.length && preserveOrder) {
  replacedState[preserveOrder] = propValue.map(item => item.id);
}
```

When a second dataset is replaced before the first one's layers have merged back, the list still
holds the **first** dataset's layers, so the condition passes and `preserveLayerOrder` is overwritten
with the second dataset's layers — an empty list if it has none. `insertItemBasedOnPreservedOrder`
then inserts every layer at the front, and the first dataset's layers come back reversed.

Observed state transitions in an app (two layers on dataset A, none on dataset B, refreshed A then B):

```
prepare A   preserveLayerOrder [points, heatmap]   layerToBeMerged [points, heatmap]
prepare B   preserveLayerOrder []                  layerToBeMerged [points, heatmap]
merge A     layerOrder [heatmap, points]
```

## Fix

Only overwrite the preserved order when the dataset being replaced actually had items to park
(`replacedItem` is `null` otherwise). One condition added.

## Test

`test/node/reducers/composer-state-test.js` — *replaceDataInMapUpdater: replacing a dataset with no
layers before the previous replace merges*: loads `sampleConfig` (two layers on one dataset) plus a
second dataset without layers, replaces both back to back, applies the tasks, and checks the layer
order.

- Before the fix: `expected [ '3zucml7', 'nob639j' ]`, `actual [ 'nob639j', '3zucml7' ]`.
- After: `composer-state-test.js` 704/704, `vis-state-test.js` 4899/4899.


igorDykhta (approved):
Thanks again for your great contributions!

PR #3735, opened 2026-09-17
fix: keep split map panels index-aligned when merging split maps

javandres:
fix: keep split map panels index-aligned when merging split maps

## Summary

Merging split maps into a map breaks its panels. Three flows show it:

| Flow | origin/master | This PR |
|---|---|---|
| `toggleSplitMap`, then `replaceDataInMap` | 3 panels, the layer shows in all of them | 2 panels, right panel stays empty |
| Saved split config (one layer per side), two datasets replaced back to back, 3 times | 4 → 8 → 16 panels, every layer in every panel | 2 panels, one layer per side |
| `addDataToMap` with a dual map saved as `[{layers: {a: true, b: true}}, {layers: {}}]` | 1 panel | 2 panels, right panel empty |

## How to reproduce

```js
import {addDataToMap, toggleSplitMap, replaceDataInMap} from '@kepler.gl/actions';

dispatch(addDataToMap({datasets: {info: {id: 'trips'}, data}}));
dispatch(toggleSplitMap());
// visState.splitMaps: [{layers: {<layerId>: true}}, {layers: {}}]

dispatch(replaceDataInMap({datasetToReplaceId: 'trips', datasetToUse: {info: {id: 'trips'}, data}}));
// once the dataset is created, visState.splitMaps on origin/master:
// [{layers: {<layerId>: true}}, {layers: {<layerId>: true}}, {layers: {<layerId>: true}}]
```

The new tests cover all three flows. Run against the `src` of origin/master, 16 of their assertions fail.

## Causes

1. `mergeSplitMaps` appended panels without layers (`merged.push(sm)`) and layers still waiting to be merged (`[...splitMapsToBeMerged, ...unmerged]`), instead of matching them by index. The waiting panels could also be sparse: spreading them left `undefined` entries, which throw on the next merge.
2. `prepareStateForDatasetReplace` overwrote `splitMapsToBeMerged` with a snapshot of the current split maps. When a second replace runs before the first one has merged back, that snapshot no longer has the first dataset's layers, so their visibility per panel is lost.
3. `postMergeUpdater` added each new layer to every panel that didn't list it, including layers the merged split maps had just placed. A panel saved without a layer got it back. Panels created by `mergeSplitMaps` had the same problem: they started with every visible layer.
4. `mergeSplitMaps` wrote `layers` into panel objects shared with the previous state.

## Changes

- `mergeSplitMaps` matches panels by index across `splitMaps`, `state.splitMaps` and `splitMapsToBeMerged`, never leaves holes, and builds new panel objects.
  - On a map that is not split, it creates every panel once one of them can be merged (it has no layers, or one of its layers exists). While all layers are still loading, the map stays unsplit, as before.
  - Created panels start with the current layers that the merged split maps don't list. A layer listed in the split maps only shows in the panels that list it.
- `prepareStateForDatasetReplace` combines its snapshot, by index, with the split maps already waiting to be merged.
- `postMergeUpdater` only adds the new layers that no panel lists.
- New util `combineSplitMapsByIndex` in `split-map-utils`.

## Tests

- `vis-state-merger-test`: new `mergeSplitMaps -> keep panels index-aligned` (holes, waiting layers, current layers, split map layers, no mutation).
- `composer-state-test`: the back-to-back replace test now compares the panels with `deepEqual`. New tests replace a dataset in DUAL_MAP mode and load a dual map saved with an empty panel, including saving its config while the layers load.
- Node suite: 12177 of 12179 pass. The 2 failures are in `processors/file-handler-test.js` (arrow `Float64`) and also fail on origin/master.
- Browser suite (jsdom, Node 20): 3717 of 3717 pass. `tsc --noEmit` and eslint pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


igorDykhta (approved):
Great fix as always!

